PR #87, opened 2020-02-23
Update README.md

Malvoz:
Mainly:
- [x] Remove the webcomponents.org badge, since [the link](https://www.webcomponents.org/element/Maps4HTML/Web-Map-Custom-Element) doesn't return anything useful.
- [x] Change <span>@</span>{attr} to `{attr}` because it is more common, and I find it easier to read.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,30 +1,28 @@
 
-# Customized Built-In &lt;map&gt; Element
+# Customized built-in &lt;map&gt; element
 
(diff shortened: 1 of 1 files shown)


prushforth:
I did a little bit of work on npm to set up a package to allow people to install this via npm install web-map, but I haven't finished; have been meaning to get back to it (they prevent some operations from happening more frequently than once every 24 hours, which is  a barrier when you are error prone ;-). I think the Web Components.org link will come back to life after the npm work is done, but OK to remove for now.

Issue #88, opened 2020-02-23
`.mapml-control-layers` invalid HTML


Malvoz:
<details><summary>HTML</summary>

``` html
<form class="leaflet-control-layers-list style-scope web-map">
    <label class="style-scope web-map">
        <details class="mapml-control-layers style-scope web-map">
            <summary class="style-scope web-map">
                <input type="checkbox" class="leaflet-control-layers-selector style-scope web-map">
                  <span draggable="true" class="style-scope web-map" style="font-style: italic;">
                    Canada Base Map - Transportation (CBMT)
                 </span>
            </summary>
            <details class="mapml-control-layers style-scope web-map">
                <summary class="style-scope web-map">opacity</summary>
                <input type="range" min="0" max="1.0" value="1.0" step="0.1" class="style-scope web-map">
            </details>
        </details>
    </label>
</form>
```

</details>

nuvalidator output:

> Error: Element `details` not allowed as child of element `label` in this context. (Suppressing further errors from this subtree.)
>
> Error: The `label` element may contain at most one `button`, `input`, `meter`, `output`, `progress`, `select`, or `textarea` descendant.
>
> Warning: `label` element with multiple labelable descendants.

The real world issue here is that none of the input elements have an accessible name and as such inaccessible to screen readers.

prushforth:
Is is as simple as giving them a name attribute? Seems there's a validation error to be concerned about as well (perhaps).

Malvoz:
You can nest `<input>` in a `<label>` element to provide the input an accessible name like so:

``` HTML
<label>
  <span>the accessible name</span>
  <input type="range">
</label>
```

Or if nesting isn't desirable you can assign the input and ID and reference it from the label using the `for` attribute:

``` HTML
<label for="a">
  <span>the accessible name</span>
</label>
<input id="a" type="range">

<!-- or -->

<input id="b" type="range">
<label for="b">
  <span>the accessible name</span>
</label>
```

Example using `<label>` within `<summary>` as the accessible name for the input:

```HTML
<details>
  <summary>
    <label for="c">the accessible name</label>
  </summary>

  <input id="c" type="range">

</details>
```

Most browser developer tools today have an accessibility tab where you can view the computed accessible name (amongst other things).

prushforth:
OK I'll have a look. Thanks!

Malvoz:
[...]

Issue #83, opened 2020-02-20
Handle fullscreen declaration(s)?


Malvoz:
<s>Add [`allow="fullscreen"`](https://w3c.github.io/webappsec-feature-policy/#iframe-allow-attribute) and/or deprecated [`allowfullscreen`](https://w3c.github.io/webappsec-feature-policy/#iframe-allowfullscreen-attribute) attributes to `<map>`, if either declaration is present</s> invoke [fullscreen control for Leaflet](http://leaflet.github.io/Leaflet.fullscreen/) if `controls` is present, unless `controlslist="nofullscreen"` is specified.

Use case: https://maps4html.org/HTML-Map-Element-UseCases-Requirements/#use-case-fullscreen

Malvoz:
Might be worth checking out @runette's [fork](https://github.com/runette/Leaflet.fullscreen).

prushforth:
Thanks for the updated allow="fullscreen" vs allowFullScreen.    Do you think we should go with allow=, since allowFullScreen is deprecated?  The prose around allowFullScreen leads into deep browsing context waters, but I think there's a big discussion to be had about the map element's layers potentially each being a browsing context.   But that should probably be a separate issue.

> Might be worth checking out @runette's [fork](https://github.com/runette/Leaflet.fullscreen).

Yes that would likely be pretty easy to add into the project.  I was surprised that (Chrome) did actually allow it to go fullscreen even using http (I thought fullscreen was only allowed in https).

Malvoz:
My intention with this was to have the `<map>` function similar to `<video>` but what I suggested is not how that works. `<video controls>` means fullscreen button is visible, and adding [`controlslist="nofullscreen"`](https://wicg.github.io/controls-list/html-output/multipage/embedded-content.html#attr-media-controlslist) disables it.

prushforth:
I have been playing around with this, and I can get it to work, but because the map container is in a shadow root, when the fullscreen event is generated, I believe it [retargets ](https://fullscreen.spec.whatwg.org/#dom-document-fullscreenelement)the fullscreen element from the map container div to the ancestor div that contains the shadow root. So it doesn't correctly set the classes associated to the fullscreen-on state.  All this to say if I include this library I will have to change the code a bit, so we'll own it.  I don't think that's too big an issue, because the code isn't too big or complicated.  What do you think? @Malvoz

prushforth:
[...]

PR #91, opened 2020-03-12
Add provisional fullscreen support.  Fix generated layer control HTML.

prushforth:
This may be a confusing PR, because it mixes quite a bit of stuff from I think [this](https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/83) and [this](https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/88) issue.

Hopefully you can make some sense of it.  Also, I renamed geo-map to mm-map, because 'geo' in theory is too narrow.  I want to allow the demo pages to run on Safari, and in the short term it's easier to use mm-map than to provide a fallback `<picture>` etc.


prushforth:
> > I renamed geo-map to mm-map
> 
> It's actually changed to "mm-mapp" (as opposed to "mm-map"), did you intend for that? LGTM either way!

Oops, but fine!

PR #106, opened 2020-07-03
Redoing StaticTileLayer to extend GridLayer + restructuring test folder directory

ahmadayubi:
I noticed in pull request #102  .mapml files inside the test directory were all updated, I reverted those changes as they are code breaking changes. For example meta tags were changed from <meta name="..." **/**> to <meta name="...">, removing the **/**. For the XML parser to notice the meta tag it needs the **/** or it won't register as an element, from my understanding (wouldn't pass the test with those cosmetic changes in place). Might be worth while checking if any other XML files are affected by PR #102


Malvoz:
@ahmadayubi Sorry for the troubles from removing the ending forward slashes in the `.mapml` files. I'm not sure why I concluded it was safe to do so in the XML space of all things. 🤦‍♂️ 

> Might be worth while checking if any other XML files are affected by PR #102 

Yes I can take a look after this PR has been merged.

I'll also want to add back the now reverted additions of the `<mapml lang="en">` declarations, as it can become an accessibility issue for screen reader users if the attribute is not specified, even though these are "just" test files.

Issue #66, opened 2017-01-05
zoomTo fails when the zoom is the same as the current zoom


prushforth:

ahmadayubi:
Can you share steps to reproduce the bug? I tried `<map>.zoomTo(lat, lon, zoom)` with the zoom equal to the maps current zoom level, and the map still zoomed to the lat,lon.

prushforth:
It's been a while and may not apply to this version of the code; I will try it on the old polymer version and update this issue if need be.

PR #107, opened 2020-07-10
Implementing a read-only `<layer->`.extent property

### LINKED ISSUE #105: Design and implement the read-only <layer-> extent property
Discussion of the [requirement](https://github.com/Maps4HTML/HTML-Map-Element-UseCases-Requirements/issues/60)
Discussion of the [specification](https://github.com/Maps4HTML/HTML-Map-Element/issues/33)

ahmadayubi:
#### Deriving bounds of layer:
- [x] StaticTileLayer/mapMLStaticTileLayer

- [x] StaticFeature/mapMLFeatures

- [x] TemplatedImageLayer

- [x] TemplatedTileLayer

- [x] TemplatedFeatureLayer

#### Implemented bounds property for:
- [x] StaticTileLayer/mapMLStaticTileLayer

- [x] StaticFeature/mapMLFeatures

- [x] TemplatedImageLayer

- [x] TemplatedTileLayer

- [x] TemplatedFeatureLayer

#### Testing of bounds property for:
- [x] StaticTileLayer/mapMLStaticTileLayer

- [x] StaticFeature/mapMLFeatures

- [x] TemplatedImageLayer

- [x] TemplatedTileLayer

- [x] TemplatedFeatureLayer

- [x] Different projections

- [x] Different CRS

PR also changes the behavior of static features, now requiring a zoom attribute and behaving very similarly to static tiles. It also contains other tests for different layer types.


src/mapml.js
```diff
@@ -944,12 +1048,51 @@ M.MapMLLayer = L.Layer.extend({
                     _leafletLayer: this,
                     crs: this.crs
                   }).addTo(map);
+                  this._setLayerElBounds();
                 }
               }, this);
         }
+        this._setLayerElBounds();
```
prushforth, line 1055: I'm wondering if line 1055 should be inserted in the if block of line 1033, since it is being unconditionally executed here and may result in errors, consequently?

src/mapml.js
```diff
@@ -989,9 +1132,11 @@ M.MapMLLayer = L.Layer.extend({
       }
     },
     onRemove: function (map) {
-        L.DomUtil.remove(this._container);      
+        L.DomUtil.remove(this._container);
+        if(this._staticTileLayer){
```
prushforth, line 1136: This is conditional, what it replaces wasn't.  I'm not sure why it wasn't, nor indeed at this moment why some are conditional others not.

prushforth: More a question of my own logic actually.

ahmadayubi: The new staticTileLayer behaves similar to templatedLayer, where it is only added to the map when static tiles exist in the html file, the others (mapmlvectors, imageLayer) get added to every mapml layer container regardless of whether it's needed or not (not sure why this is the behavior, probably because in those cases you don't know if you need those layers or not prior to creating them). Since staticTileLayer is added conditionally, it is removed only if it exists.

This was actually a change in PR #106 , the difference is now the layer isn't set to null once it's removed (not needed to do anymore).

src/mapml.js
ahmadayubi: The method is called at the end to assign the layer element itself the .bounds property. The method also needs to run in the cases where `this._extent` doesn't exist like maps that only contain static content, placing it inside of the `if(this._extent)` would prevent that.

As for errors, the method checks certain conditions but there may have been cases that I missed, if any come to mind let me know and I can account for them.

src/mapml.js
```diff
@@ -641,8 +727,24 @@ M.Util = {
...
+    element.split(/\s+/gim).forEach(M.Util.parseNumber,a);
+    this.push(a);
+  },
+
+  parseNumber : function(element, index, array){
+    this.push(parseFloat(element));
+  },
 };
+
+M.extractInputBounds = M.Util.extractInputBounds;
```
prushforth, line 743: I think we need a M.parseNumber here...

src/mapml.js
```diff
@@ -916,15 +1018,17 @@ M.MapMLLayer = L.Layer.extend({
...
-          this._tileLayer = M.mapMLTileLayer(this.href?this.href:this._href, 
-          {pane: this._container,
-           _leafletLayer: this});
-        }
-        // what is _mapmlTileContainer for and why have two copies of it.
-        this._tileLayer._mapmlTileContainer = this._mapmlTileContainer;
-        map.addLayer(this._tileLayer);       
-        this._tileLayer._container.appendChild(this._mapmlTileContainer);
+        //only add the layer if there are tiles to be rendered
+        if((!this._staticTileLayer || this._staticTileLayer._container === null) && this._mapmlTileContainer.getElementsByTagName("tiles").length > 0){
```
prushforth, line 1022: Style comment: can we break up really long lines of code onto multiple lines, with indenting so as to show that the next line belongs to the previous if appropriate.  Break after operators and at logical points so as to try to preserve readability.  Thanks!

src/mapml.js
```diff
@@ -961,12 +1057,56 @@ M.MapMLLayer = L.Layer.extend({
...
+    _setLayerElBounds: function(){
+      var localBounds;
+      let layerTypes = ["_staticTileLayer","_imageLayer","_mapmlvectors","_templatedLayer"];
+      /* jshint ignore:start */
+      layerTypes.forEach((type) =>{
+        if(this[type]){
+          switch(type){
+            case "_templatedLayer":
+              for(let j =0;j<this[type]._templates.length;j++){
+                if(this[type]._templates[j].layer.layerBounds){
```
prushforth, line 1079: If you drag [this layer](https://geogratis.gc.ca/mapml/en/cbmtile/fdi/) onto a map run by this PR, you'll get an undefined error: 
image

ahmadayubi: I added a fix in commit ahmadayubi@3b8331020c9d9f9472afb61eb5188444964853fa for this error. It had to do with `<link rel="query">` and how those templates are handled. Should work now. Good catch.

src/mapml.js
```diff
@@ -598,7 +601,113 @@ window.M = M;
     }
   });
 }());
-M.Util = {
+M.Util = {  
+  extractInputBounds: function(template){
+    if(!template) return undefined;
+    let inputs = template.values, projection = template.projection || FALLBACK_PROJECTION, value = 0, boundsUnit = FALLBACK_CS;
+    let bounds = L.bounds(L.point(0,0),L.point(5,5)), nMinZoom = 0, nMaxZoom = this[projection].options.resolutions.length - 1;
```
prushforth, line 608: This seems odd to have a particular  bounds in here.

ahmadayubi: Good catch, it was supposed to be updated to `...crs.tilematrix.bounds(0);`, it's used as a fallback incase bounds cannot be derived from the template

src/mapml.js
```diff
@@ -2044,6 +2325,23 @@ M.TemplatedFeaturesLayer =  L.Layer.extend({
...
+                        this.layerBounds.overlaps(mapBounds);
+      if(Object.keys(this.layerBounds).length === 0){ //temporary to allow layers with no bounds currently to still work, remove once bounds
+        this.isVisible = true;  //are implemented for all layer types that use templatedTileLayer
+      }
+      if(!(this.isVisible)){
+        this._features.clearLayers();
+        return;
+      }
+      this._onMoveEnd();
+    },
```
prushforth, line 2351: extra comma, maybe need to run jslint /hint

src/mapml.js
```diff
@@ -2428,8 +2732,14 @@ M.TemplatedTileLayer = L.TileLayer.extend({
     initialize: function(template, options) {
       // _setUpTileTemplateVars needs options.crs, not available unless we set
       // options first...
+      let inputData = M.extractInputBounds(template);
+      this.zoomBounds = inputData.zoomBounds;
+      this.layerBounds=inputData.bounds;
+      this.isVisible = true;
+      Object.assign(options,this.zoomBounds);
```
prushforth, line 2743: Instead of using Object.assign there are Leaflet utilties for this, for example on line 2750 we use L.extend(options, {...}) to extend the leaflet options before passing to the initialize call.  We should stick to this style; I think it enhances readability.

src/mapml.js
```diff
@@ -2450,6 +2773,17 @@ M.TemplatedTileLayer = L.TileLayer.extend({
         this._updateOpacity();
       }
     },
+    _setBoundsFlag : function(e){
+      let mapZoom = this._map.getZoom();
+      let mapBounds = M.pixelToPCRSBounds(this._map.getPixelBounds(),mapZoom,this._map.options.projection);
+      this.isVisible = mapZoom <= this.options.maxZoom && mapZoom >= this.options.minZoom && 
+                        this.layerBounds.overlaps(mapBounds);
+      if(Object.keys(this.layerBounds).length === 0){ //temporary to allow layers with no bounds currently to still work, remove once bounds
```
prushforth, line 2785: Is this still required? (Comment indicates maybe not?)

ahmadayubi:
Also closes #66  , seems to be an outdated issue that was fixed somewhere along the process (most likely prior to this PR, but not sure when).

src/mapml.js
```diff
@@ -1425,7 +1671,15 @@ M.MapMLLayer = L.Layer.extend({
...
                       // template has a matching input for every variable reference {varref}
-                      layer._templateVars.push({template:decodeURI(new URL(template, base)), title:title, rel: trel, type: ttype, values: inputs});
+                      layer._templateVars.push({
+                        template:decodeURI(new URL(template, base)), 
+                        title:title, 
+                        rel: trel, 
+                        type: ttype, 
+                        values: inputs, 
+                        zoomBounds:zoomBounds, 
+                        projection:serverExtent.getAttribute("units") || server.getAttribute("content") || "NONE" // TODO: dont think it ever reaches "NONE", need to check
```
prushforth, line 1681: in my IDE the "server" variable is showing global.  Is there something to complete here (TODO?)

src/mapml.js
```diff
@@ -1935,7 +2213,9 @@ M.TemplatedImageLayer =  L.Layer.extend({
             axis = inputs[i].getAttribute("axis"), 
             name = inputs[i].getAttribute("name"), 
             position = inputs[i].getAttribute("position"),
-            select = (inputs[i].tagName.toLowerCase() === "select");
+            select = (inputs[i].tagName.toLowerCase() === "select"),
+            min = inputs[i].getAttribute("min") || min,
```
prushforth, line 2217: Can select have min/max attributes actually?

src/mapml.js
```diff
@@ -3132,53 +3451,58 @@ M.templatedTileLayer = function(template, options) {
...
+                        this._map.getPixelBounds(),
+                        this._map.getZoom(),
+                        this._map.options.projection));
+    
+    if(!(this.isVisible))return; //onMoveEnd still gets fired even when layer is out of bounds??, most likely need to overrride _onMoveEnd
+    this._parentOnMoveEnd();
   },
 
   _isValidTile(coords) {
+    //return true;
```
prushforth, line 3499: could delete?

src/mapml.js
```diff
@@ -3191,48 +3515,51 @@ M.MapMLStaticTileLayer = L.GridLayer.extend({
...
+        layerBounds[sCoords[2]].extend(L.point(pixelCoords.x ,pixelCoords.y ));
+        layerBounds[sCoords[2]].extend(L.point(((pixelCoords.x+TILE_SIZE) ),((pixelCoords.y+TILE_SIZE) )));
       } else{
-        layerBounds[sCoords[2]] = L.bounds(L.point(pixelCoords.x * zoomConstant,pixelCoords.y * zoomConstant),L.point(((pixelCoords.x+256) * zoomConstant),((pixelCoords.y+256) * zoomConstant)));
+        layerBounds[sCoords[2]] = L.bounds(
+                                    L.point(pixelCoords.x ,pixelCoords.y ),
+                                    L.point(((pixelCoords.x+TILE_SIZE) ),((pixelCoords.y+TILE_SIZE) )));
       }
     }
+    // TODO: optimize by removing 2nd loop, add util function to convert point in pixels to point in pcrs, use that instead then this loop
```
prushforth, line 3535: Is this something better recorded as an issue?

src/mapml.js
Malvoz: https://html.spec.whatwg.org/multipage/form-elements.html#the-select-element:
> [Global attributes](https://html.spec.whatwg.org/multipage/dom.html#global-attributes)
> `autocomplete` — Hint for form autofill feature
> `disabled` — Whether the form control is disabled
> `form` — Associates the element with a form element
> `multiple` — Whether to allow multiple values
> `name` — Name of the element to use for form submission and in the form.elements API
> `required` — Whether the control is required for form submission
> `size` — Size of the control

prushforth: I guess we might say that because we're working beyond the pale :-), can `select` have `min/max` attributes even in the MapML world, and I think the answer is still "no", as far as I can see a need at least.

ahmadayubi: The loop goes through all the inputs in an extent including `input type="location"`, this was an artifact from the first attempt to derive the bounds from the given inputs. It will be removed in the next commit as it isn't used anymore.

src/mapml.js
ahmadayubi: Maybe, wdyt? It's referring to a for loop that can be removed by adding more code, I wasn't sure if the small optimization was worth the additional code and I'm still unsure, if you have any suggestions let me know.

src/mapml.js
```diff
@@ -1336,10 +1577,13 @@ M.MapMLLayer = L.Layer.extend({
         }
         function _processInitialExtent(content) {
             var mapml = this.responseXML || content;
+            if(!this.responseXML && this.responseText) mapml = new DOMParser().parseFromString(this.responseText,'text/xml');
             if (this.readyState === this.DONE && mapml.querySelector) {
-                var serverExtent = mapml.querySelector('extent'),
+                var serverExtent = mapml.querySelector('extent') || mapml.querySelector('meta[name=projection]'),
```
prushforth, line 1582: The || here is causing the subsequent application of _synthesizeExtent where there is a `<meta[name=projection]>`, which probably is an error, because we would want to use the *other*  meta elements (as well as this one, granted) to synthesize an extent.  I appreciate that the logic of this section is not really clear; we should discuss and maybe we can clear up that logic so it reads better.

src/mapml.js
```diff
@@ -2505,72 +2829,75 @@ M.TemplatedTileLayer = L.TileLayer.extend({
...
-          break;
-        case 'LINESTRING':
-          coordinates = [];
-          geometry.getElementsByTagName('coordinates')[0].textContent.match(/(\S+ \S+)/gim).forEach(splitCoordinate, coordinates);
-          renderLinestring(this.coordsToPoints(coordinates, 0, tileCoords), geometry);
-          break;
-        case 'MULTILINESTRING':
-          members = geometry.getElementsByTagName('coordinates');
-          for (member=0;member<members.length;member++) {
+      //if(this.isVisible){
```
prushforth, line 2832: odd that this whole block is showing an update but there doesn't appear to be any changes, at least by visual inspection.

I was thinking about rendering and maybe a solution to the rendering issue and this essentially duplicate code will be to subclass L.SVG (if possible) and set our renderer as an option for classes that need to render vectors.  I will look into this a bit, would appreciate your insights.

src/mapml.js
```diff
@@ -3132,53 +3450,57 @@ M.templatedTileLayer = function(template, options) {
...
-  _setBoundsFlag : function(e){
-    let zoomLevel = this._map.getZoom();
+  _handleMoveEnd : function(e){
+    let mapZoom = this._map.getZoom();
+    let zoomLevel = mapZoom;
     zoomLevel = zoomLevel > this.options.maxNativeZoom? this.options.maxNativeZoom: zoomLevel;
     zoomLevel = zoomLevel < this.options.minNativeZoom? this.options.minNativeZoom: zoomLevel;
-    this.isVisible = this._withinBounds(this._map.getPixelBounds(), this._bounds[zoomLevel],this._map.options.crs.options.resolutions, this._map.getZoom());
-    if(!(this.isVisible))return;
-    this.fire('moveend',e,true);
```
prushforth, line 3165: Why is this necessary?  In the code below you invoke the parent moveend handler directly.  Is this a relic, or intentional?

src/mapml.js
```diff
@@ -3132,53 +3450,57 @@ M.templatedTileLayer = function(template, options) {
...
-    this._bounds = this._getLayerBounds(this._groups,this._map.options.crs.options.resolutions); //stores meter values of bounds
-    this.isVisible = this._withinBounds(this._map.getPixelBounds(), this._bounds[this._map.getZoom()],this._map.options.crs.options.resolutions,this._map.getZoom());
+    this._bounds = this._getLayerBounds(this._groups,this._map.options.projection); //stores meter values of bounds
+    this.layerBounds = this._bounds[Object.keys(this._bounds)[0]];
+    for(let key of Object.keys(this._bounds)){
+      this.layerBounds.extend(this._bounds[key].min);
+      this.layerBounds.extend(this._bounds[key].max);
+    }
     L.GridLayer.prototype.onAdd.call(this,this._map);
+    this._map.fire('moveend',true);
```
prushforth, line 3467: Why is this not an  invocation of parent moveend handler like you do below?  Is it a relic?

ahmadayubi: This allows the newly added layer to fire the moveend handlers of the map including the handler of MapMLLayerControl, the layer responsible for disabling the controls in the panel. This allows a layer to be instantly disabled if its selected while out of bounds.

src/mapml.js
```diff
@@ -3270,16 +3595,143 @@ M.MapMLFeatures = L.FeatureGroup.extend({
...
+    _handleMoveEnd : function(){
+      let zoomLevel = this._map.getZoom();
+      zoomLevel = zoomLevel > this.zoomBounds.maxNativeZoom? this.zoomBounds.maxNativeZoom: zoomLevel;
+      zoomLevel = zoomLevel < this.zoomBounds.minNativeZoom? this.zoomBounds.minNativeZoom: zoomLevel;
+      this.isVisible = this._checkZoom() && this._layers && this.layerBounds && 
+                        this.layerBounds.overlaps(
+                          M.pixelToPCRSBounds(
+                            this._map.getPixelBounds(),
+                            this._map.getZoom(),this._map.options.projection));
+      this._removeCSS();
```
prushforth, line 3638: Why do we need to remove css on moveend?

ahmadayubi: That is related to an issue from a few months ago, where we needed to garbage collect styles and stylesheets. It used to be the moveend handler but now since static features get assigned a different moveend handler the old handler still needs to be invoked to garbage collect the styles and stylesheets if there are any.

prushforth: ok

src/mapml.js
ahmadayubi: I believe this was to account for the removal of unutilized `<extent>` (i.e. `<extent units="CBMTILE"></extent>`), prior to this PR some layers contained an extent element with nothing inside, this was replaced with `<meta name="projection" content="CBMTILE">`, so rather than querying for the empty extent to get the projection (formerly `units="CBMTILE"`), those layers now query for the `meta name="projection"` element, a discussion about the logic would be good

src/mapml.js
ahmadayubi: It was indented because I wanted to prevent rendering if the layer's isVisible property was false, I felt that this was the wrong approach as this method should not even been invoked if the layer's isVisible property was false. The indentation will be undone

prushforth:
Let's do this!

PR #108, opened 2020-07-13
Add native appearance to map element(s)

Malvoz:
- [x] Add default size of 300x150px to the map element(s)
- [x] Restore default UA outline styles for accessible sequential navigation
- [x] Enable author controlled `background-color` on the map element(s)
- [x] Pre-style to mitigate FOUC

The new default styles together with the pre-styling have been carefully crafted to ensure that:
- Default styles of the map element(s) are only applied when custom elements is supported (using the [`:defined`](https://developer.mozilla.org/en-US/docs/Web/CSS/:defined) selector), and with a low CSS specificity such that authors can easily override to customize it (e.g. `width`, `height`, `border`, `background-color`).
- Fallback content is only displayed when JS is disabled/custom elements not supported, otherwise they're hidden
(caveat: `` must come after `<map>` in the DOM order to prevent FOUC).
- No need for `<layer- style="display: none">` or [`<map hidden>`](https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/592dc2ae9e9536b676aa4b399c89357a1fe70ec0/index-map-area.html#L24) on every element to prevent FOUC.

@prushforth Please try it out locally (perhaps with added fallback content/inline layers/unsupporting browser and JS disabled to get the full experience 😅) and let me know what you think.

### DIFF
--- a/index-map-area.html
+++ b/index-map-area.html
@@ -6,27 +6,62 @@
     <title>index-map-area.html</title>
     <script type="module" src="dist/web-map.js"></script>
     <style>
-      html {
+      html,
+      body {
         height: 100%;
-      }
-      body,
-      map {
-        height: inherit;
       }
       * {
         margin: 0;
         padding: 0;
       }
+      
+      /* Specifying the `:defined` selector is recommended to style the map
+      element, such that styles don't apply when fallback content is in use
+      (e.g. when scripting is disabled or when custom/built-in elements isn't
+      supported in the browser). */
+      map[is="web-map"]:defined {
+        /* Responsive map. */
+        /* max-width: 100%; */
+        
+        /* Full viewport. */
+        /* width: 100%; */
+        /* height: 100%; */
+        
+        /* Remove default (native-like) border. */
+        /* border: none; */
+      }
+      
+      /* Pre-style to avoid FOUC of inline layer- and fallback content. */
+      map[is="web-map"]:not(:defined) + img[usemap],
+      map[is="web-map"]:not(:defined) > :not(area):not(.web-map) {
+        display: none;
+      }
+      /* Ensure inline layer content is hidden if custom/built-in elements isn't
+      supported, or if javascript is disabled. This needs to be defined separately
+      from the above, because the `:not(:defined)` selector invalidates the entire
+      declaration in browsers that do not support it. */
+      layer- {
+        display: none;
+      }
     </style>
+    <noscript>
+      <style>
+        /* Ensure client-side image map fallbacks are displayed if custom/built-in
+        elements is supported but javascript is disabled. */
+        map[is="web-map"]:not(:defined) + img[usemap] {
+          display: initial;
+        }
+      </style>
+    </noscript>
   </head>
   <body>
-    <img usemap="#dowslake" src="map1.png" width="700" height="400" alt="Dow's Lake area">
-    <map name="dowslake" is="web-map" zoom="17" lat="45.398043" lon="-75.70683" controls hidden>
+    <map name="dowslake" is="web-map" zoom="17" lat="45.398043" lon="-75.70683" controls>
       <layer- id="osm" src="https://geogratis.gc.ca/mapml/en/osmtile/osm/" label="Open Street Map" checked></layer->
       <area is="map-area" id="doughnut" alt="Circle" href="https://example.com/circle/" coords="250,250,25" shape="circle" style="fill: white; stroke: aqua; stroke-width: 5px; fill-opacity: 0;">
       <area is="map-area" id="hole" coords="250,250,7" shape="circle" style="fill: blue; stroke: none; fill-opacity: 0.3;">
       <area is="map-area" id="rect" href="https://example.com/rectangle/" alt="Rectangle" coords="345,290,415,320" shape="rect" style="fill: greenyellow; stroke: blue; stroke-width: 3px; fill-opacity: 0.4;">
       <area is="map-area" id="poly" href="https://example.com/polygon/" alt="Polygon" coords="392,116,430,100,441,128,405,145" shape="poly" style="fill: pink; stroke: blue; stroke-width: 3px; fill-opacity: 0.4;">
     </map>
+    <img usemap="#dowslake" src="map1.png" width="700" height="400" alt="Dow's Lake area">
   </body>
 </html>
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -1,3 +1,11 @@
+.leaflet-container {
+  /* Override the `background-color` set by leaflet.css, enables inheritance from
+     the map or <body> element (same as iframes) to give the author more control. */
+  background-color: transparent;
+  max-height: 100%;
+  max-width: 100%;
+}
+
 /* this is required by tiles which are actually divs with multiple images in them */
 .leaflet-tile img {
         position: absolute;
@@ -41,4 +49,32 @@
 }
 .leaflet-control-layers label {
     display: inline;
+}
+
+.leaflet-control-layers.leaflet-control {
+  margin-right: 10px;
+  margin-left: 10px;
+}
+
+/* Disable dragging of controls. */
+.leaflet-control a {
+  -webkit-user-drag: none;
+}
+
+/* Hide unintended highlighting of controls from clicking the map display in
+   quick succession. This is a workaround for `user-select: contain`, since it has
+   virtually no browser support: https://www.chromestatus.com/feature/5730263904550912. */
+.leaflet-control a::selection,
+.leaflet-popup-close-button::selection,
+.leaflet-control-attribution::selection {
+  background-color: transparent;
+}
+
+/* Restore the default focus outline of UA stylesheets,
+   which Leaflet unfortunately removes (https://github.com/Leaflet/Leaflet/issues/6986). */
+.leaflet-container :focus {
+  outline-color: -webkit-focus-ring-color!important;
+  outline-style: auto!important;
+  outline-width: thin!important;
+  outline: revert!important;
 }
--- a/src/mm-mapp.js
+++ b/src/mm-mapp.js
@@ -89,7 +89,7 @@ export class MmMapp extends HTMLElement {
     `contain: content;` + // Contain layout and paint calculations within the map element.
     `display: inline-block;` + // This together with dimension properties is required so that Leaflet isn't working with a height=0 box by default.
     `overflow: hidden;` + // Make the map element behave and look more like a native element.
-    `height: 150px;` +
+    `height: 150px;` + // Provide a "default object size" (https://github.com/Maps4HTML/HTML-Map-Element/issues/31).
     `width: 300px;` +
     `border-width: 2px;` +
     `border-style: inset;` +
--- a/src/mm-mapp.js
+++ b/src/mm-mapp.js
@@ -72,22 +72,40 @@ export class MmMapp extends HTMLElement {
   constructor() {
     // Always call super first in constructor
     super();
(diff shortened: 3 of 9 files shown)


index-map-area.html
```diff
@@ -6,27 +6,62 @@
...
-      body,
-      map {
-        height: inherit;
-      }
       * {
         margin: 0;
         padding: 0;
       }
+      
+      /* Specifying the `:defined` selector is recommended to style the map
```
prushforth, line 18: I didn't know this, thank you.

index-map-area.html
```diff
@@ -6,27 +6,62 @@
...
+        /* Full viewport. */
+        /* width: 100%; */
+        /* height: 100%; */
+        
+        /* Remove default (native-like) border. */
+        /* border: none; */
+      }
+      
+      /* Pre-style to avoid FOUC of inline layer- and fallback content. */
+      map[is="web-map"]:not(:defined) + img[usemap],
```
prushforth, line 35: I'm a little worried about the complexity of these selectors for users of the polyfill.  I don't recall the gory details at the moment as to why, but I had defined display: none in the connectedCallback of layer.js  for the `<layer->`  with the comment that it avoided displaying inline mapml content but didn't prevent the FOUC; at the moment  I can't really understand what the distinction I was making is.

Malvoz: This CSS is totally optional, it is to "automatically" prevent FOUC (with caveat that `` must come after `<map>` in the DOM order). Defining this in any of the script modules will be too late to prevent FOUC. Of course developers can instead set this themselves, but that leaves it up to each and their own to 1) realize FOUC and 2) come up with styles that prevent FOUC but does not break fallback behavior (which is what we're doing here, _not_ blindly setting client-side image maps to `display: none`).

> I had defined display: none in the connectedCallback of layer.js for the <layer-> with the comment that it avoided displaying inline mapml content but didn't prevent the FOUC

That's still done, only it is now replaced with [this](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/108/files#diff-f073b5f0d60eb22d580b40dd82d5bdf3R98) for `<mm-mapp>` and [this](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/108/files#diff-d6164714af1bf6c21f54c11ca3e1a588R108) for `<map>`, which has the benefit in hiding not only layer elements, but all fallback content (since this is set from the script module, custom elements is supported and we can safely hide any potential fallback child elements).

I suppose, working with custom elements will never be as pretty as native ones 😉

prushforth: > with caveat that  must come after <map> in the DOM order

That I didn't know either.  Do you mean to prevent FOUC, or just generally AND it prevents FOUC.  

> > I had defined display: none in the connectedCallback of layer.js for the  with the comment that it avoided displaying inline mapml content but didn't prevent the FOUC
> 
> That's still done, only it is now replaced with [this](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/108/files#diff-f073b5f0d60eb22d580b40dd82d5bdf3R98) for `<mm-mapp>` and [this](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/108/files#diff-d6164714af1bf6c21f54c11ca3e1a588R108) for `<map>`, which has the benefit in hiding not only layer elements, but all fallback content (since this is set from the script module, custom elements is supported and we can safely hide any potential fallback child elements).

Very nice solution.  This PR is an eye-opener, thank you.  I'll keep reading.  

By the way, I have been noodling about how to actually implement a better polyfill, one that actually contains DOM nodes that the author can interact with.  I was thinking that we could walk the DOM tree and attach the mapml.js layer subtype objects to the appropriate nodes.  The resulting events and so on might make more sense in such a context.  Not sure if I'm describing this right, but imagine you have a  `<layer- ...><extent><input type="location" ...>...<link tref="..." ...></extent></layer->`. Instead of creating an `<extent->` or equivalent `<mapml-extent>` custom element, you just parse the DOM nodes that you're given and attach the appropriate mapml (sub) layer type to the appropriate node.  The mapml layer type continues to render itself in the shadow DOM, but maybe you have a better chance of simulating the behaviour of the node-based existence of the render item.  

We can discuss this outside of this PR, I just wanted to ask you your opinion.  Thanks

src/mm-mapp.js
```diff
@@ -72,22 +72,40 @@ export class MmMapp extends HTMLElement {
   constructor() {
     // Always call super first in constructor
     super();
-    // SUPER IMPORTANT TO SET THIS UP FIRST SO THAT LEAFLET ISN'T WORKING WITH
```
prushforth, line 75: Hi Robert, Can you explain why this isn't as important as I thought?  It points to a fundamental lack of knowledge on my part I think.

src/mm-mapp.js
```diff
@@ -72,22 +72,40 @@ export class MmMapp extends HTMLElement {
...
+    
     let tmpl = document.createElement('template');
     tmpl.innerHTML =
     `<link rel="stylesheet" href="${new URL("leaflet.css", import.meta.url).href}">` +
     `<link rel="stylesheet" href="${new URL("leaflet.fullscreen.css", import.meta.url).href}">` +
     `<link rel="stylesheet" href="${new URL("mapml.css", import.meta.url).href}">`;
+    
     let shadowRoot = this.attachShadow({mode: 'open'});
     this._container = document.createElement('div');
-    // you have to include this otherwise you have to use quirks mode,
```
prushforth, line 85: Worried that this is gone

src/mm-mapp.js
```diff
@@ -72,22 +72,40 @@ export class MmMapp extends HTMLElement {
   constructor() {
     // Always call super first in constructor
     super();
-    // SUPER IMPORTANT TO SET THIS UP FIRST SO THAT LEAFLET ISN'T WORKING WITH
-    // A HEIGHT=0 BOX BY DEFAULT.
```
prushforth, line 76: Worried that this is gone

src/web-map.js
```diff
@@ -76,26 +76,48 @@ export class WebMap extends HTMLMapElement {
   constructor() {
     // Always call super first in constructor
     super();
-    this.style.display = "block";
```
prushforth, line 79: Did this get moved?

src/web-map.js
```diff
@@ -76,26 +76,48 @@ export class WebMap extends HTMLMapElement {
...
+
     let tmpl = document.createElement('template');
     tmpl.innerHTML =
     `<link rel="stylesheet" href="${new URL("leaflet.css", import.meta.url).href}">` +
     `<link rel="stylesheet" href="${new URL("leaflet.fullscreen.css", import.meta.url).href}">` +
     `<link rel="stylesheet" href="${new URL("mapml.css", import.meta.url).href}">`;
 
     const rootDiv = document.createElement('div');
-    // without this you have to omit the doctype, which is bad because
```
prushforth, line 87: Worried about this too

index-map-area.html
Malvoz: > > with caveat that  must come after in the DOM order
> 
> That I didn't know either. Do you mean to prevent FOUC, or just generally AND it prevents FOUC.

_Only_ to prevent FOUC, but we still want to ensure it's hidden if the browser supports custom elements, which you're already doing in (I haven't removed): https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/592dc2ae9e9536b676aa4b399c89357a1fe70ec0/src/web-map.js#L174-L178

src/mapml.css
```diff
@@ -42,3 +50,31 @@
...
+  margin-right: 10px;
+  margin-left: 10px;
+}
+
+/* Disable dragging of controls. */
+.leaflet-control a {
+  -webkit-user-drag: none;
+}
+
+/* Hide unintended highlighting of controls from clicking the map display in
```
prushforth, line 64: wow

src/web-map.js
Malvoz: Yes, moved to [here](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/108/files#diff-d6164714af1bf6c21f54c11ca3e1a588R97). However it is now `display: inline-block` as it aligns better, quite literally, with `<iframe>`/`<video>` if you have them side by side  (they're `display: inline` by default, the closest we can get to native with the map element is `display: inline-block`).

src/mm-mapp.js
Malvoz: This is the same as https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/108#discussion_r453887570.

src/mm-mapp.js
```diff
@@ -72,22 +72,40 @@ export class MmMapp extends HTMLElement {
...
-    this._container.style.height = "100%";
+    
+    // Set default styles for the map element.
+    let mapDefaultCSS = document.createElement('style');
+    mapDefaultCSS.innerHTML =
+    `:host {` +
+    `contain: content;` + // Contain layout and paint calculations within the map element.
+    `display: inline-block;` + // This together with dimension properties is required so that Leaflet isn't working with a height=0 box by default.
+    `overflow: hidden;` + // Make the map element behave and look more like a native element.
+    `height: 150px;` +
```
prushforth, line 92: The dimensions come from an issue in another repo, if I'm recalling correctly.  Maybe we should link to it in the PR.

src/mm-mapp.js
Malvoz: This is the same as https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/108#discussion_r453887570.

src/mm-mapp.js
prushforth: [Here it is](https://github.com/Maps4HTML/HTML-Map-Element/issues/31)

Malvoz: Good point, addressed in https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/108/commits/159394c1cec48edeff914423ae26a3fe10b5e44b.

prushforth:
Robert: Looks good to me!  Thanks for your help.

src/mm-mapp.js
Malvoz: This was setting the map element to a height of 100%, that's unlike any standard element, I believe it also required inheriting height from the body, which is probably not what authors would expect. It does require setting `html { height|min-height: 100%; }` though. Which I think is better, it's not unlike [the behavior of `<div>`s](https://stackoverflow.com/questions/7049875/why-doesnt-height-100-work-to-expand-divs-to-the-screen-height).

src/web-map.js
Malvoz: Same as https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/108#discussion_r453903771.

src/mm-mapp.js
prushforth: > This was setting the map element to a height of 100%, that's unlike any standard element, I believe it also required inheriting height from the body, which is probably not what authors would expect.

Because the map element contains a div that contains a shadowRoot that in turn contains the Leaflet container div, I was trying to get the size of the Leaflet container div set to inherit the size of the map element which in the case of the example I was trying to set to the size of the user's screen.  It makes more sense to have a default size, but I think that still leaves us with using the size of the map element to set the size of the Leaflet container, and to make everything in between the Leaflet container and the map element be 100% of the size of its parent: 

map > div > shadowRoot > div.leaflet-container

Malvoz: I believe this is the current situation, the shadowRoot host element has `display: contents` (i.e. it doesn't participate in box layout):

https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/5f4ebde5b7d2e82fa8468687def65c35ac5b2de4/src/web-map.js#L104-L106

PR #124, opened 2020-07-30
Extends the <map> api and adds a context menu to map which makes use of the api

### LINKED ISSUE #18: Create a right-click / context menu for the map element
Should be included as a [Use Case in the Use Cases and requirements document](https://github.com/Maps4HTML/HTML-Map-Element-UseCases-Requirements/issues/6).

- [x] View Source
- [x] Back button
- [ ] Save as image
- [x] Coordinates
- [ ] [View layer metadata](https://github.com/Maps4HTML/MapML/issues/36)

ahmadayubi:
Closes #18 
menu

Extends the `<map>` element api with the following functions:
- back()
- foward()
- reload()
- viewSource()

Adds keyboard shortcuts for context menu items aswell


src/mapml.css
```diff
@@ -102,7 +106,6 @@
   border-top: 1px solid transparent;
   border-bottom: 1px solid transparent;
   cursor: default;
-  outline: none;
```
Malvoz: Is there a reason not to keep outlines here? This gets overridden:
https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/2edf1b110ab81d7b67c896a8ee7344a3f4ec3dd6/src/mapml.css#L73-L80

src/mapml.js
```diff
@@ -1034,8 +1034,384 @@ M.QueryHandler = L.Handler.extend({
...
+M.ContextMenu = L.Handler.extend({
+  _touchstart: L.Browser.msPointer ? 'MSPointerDown' : L.Browser.pointer ? 'pointerdown' : 'touchstart',
+
+  initialize: function (map) {
+    L.Handler.prototype.initialize.call(this, map);
+
+    //setting the items in the context menu and their callback functions
+    this._items = [
+      {
+        text:"Back (B)",
```
Malvoz, line 1047: These items should preferably use the [`<kbd>`](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-kbd-element) element to denote keyboard user input, for example:
```diff
-text:"Back (B)"
+text:"Back (<kbd>B</kbd>)"
```

src/mapml.js
```diff
@@ -1034,8 +1034,384 @@ M.QueryHandler = L.Handler.extend({
...
+      return this._createSeparator(container, index);
+    }
+
+    var itemCls = 'mapml-contextmenu-item',
+        el = this._insertElementAt('a', itemCls, container, index),
+        callback = this._createEventHandler(el, options.callback, options.context, options.hideOnSelect),
+        html = '';
+
+    el.innerHTML = html + options.text;
+    el.href = '#';
```
Malvoz, line 1180: These are generally button-like (i.e. doesn't initiate navigation); set attribute `role="button"`.

ahmadayubi:
@prushforth 
> I recommend that the copy coordinates include the zoom, and also return the location of the click event (so that it's dynamic (it currently returns the same Lat/Lon every time), you can query the map for the coordinates of a location you right click on). Suggestion: instead of including the words "Latitude" and Longitude" you could include a string "z,lat,lon" or as I prefer "z,lon,lat" as a descriptor of the coordinates returned or something like that. I was thinking that you could have a set of pull out menu choices for different formats and CRS, but perhaps that's out of reach for the moment?

It should be dynamically returning the lat long, can you share what you were doing to get the same lat lng everytime? If it's not dynamic its a bug as it was dynamic for me. As for the format I can update that.

Adding a sub menu is a good idea, I'll look into that.

Malvoz:
```console
web-map.js:495
Uncaught TypeError: Cannot read property 'lat' of undefined
    at HTMLMapElement.forward (web-map.js:495)
    at NewClass._goForward (mapml.js:1137)
    at HTMLAnchorElement.<anonymous> (mapml.js:1241)
    at HTMLAnchorElement.handler (leaflet-src.js:2679)
```

Steps to reproduce:

1. Pan map
2. Reload
3. Forward

prushforth:
@ahmadayubi 

> It should be dynamically returning the lat long

It was user error, probably.  I can't reproduce it today.

prushforth:
@ahmadayubi please include the license text for the context menu code in mapml.js https://github.com/aratcliffe/Leaflet.contextmenu/blob/master/LICENSE.md 

From that document, we're required to do so:

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

src/mapml.js
```diff
@@ -1177,7 +1234,7 @@ M.ContextMenu = L.Handler.extend({
         html = '';
 
     el.innerHTML = html + options.text;
-    el.href = '#';
+    el.role="button";
```
Malvoz, line 1237: The Accessibility Object Model isn't implemented in all browsers (https://wicg.github.io/aom/caniuse.html), perhaps my [comment](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/124#pullrequestreview-458815592) was misleading, we should do `setAttribute('role', 'button')` instead.

src/mapml.css
ahmadayubi: I had to remove the `overflow-y: auto`, it was causing issues with the submenu, not 100% sure but removing the line seemed to fix it, maybe you might know why?

Malvoz: Allowing overflow would enable us to add more options to the menu and not have any of them clipped to the default size of the map element, but I think in most cases developers will increase the default dimensions so probably not an issue.

> not 100% sure but removing the line seemed to fix it, maybe you might know why?

It's not immediately apparent to me why that is, but before digging deeper, there's substantial accessibility issues with the current implementation:

1. The submenu cannot be accessed using a keyboard.
2. The submenu's items come after other primary menu items in the DOM order. So opening a submenu and then tabbing would focus the next primary menu item, as opposed to the first item in that submenu.
3. (Upsteam issue: There's no focus delegation on pressing Shift + F10 to open the context menu, this means that once the context menu has been opened using the keyboard shortcuts, tabbing will focus other map components before the menu items.)

Malvoz: It's unfortunate [`<menu>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu) was abandoned, or that the plugin [doesn't support submenus](https://github.com/aratcliffe/Leaflet.contextmenu/issues/31).

Would it make the implementation easier if we use the [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) element (also used in the layer control) where the `open` attribute could be toggled on hovering the menu items that have submenus? It'd require some additional CSS but I think it's doable.

```html
<div class="mapml-contextmenu">
  <!-- primary items goes here -->
  <a class="mapml-contextmenu-item" href="#" role="button">
    Toggle Controls (<kbd>T</kbd>)
  </a>
  ...
  <!-- primary item with submenu -->
  <details>
    <summary>
      <a class="mapml-contextmenu-item" href="#" role="button">
        Copy Coordinates (<kbd>C</kbd>)
      </a>
    </summary>
    <div class="mapml-contextmenu mapml-submenu">
      <!-- submenu items goes here -->
      <a class="mapml-contextmenu-item" href="#" role="button">
        map
      </a>
      ...
    </div>
  </details>

  <!-- primary items goes here -->
</div>
```

ahmadayubi: Would using `<details>` create a dropdown menu? And if so is it possible to make it a popup submenu instead?

Also in relation to accessibility, you are now able to open the submenu using `C`, you do need to tab through all the elements, although this behavior is common it would be ideal to have it focus the context menu, I tried playing around with tab index but I didn't get the desired behavior, any suggestions?

Malvoz: It should be possible to make it a popup submenu...

<details><summary>I think</summary>

```html
<style>
.mapml-contextmenu {
  position: absolute;
  width: 150px;
  left: 150px;
}

.mapml-contextmenu details[open] .mapml-submenu {
  position: absolute;
  top: 0;
  width: 120px;
  transform: translate(-120px, 0);
}

/* Remove default disclosure widget (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#Customizing_the_disclosure_widget) */
.mapml-contextmenu summary {
  display: inline-block;
  list-style: none;
}
.mapml-contextmenu ::-webkit-details-marker {
  display: none;
}

/* Presentational, ignore these. */
map {
  contain: content;
  display: inline-block;
  overflow: hidden;
  height: 150px;
  width: 300px;
  border-width: 2px;
  border-style: inset;
}
.mapml-contextmenu,
.mapml-submenu {
  border: 2px dotted black;
}
.mapml-contextmenu-item {
  color: #222;
  cursor: default;
  text-decoration: none;
}
</style>
<map>
  <div class="mapml-contextmenu">
    <a class="mapml-contextmenu-item" href="#" role="button">item 1</a>
    <details class="mapml-contextmenu-item has-mapml-submenu">
      <summary>item 2</summary>
      <div class="mapml-submenu">
        <a class="mapml-contextmenu-item" href="#" role="button">item 2.1</a>
      </div>
    </details>
    <a class="mapml-contextmenu-item" href="#" role="button">item 3</a>
  </div>
</map>
```
</details>

`<details>` is just nice because it has built in accessibility. Otherwise, the [WAI-ARIA Authoring Practices](https://www.w3.org/TR/wai-aria-practices-1.2/) has a set of established patterns, such as  [buttons](https://www.w3.org/TR/wai-aria-practices-1.2/#button) (i.e. the menu items `<a role="button">`) and their [recommended keyboard interactions](https://www.w3.org/TR/wai-aria-practices-1.2/#keyboard-interaction-3), and [menu button](https://www.w3.org/TR/wai-aria-practices-1.2/#menubutton) (i.e. the menu items which opens submenus) which might be helpful.

src/web-map.js
```diff
@@ -447,6 +457,65 @@ export class WebMap extends HTMLMapElement {
...
+    let mapEl = this,
+        initialLocation = mapEl._history.shift();
+    mapEl._history = [initialLocation];
+    mapEl._historyIndex = -1;
+    mapEl._traversalCall = true;
+    mapEl.zoomTo(initialLocation.lat,initialLocation.lng,initialLocation.zoom);
+  }
+
+  viewSource(){
+    let mapHTML = this.outerHTML,
```
prushforth, line 508: I think that since we insert the `<style></style>` as a child of the <map> element, we should not show it, or any other DOM nodes that *we* create... it's supposed to be the source *before* script has run, is what I'm saying.

ahmadayubi: That's a good point, I'll add that change

prushforth: Yeah, it's not going to be perfect I think because I believe the attributes are reflecting property changes, but still.

ahmadayubi:
@Malvoz I recently pushed changes in regards to accessibility, the changes are:

- Opening the context menu using keyboard shortcuts sets the focus to the first item in the menu
- Pressing C opens the submenu and focuses the first item in that submenu

Do these changes follow convention and are there any other changes that are crucial to have in terms of accessibility in regards to the contextmenu and submenu?

Malvoz:
@ahmadayubi 

> - Opening the context menu using keyboard shortcuts sets the focus to the first item in the menu
> - Pressing C opens the submenu and focuses the first item in that submenu

Does it work in Chrome for you?

Malvoz:
> any other changes that are crucial

 For `<a role="button">` that controls a submenu:
  - It should _at least_ have [`aria-haspopup="true"`](https://www.w3.org/TR/wai-aria-1.2/#aria-haspopup) (indicates to the user that the element controls the visibility of other content).
  - It'd be optimal if the button also has [`aria-expanded="true|false"`](https://www.w3.org/TR/wai-aria-1.2/#aria-expanded) to convey the current state of the controlled element. I.e. if the submenu is expanded or collapsed.
And has an [`aria-controls`](https://www.w3.org/TR/wai-aria-1.2/#aria-controls), where the value is the ID of the submenu's container element.

Perhaps less important:
- The submenu should be adjacent to the controlling menu item in the DOM. (While delegating focus helps here, if a user is tabbing back after a submenu has been opened, they'll probably expect the element that opened the submenu to be focused.)
- The menu items should be children of list items in an unordered list (the submenu too should be an unordered list). This is so that screen readers can announce how many items are available to the user.

ahmadayubi:
> Does it work in Chrome for you? It doesn't for me (v84.0.4147.105), does in Firefox!

It did work in Chrome for me, it only focuses the first items when the mouse right click isn't used initially, could you specify what part didn't work for you?

Malvoz:
@ahmadayubi this is what I'm doing:

1. Tab until map or map component is focused
2. <kbd>shift</kbd> + <kbd>F10</kbd> to open the contextmenu
3. tab until "Copy coordinates" button recieves focus.
4. Press <kbd>enter</kbd>
5. Tab once: "View map source" has focus 🤷‍♂️

ahmadayubi:
The latest commit adds the following:
>It should at least have aria-haspopup="true" (indicates to the user that the element controls the visibility of other content).

>It'd be optimal if the button also has aria-expanded="true|false" to convey the current state of the controlled element. I.e. if the submenu is expanded or collapsed.
And has an aria-controls, where the value is the ID of the submenu's container element.

>Buttons (all menu items) activate on both enter and space keys (per recommendation: https://www.w3.org/TR/wai-aria-practices-1.2/#keyboard-interaction-3).

>The submenu should be adjacent to the controlling menu item in the DOM. (While delegating focus helps here, if a user is tabbing back after a submenu has been opened, they'll probably expect the element that opened the submenu to be focused.)

And fixes the following issue: 
>Tab until map or map component is focused
>shift + F10 to open the contextmenu
>Tab until "Copy coordinates" button recieves focus
>Press enter: "Copy coordinates" still has focus 🤷‍♂️
>Tab once: "View map source" has focus

Converting from using `<div>` soley to using `<ul> + <li>` proved to be a timely endeavour, and hasn't been added yet, it might be worth opening an issue about it and coming back to it?

Malvoz:
> it might be worth opening an issue about it and coming back to it?

I think so, there are some small things such as conveying the "Toggle controls" pressed state, etc. but this is definitely an improvement and works quite well.

Malvoz:
BTW, the contextmenu currently hijacks the right-click for (e.g.) attribution links!

ahmadayubi:
@Malvoz 
> BTW, the contextmenu currently hijacks the right-click for (e.g.) attribution links!

What are the consequences of hijacking the right-click for attribution links (for example)? I believe prior to adding the context menu right-clicking an attribution link did nothing, is anything lost in hijacking the right-click or is this more of a conventional issue?

Malvoz:
@ahmadayubi it'd be nice if we could preserve the default behavior of those links, for example the ability to right-click and open in a new window.

ahmadayubi:
@Malvoz Prior to the contextmenu you weren't able to right-click the attribution links, the right-click seemed to have been "hijacked" prior to this and could possibly be internal to leaflet? Not sure, but I can do some digging. This addition of the context menu specifically does not remove any abilities that were not already missing, which should help in narrowing down where exactly that ability was lost.

Malvoz:
@ahmadayubi I don't think it's internal to Leaflet though, right-clicking attribution links is possible here: https://leafletjs.com/examples/quick-start/example.html. Since that issue wasn't introduced in this PR, we should track it in a new issue.

prushforth:
> @ahmadayubi I don't think it's internal to Leaflet though, right-clicking attribution links is possible here: https://leafletjs.com/examples/quick-start/example.html. Since that issue wasn't introduced in this PR, we should track it in a new issue.

Probably introduced by my code. Let's merge this PR and open another issue as suggested.

PR #137, opened 2020-08-14
Rename mm-mapp to mapml-viewer

### LINKED ISSUE #104: Rename <mm-app> to <mapml-viewer>
`<mm-mapp>` is hard to remember, and it isn't self-descriptive. I see from the commit messages in https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/91 that the intention was initially `<mm-app>`, which was later supposed to be changed to `<mm-map>`, but was actually changed to `<mm-mapp>` (notably https://geogratis.gc.ca/mapml/ still uses `<mm-app>`).

Can we settle on something more descriptive, that is easier to remember, and maybe more attractive to developers?
Perhaps `<map-embed>` or `<map-viewer>`?

A `<map->` autonomous custom element may also an option(?), but perhaps it's best to have something more distinguishable from `<map>`.

ahmadayubi:
Renames mm-mapp to mapml-viewer and adds a handful of tests related to `<mapml-viewer>`'s latest api additions

Closes #104

### DIFF
--- a/src/mapml-viewer.js
+++ b/src/mapml-viewer.js
@@ -5,7 +5,7 @@ import './mapml.js';       // refactored URI usage, replaced with URL standard
 import './Leaflet.fullscreen.js';
 import { MapLayer } from './layer.js';
 
-export class MmMapp extends HTMLElement {
+export class MapViewer extends HTMLElement {
   static get observedAttributes() {
     return ['lat', 'lon', 'zoom', 'projection', 'width', 'height', 'controls'];
   }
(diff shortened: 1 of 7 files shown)


src/mapml-viewer.js
```diff
@@ -103,7 +103,7 @@ export class MmMapp extends HTMLElement {
     // Set default styles for the map element.
     let mapDefaultCSS = document.createElement('style');
     mapDefaultCSS.innerHTML =
-    `:host {` +
```
Malvoz, line 106: Particular reason for this?

ahmadayubi: Previously in mm-mapp the mapDefaultCSS was added to the shadowRoot, but for web-map mapDefaultCSS was added to the document head.

I thought I'd keep it consistent with web-map and add mapDefaultCSS to document head in mapml-viewer (previously mm-mapp), with this change :host had to be changed to mapml-viewer.

Any concerns about the change?

Malvoz: The inconsistency stems from web-map having a child div as the shadow root (due to limitations) so I couldn't use `:host` there, therefore I had to take the approach of appending to head (as first child to get lower specificity than the author's styles), but I _think_ `:host` is preferable, which was possible for mm-mapp. I believe `:host` is better performance-wise, and this is "critical" CSS for the map element.

Malvoz: IIRC `:host` always has lower specificity (similar to UA styles) than the author's styles, which would guarantee the author never have to fight specificity.

ahmadayubi: Alright good points, just reverted that change

prushforth:
@ahmadayubi would you refactor the file index-mm-mapp.html please.  Other than that I think it's ok now.  Thanks @Malvoz for the review.

PR #138, opened 2020-08-14
Updates for release 0.8.0

prushforth:
Once this is merged, need to tag and release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,30 +1,29 @@
 
-# web-map Customized Built-In &lt;map&gt; Element
+# Customized built-in &lt;map&gt; element
 
-[![Build Status](https://travis-ci.org/prushforth/Web-Map-Custom-Element.svg?branch=master)](https://travis-ci.org/prushforth/Web-Map-Custom-Element) [![Published on webcomponents.org](https://img.shields.io/badge/webcomponents.org-published-blue.svg)](https://www.webcomponents.org/element/Maps4HTML/Web-Map-Custom-Element)
+![Build Status](https://api.travis-ci.com/Maps4HTML/Web-Map-Custom-Element.svg?branch=master)
 
-The Customized Built-In &lt;map&gt; Element is a prototype [implementation](http://maps4html.github.io/Web-Map-Custom-Element/) of the [HTML-Map-Element specification](http://maps4html.github.io/HTML-Map-Element/spec/).
+The customized built-in `<map>` element is a prototype [implementation](http://maps4html.github.io/Web-Map-Custom-Element/)
+of the [HTML-Map-Element specification](http://maps4html.github.io/HTML-Map-Element/spec/).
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -1,30 +1,29 @@
 
-# web-map Customized Built-In &lt;map&gt; Element
+# Customized built-in &lt;map&gt; element
 
-[![Build Status](https://travis-ci.org/prushforth/Web-Map-Custom-Element.svg?branch=master)](https://travis-ci.org/prushforth/Web-Map-Custom-Element) [![Published on webcomponents.org](https://img.shields.io/badge/webcomponents.org-published-blue.svg)](https://www.webcomponents.org/element/Maps4HTML/Web-Map-Custom-Element)
+![Build Status](https://api.travis-ci.com/Maps4HTML/Web-Map-Custom-Element.svg?branch=master)
```
Malvoz, line 4: TBH I didn't know there was a `builds` branch! I don't have much experience with managing repos, and don't fully understand why this is needed. Should this URL point to `branch=master`?

README.md
```diff
@@ -1,30 +1,29 @@
...
 
-The Customized Built-In &lt;map&gt; Element is a prototype [implementation](http://maps4html.github.io/Web-Map-Custom-Element/) of the [HTML-Map-Element specification](http://maps4html.github.io/HTML-Map-Element/spec/).
+The customized built-in `<map>` element is a prototype [implementation](http://maps4html.github.io/Web-Map-Custom-Element/)
+of the [HTML-Map-Element specification](http://maps4html.github.io/HTML-Map-Element/spec/).
```
Malvoz, line 7: Should we care to change this URL to https://maps4html.org/MapML/spec/ here?

README.md
prushforth: So the reason that I think we need a builds branch is so that a user can npm install the package and it doesn't need to be built in order to be useful and also so that people who do npm install it aren't deluged in content that is specific to development and testing.  I thought it would make it simpler for users, but I realize that just getting your system ready to do 'npm install' is actually a big headache, that is not necessary to use the elements.  So having a branch that only contains what you need to run / use the elements is what I was shooting for.  I'm not quite sure I've hit the mark, as I don't know how you download the content of the builds branch by itself except through the npm install process.

prushforth: @ahmadayubi 
> Everything looks and works as it should to this point, I did notice mapml.min.js is empty, would be be valuable to provide at minified version at this time or is it a placeholder for the future?

It's a mistake in the Gruntfile.js.  I will fix it.  Since mapml.min.js is not referenced anywhere, it' probably not that useful, but maybe if someone wants to use the custom elements on their own web site they might find a use for it (performance, download size).

prushforth:
I mistakenly force pushed to this repo's builds branch instead of to my clone from which this PR came. The result is the same, however.  I will tag and release 0.8.0 asap.

PR #139, opened 2020-08-17
Update Gruntfile to put uglify after rollup, comment out concat which is

prushforth:
duplicative to rollup, I think.


ahmadayubi:
What is the c.js file used for?

prushforth:
> What was the c.js file used for?

It was a mistake, probably something that was created in development / testing of the Gruntfile, but was not used.  The concat task is / was supposed to do something like what rollup does, I believe.

PR #141, opened 2020-08-17
Reset inheritable styles

### LINKED ISSUE #140: Should we reset inheritable CSS properties?
(Background/reference: https://developers.google.com/web/fundamentals/web-components/shadowdom#reset)

By default, _some_ (author provided) CSS properties are inherited from body/html elements into the map element, e.g. `text-transform`, `letter-spacing`, `font-size` etc.

This may cause unexpected issues in the map layout or its components. The following CSS is applied in both cases as shown below, whereas in the bottom map elment `:host { all: initial; }` is applied:

```css
html {
  letter-spacing: 5px;
}
```

As such we may want to reset everything before `leaflet.css`/`mapml.css`/other  `:host` styles are applied (result would be that of the bottom map), by setting the [`all`](https://developer.mozilla.org/en-US/docs/Web/CSS/all) property to [`initial`](https://developer.mozilla.org/en-US/docs/Web/CSS/initial_value). Of course, in some cases we may want to explicitly allow inheritance if we find use cases for it by setting `{property}: inherit` in `mapml.css`/`:host`.

Malvoz:
Close #140.

### DIFF
--- a/src/mapml-viewer.js
+++ b/src/mapml-viewer.js
@@ -104,7 +104,7 @@ export class MapViewer extends HTMLElement {
     let mapDefaultCSS = document.createElement('style');
     mapDefaultCSS.innerHTML =
     `:host {` +
-    `all: initial;` + // Reset inherited properties from html/body, as some properties may interfere to the point where they break the layout (https://developers.google.com/web/fundamentals/web-components/shadowdom#reset).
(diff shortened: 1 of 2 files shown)


prushforth:
Thanks! I learn a little bit with each passing commit.

PR #142, opened 2020-08-18
Update index.html and package.json to refer to mapml-viewer.js

prushforth:
Hopefully this works. git procedures a little fuzzy...

### DIFF
--- a/index.html
+++ b/index.html
@@ -4,59 +4,58 @@
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width,initial-scale=1">
     <title>index-map.html</title>
-    <script type="module" src="dist/web-map.js"></script>
-    <style>
-      html,
-      body {
-        height: 100%;
-      }
-      * {
-        margin: 0;
-        padding: 0;
-      }
-      
-      /* Specifying the `:defined` selector is recommended to style the map
-      element, such that styles don't apply when fallback content is in use
-      (e.g. when scripting is disabled or when custom/built-in elements isn't
-      supported in the browser). */
-      map[is="web-map"]:defined {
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -19,21 +19,21 @@
...
         
         /* Remove default (native-like) border. */
         /* border: none; */
       }
       
       /* Pre-style to avoid FOUC of inline layer- and fallback content. */
-      map[is="web-map"]:not(:defined) + img[usemap],
-      map[is="web-map"]:not(:defined) > :not(area):not(.web-map) {
+      mapml-viewer:not(:defined) + img[usemap],
+      mapml-viewer:not(:defined) > :not(area):not(.mapml-viewer) {
```
Malvoz, line 36: These FOUC prevention styles only apply to `<map>`/`<area>`, instead use: https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/739eea7528a95a5a7ad34658e7456c8dc72c38d6/index-mapml-viewer.html#L8-L54

Malvoz: I.e., you can replace everything in `<head>` starting from `<style>` with the above.

prushforth: Thanks Robert, I'll do that

Malvoz: @prushforth there's a fix in https://github.com/prushforth/Web-Map-Custom-Element/pull/5, not sure if this is an unconventional way of collabing.

prushforth: I think it might work, but I corrected my earlier omission of first pulling upstream/master and then applying changes on top of that, so I will see if the checks passed. I also left the map in the non-default size state, because I plan to put the index.html into the builds branch for release / demo / see how this thing works usage.

PR #144, opened 2020-08-18
Adds debug mode to MapML

### LINKED ISSUE #136: Add debug mode to map
Debug mode features needed:
- Tile grid overlay on top of map pane, with z: #, col:# , row# in top left corner of each tile
- Coordinate panel at the bottom left corner of the map that updates on mouse move, displaying the coordinates in tile, tilematrix, map, tcrs, pcrs, gcrs
- Rectangle overlay on each layer to display its layer bounds 

If anyone has any other suggestions please feel free to leave a comment.

ahmadayubi:
Closes #136 

Debug features added: 

- [x] Tile grid overlay on top of map pane, with z: #, col:# , row# in top left corner of each tile
- [x] Coordinate panel at the bottom left corner of the map that updates on mouse move, displaying the coordinates in tile, tilematrix, map, tcrs, pcrs, gcrs
- [x] Rectangle overlay on each layer to display its layer bounds
- [x] Circle to indicate 0,0 of the given projection
- [x] Extends map api with .toggleDebug() to enable or disable debug mode

debug

Tests for debug mode need to be added.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -2,23 +2,22 @@ module.exports = function(grunt) {
 
   grunt.initConfig({
     pkg: grunt.file.readJSON('package.json'),
-    concat: {
-      options: {
-        separator: ';'
-      },
-      dist: {
-        src: ['src/**/a.js;src/**/b.js'],
-        dest: 'dist/c.js'
-      }
-    },
+//    concat: {
+//      options: {
+//        separator: ';'
+//      },
+//      dist: {
+//        src: ['dist/mapml.js'],
+//        dest: 'dist/c.js'
+//      }
+//    },
     uglify: {
       options: {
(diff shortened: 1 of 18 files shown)


ahmadayubi:
Should there be a toggle debug item in the context menu? I see the value in adding it but from my own perspective, I'm not sure if it's a valuable menu item for most users or not. Any opinions?

prushforth:
Good idea.

prushforth:
OK, let's merge this into the debug-layer branch where we can work on it without it getting the the workflow.  The current issue is that Leaflet path features, such as L.Polygon all require coordinates in L.LatLng, so bounding boxes created in projected coordinates can sometimes render quite oddly.  At least we think that's the biggest issue. The layer and map extent properties seem to work for enabling/disabling the layer, so the weird rendering of extents that happens in CBMTILE and others *seems* to be a result of projected extent corners not transforming well to gcrs coordinates (Lat/lon).

PR #151, opened 2020-08-22
Programmatically identifiable map

Malvoz:
Gives the focusable `.leaflet-container` a descriptive accessible name, as well as a role of `region`. See inline comments for the upstream Leaflet issue for more details.

### DIFF
--- a/src/mapml-viewer.js
+++ b/src/mapml-viewer.js
@@ -193,6 +193,11 @@ export class MapViewer extends HTMLElement {
           }
         }
 
+        // Make the Leaflet container element programmatically identifiable
+        // (https://github.com/Leaflet/Leaflet/issues/7193).
+        this._container.setAttribute('role', 'region');
+        this._container.setAttribute('aria-label', 'Interactive map');
+
         this._setUpEvents();
         // this.fire('load', {target: this});
       }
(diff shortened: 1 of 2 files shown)


prushforth:
I certainly endorse the idea of announcing that the `<map>` or `<mapml-viewer>` has a role and/or accessible name that is appropriate.  I wonder if we should not be specifying this in MapML as well.  I would think that `<map>` / `<mapml-viewer>` element should carry these properties?  

I think maybe we need to define/propose a new role such as interactive-map, or more simply "map" (that's not taken by programmers yet, is it?), because maps should be an accessibility tool for the real world.  IIUC then you wouldn't need the aria-role??

Maybe @nchan0154 could comment, too, since I'm way way out of my depth here.  In general I am +1 on this.

nchan0154:
Disclaimer, I am certainly not an expert on the spec side of things, but I agree with your instincts @prushforth. What makes most sense to me is that a native `<map>` element is announced as such (or something similar that would imply the implicit functionality a map has such as zooming and panning). Additionally, because a native map element would not replace all custom web maps, a new role should also be suggested as region is the most fitting right now and that isn't all that descriptive.

Malvoz:
> I think maybe we need to define/propose a new role such as interactive-map, or more simply "map" (that's not taken by programmers yet, is it?)

There's no implicit role of `<map>` today, but that may change, see https://github.com/w3c/aria/issues/698.

@nchan0154 I agree - and thanks for reviewing!

PR #170, opened 2020-09-22
Allows layer deselecting after debug mode is re-enabled

### LINKED ISSUE #161: debug-layer Branch: When toggling debug mode on-off-on, the layers extents don't get removed when layer is unselected
Steps to reproduce:
- Toggle debug mode on 
- Toggle debug mode off
- Toggle debug mode on again
- Deselect a layer in the layer controls panel

Expected:
The deselected layers debug extent should also be removed from the map. Instead a TypeError is thrown in the console.

ahmadayubi:
Lets users enable -> disable -> re-enable debug mode and still have the option of deselecting layers without throwing an error in the console.

Closes #161


prushforth:
Thanks! I will check it out and if I have time, upload it to [here](https://geogratis.gc.ca/mapml/en/) for use by workshoppers.

PR #184, opened 2020-10-09
Add code to support <layer- hidden> attribute

### LINKED ISSUE #183: Fix <layer- hidden> attribute reflection and implementation
While the code existed to update the layer control when the hidden attribute was toggled, it isn't hooked up, and doesn't have any test coverage.

prushforth:
Closes #183


prushforth (commented):
@ahmadayubi I commented out these tests because I couldn't figure out why they were failing.  Seems possible that it's due to inter-test state coupling that changes depending on circumstances that are difficult to predict?  I'll let you have a crack at figuring it out, I'm sure you will be more successful than I was. Thanks!

PR #200, opened 2020-11-03
Adds 'reload' button to the map when controls are toggled on

### LINKED ISSUE #119: Add a home button to pan and zoom map back to initial location
Add a home button above the zoom controls that clears the maps panning and zooming history and returns back to the map's initial location on load.

ahmadayubi:
Adds a reload button that takes the map back to the initial state, closes #119 
Using reload as the label rather than using home is consistent with what the function is called in the context menu.

### DIFF
--- a/src/mapml-viewer.js
+++ b/src/mapml-viewer.js
@@ -201,6 +201,7 @@ export class MapViewer extends HTMLElement {
         if (this.controls) {
           this._layerControl = M.mapMlLayerControl(null,{"collapsed": true, mapEl: this}).addTo(this._map);
           this._zoomControl = L.control.zoom().addTo(this._map);
+          this._reloadButton = M.reloadButton().addTo(this._map);
           if (!this.controlslist.toLowerCase().includes("nofullscreen")) {
             this._fullScreenControl = L.control.fullscreen().addTo(this._map);
           }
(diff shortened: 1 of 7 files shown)


src/mapml/control/Control.Home.js
```diff
@@ -0,0 +1,63 @@
...
+
+  onAdd: function (map) {
+    let container = L.DomUtil.create("div", "mapml-home-button leaflet-bar");
+
+    let link = L.DomUtil.create("a", "mapml-home-button", container);
+    link.innerHTML = "&#x2302";
+    link.href = "#";
+    link.title = "Home";
+    link.setAttribute('role', 'button');
+    link.setAttribute('aria-label', "Home");
```
Malvoz, line 14: I understand that it's probably a convention to name a button with the to functionality reset to the initial location/starting point "home", but in the context of a web map I can see how this could be confusing to users with screen readers, they may interpret this as a button for location their own home. Can we use something else?

prushforth: The intent of the 'home' button (a little 'house' icon is a common visual metaphor these days) is to reset the map to the original zoom / lat / lon **maybe** without changing other conditions (such as opacities, new / added layers etc., tbdiscussed).  What would be an appropriate role/label for such a function?

/cc @shepazu

Malvoz: > What would be an appropriate role/label for such a function?

The best I can think of is "Reset to initial location/view". "location" doesn't make much sense in non-geo maps though.

ahmadayubi: In the context menu its known as reload, would that work?

PR #206, opened 2020-11-10
Convey the disabled state of reloadButton to ATs

Malvoz:
See similar upstream issue: https://github.com/Leaflet/Leaflet/issues/7211.


ahmadayubi (approved):
Looks good, setting the aria-disabled attribute won't break anything, its a seamless addition.

prushforth:
Looks good to me.

Issue #215, opened 2020-12-01
Increase default size of controls


Malvoz:
We should strive to meet the requirements of WCAG [SC 2.5.5 Target Size](https://www.w3.org/TR/WCAG21/#target-size) by making tap targets at least 44x44px (or maybe even 48x48 per Google's web.dev recommendation: https://web.dev/tap-targets/).

While we'd have to consider all tap targets, including (I suppose) context menu items, the default controls would be a good start. Changes would be needed on the following lines:

https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/a3d084d2e3f88c8b314cce7f98c9bbff0fc29709/src/mapml.css#L44-L47

https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/a3d084d2e3f88c8b314cce7f98c9bbff0fc29709/src/mapml.css#L52-L53

https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/a3d084d2e3f88c8b314cce7f98c9bbff0fc29709/src/mapml.css#L102-L109

IMO we should take the opportunity to make the size of the layer control equal to that of the other controls.

Notably, and sadly, none of the reviewed web map tools conform to this success criterion:  
https://github.com/Malvoz/web-maps-wcag-evaluation/blob/master/README.md#255-target-size-level-aaa

Opinions?

shepazu:
This seems like a sensible first step. 

Beyond controls, we need to make sure that tap-target features (e.g. businesses or other geometric areas) also have a minimum size. That could get tricky.

Malvoz:
@shepazu FWIW, there's a Leaflet plugin that tries to deal with that by "making an invisible touch surface around paths, like polygons or polylines, to make it easier to hit them with your clumsy fingers": https://github.com/perliedman/leaflet-touch-helper

prushforth:
For features, could such a magnifier be part of a user stylesheet, or user preferences type of situation?

Malvoz:
Related CSS WG issue to ensuring minimum tap target size for features: https://github.com/w3c/csswg-drafts/issues/4708.

OTOH in maps it's pretty common that features are stacked on top of each other which is problematic for users with a pointer device/touch, a [Feature Index](https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/397) may allow users to interact with stacked features.

Issue #216, opened 2020-12-01
Collapse attribution behind a control


Malvoz:
There are usability issues with the attribution (some of which are mostly apparent on small screens/in small sized maps).

## Problems

Here's a default size map with added `opacity` to the attribution container to demonstrate:

1. Controls may be hidden behind the attribution (means we fail SC [2.4.7 Focus Visible](https://www.w3.org/TR/WCAG21/#focus-visible)):<br>
  ![attribution-z-index](https://user-images.githubusercontent.com/26493779/100800489-f6dcd000-3426-11eb-8ed7-47ded1d99b7a.png)<br>

2. There are no constraints on the element containing the attribution links, which may result in the following depending on the amount of layer sources and/or the map element's size:<br>
  ![attribution-overload](https://user-images.githubusercontent.com/26493779/98679427-4cf6b000-2360-11eb-95a6-002cc900c271.png)<br>

3. Sensory overload
   - Visual
   - Auditory (screen readers announce everything in the attribution on page load because it is visible)

4. Each attribution link adds to the amount of <kbd>Tab</kbd> key presses required to sequentially navigate past the map.

## Potential solution

Collapse attribution behind a control, like [MapBox](https://maps4html.org/HTML-Map-Element-UseCases-Requirements/examples/create-map.html#mapbox-embed) does:

![toggle-attribution](https://user-images.githubusercontent.com/26493779/143277052-1d83909f-10f0-442c-b62a-fe5db5de0a71.png)

Collapsing attribution may have several benefits:

- There's no visible Layout Shift (so does not contribute to the [Cumulative Layout Shift](https://web.dev/cls/)) when populating the attribution container with attributions after fetching layer sources.<br>
  <img src="https://user-images.githubusercontent.com/26493779/127612277-dc724fba-7d41-42b8-97ab-ece05b2350d7.png">

- Improves [Perceived Performance](https://developer.mozilla.org/en-US/docs/Learn/Performance/Perceived_performance). 

- Mitigates sensory overload for some users.

- Less <kbd>Tab</kbd> key presses required to sequentially navigate past the map.

prushforth:
FWIW, the widget is supposed to de-dupe the attribution, so in the case of many layers from the same source, only one attribution entry should be created.  This may not work, but that's the theory.

Malvoz:
> widget is supposed to de-dupe the attribution

That is probably the case, I simply modified the DOM in devtools to present the potential issue.

[...]

PR #223, opened 2020-12-09
Increase default size of primary tap targets

Malvoz:
Fix https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/215. Although there may still be work to be done in this area, that issue is specifically about increasing the size of the default controls so I think it makes sense to close that one out, given this PR.

- [x] Increase default size of controls to 44x44px
- [x] Decrease margin of controls (not needed per se, but increases the size of the interaction area on the map display, especially in small sized maps)

Increasing the size of controls obviously leaves less room for additional controls, but that's a separate issue (https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/224).

### Current
old buttons

### New
new buttons

(albeit with added `noreload` to `controlslist`)

### New: larger map size, no controls excluded
new buttons, big map

### DIFF
--- a/index-mapml-viewer.html
+++ b/index-mapml-viewer.html
@@ -54,7 +54,7 @@
     </noscript>
   </head>
   <body>
-    <mapml-viewer projection="CBMTILE" zoom="2" lat="45" lon="-90" controls controlslist="nofullscreen">
+    <mapml-viewer projection="CBMTILE" zoom="2" lat="45" lon="-90" controls controlslist="nofullscreen noreload">
       <layer- label="CBMT" src="https://geogratis.gc.ca/mapml/en/cbmtile/cbmt/" checked></layer->
     </mapml-viewer>
   </body>
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -38,19 +38,27 @@
 /*
  * Controls.
  */
+ 
+ .leaflet-top .leaflet-control {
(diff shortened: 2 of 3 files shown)


ahmadayubi:
With the increased size of controls the already small map is almost 1/4 covered by controls now. does the small size of the default map have to be as small as a video?

Malvoz:
There was agreement to set the [default object size to 300x150](https://github.com/Maps4HTML/MapML/issues/96#issuecomment-651147927), but technically MapML could specify any such default size.

It's likely more default controls will be added, should we further increase the default size then to make room for those? I don't think increasing the default size solves the problem.

Here's a screenshot of a 300x150 video element in Chrome, it contains 4 controls that are 48x48 px:

video-buttons-chrome 

so I think it's doable with large buttons, although that layout is obviously different.

Malvoz:
https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/216 could also free up some more space btw.

ahmadayubi:
I agree that the controls should show conditionally, I think full screen button should have that behavior, although after adding reload button I'm not sure if that behavior still works properly or not. This PR can be merged without issue though, I can work on conditional controls after I get the current issue I'm working on resolved.

ahmadayubi (approved):

PR #227, opened 2020-12-15
Adds ability to define a custom TCRS that MapML can use

### LINKED ISSUE #220: Design an API to dynamically define a TCRS, which can be correctly referenced by the <map projection="..."> attribute
Implement / prototype response to this [issue](https://github.com/Maps4HTML/HTML-Map-Element-UseCases-Requirements/issues/240) in the UCR

ahmadayubi:
Lets users define a custom TCRS. To do so the user must do as follows:

1.  Define a template using a JSON in the following format:
```json
{
  "projection":"MYPROJ",
  "proj4string": "+proj=utm +zone=33 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs",
  "code": "EPSG:3006",
  "resolutions": [1,2,3,4,5],
  "origin": [0, 0],
  "bounds": [[-34655800, -39000000], [10000000, 39310000]],
  "tilesize": 256
}
```
tilesize must be 256 | 512 | 1024 | 2048 | 4096

2. Define a TCRS and get the projection name back using mapElement.defineCustomProjection(template), pass in the template from step 1 and set the mapElement projection to the custom value:
```js
let mapElement = document.querySelector("mapml-viewer");
let myProjection= mapElement.defineCustomProjection(template);
mapElement.projection = myProjection;
```

A full example of a html file that uses custom TCRS can be found in test/e2e/mapml-viewer/customTCRS.html
Closes #220


test/e2e/mapml-viewer/customTCRS.html
```diff
@@ -0,0 +1,271 @@
...
+            0.18520870375074083,
+            0.11112522225044451,
+            0.066145965625264591
+          ],
+          crs: {
+            tcrs: {
+              horizontal: {
+                name: "x",
+                min: 0,
+                max: zoom => (M.other.options.bounds.getSize().x / M.other.options.resolutions[zoom]).toFixed()
```
prushforth, line 66: This is  not JSON, it's JavaScript; so I was wondering if there's a way we can simplify the format of what we're defining as a parameter i.e. defineCustomProjection(jsonTCRSParametersObject).  Let's discuss  and have a think about this.

ahmadayubi: If a TCRS is defined using JSON there would be certain limits, you can parse most of the data but something like `zoom => (...)` would be impossible to do simply (for security reasons it's not recommended either). Using JS/JS object would let the user utilize the full range of potential in my opinion.

ahmadayubi: You might be able to define something like `zoom => (...)` as an array though but I'm not sure if that makes the format more simple or more complex/tedious.

prushforth:
We discussed this format for a parameter to defineCustomProjection

{
  projection:"MYPROJ",
  proj4string: "+proj=utm +zone=33 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs",
  code: "EPSG:3006",
  resolutions: [...],
  origin: [x,y],
  bounds: [h1, v1, h2 , v2]
  tilesize: **256** | 512 | 1024 | 2048 | 4096
}

prushforth:
Having a bit of difficulty evaluating this PR, because I'm running into problems using the M.defineCustomProjection method due to M not being defined at the moment it is accessed:

```html
<!doctype html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width,initial-scale=1">
    <title>index-map.html</title>
    <script type="module" src="dist/mapml-viewer.js"></script>
    <script>
      let template = `
      {
        "projection":"basic",
        "proj4string":"+proj=utm +zone=33 +ellps=GRS80 +towgs84=0,0,0,0,0,0,0 +units=m +no_defs",
        "code":"EPSG:3006",
        "resolutions":[8192, 4096, 2048, 1024, 512, 256, 128],
        "origin":[0,0],
        "bounds":[[-34655800, -39000000], [10000000, 39310000]],
        "tilesize": 256
      }`;
      let cProjection = M.defineCustomProjection(template);    
      </script>
     <style> 
       html, 
       body { 
         height: 100%; 
       } 
       * { 
         margin: 0; 
         padding: 0; 
       } 
        
       /* Specifying the `:defined` selector is recommended to style the map 
       element, such that styles don't apply when fallback content is in use 
       (e.g. when scripting is disabled or when custom/built-in elements isn't 
       supported in the browser). */ 
       mapml-viewer:defined { 
         /* Responsive map. */ 
          max-width: 100%;  
          
         /* Full viewport. */ 
          width: 100%;  
          height: 100%;  
          
         /* Remove default (native-like) border. */ 
         /* border: none; */ 
       } 
        
       /* Pre-style to avoid FOUC of inline layer- and fallback content. */ 
       mapml-viewer:not(:defined) > * { 
         display: none; 
       } 
       /* Ensure inline layer content is hidden if custom/built-in elements isn't 
       supported, or if javascript is disabled. This needs to be defined separately 
       from the above, because the `:not(:defined)` selector invalidates the entire 
       declaration in browsers that do not support it. */ 
       layer- { 
         display: none; 
       } 
     </style> 
     <noscript> 
       <style> 
         /* Ensure fallback content (children of the map element) is displayed if 
         custom/built-in elements is supported but javascript is disabled. */ 
         mapml-viewer:not(:defined) > :not(layer-) { 
           display: initial; 
         } 
       </style> 
     </noscript> 
  </head>
  <body>
    <mapml-viewer projection="basic" zoom="2" lat="45" lon="-90" controls>
      <layer- label="CBMT" src="https://geogratis.gc.ca/mapml/en/cbmtile/cbmt/" checked></layer->
      <layer- label="Static MapML with tiles" checked>
        <meta name="zoom" content="min=1,max=4">
        <meta name="projection" content="basic">

        <tile zoom="2" row="-23" col="-9" src="data/cbmt/3/c17_r18.png"></tile>
        <tile zoom="2" row="-23" col="-10" src="data/cbmt/3/c16_r18.png"></tile>
      </layer->
    </mapml-viewer>
  </body>
</html>
```

yields:

index_1.html:19 Uncaught ReferenceError: M is not defined
    at index_1.html:19

The odd thing is that M **is** defined after the body loads, but by then it's too late.  Would be nice to not have to wait for events, but maybe I'm not understanding how the module business works.

ahmadayubi:
If you set the type attribute to equal module on your custom script you then get the behavior you want. So `<script type="module">`

prushforth:
> If you set the type attribute to equal module on your custom script you then get the behavior you want. So `<script type="module">`

This is what I get when I do that, if I understand correctly. My impression is that modules are intended to *prevent* you from creating variables in the global scope, and by extension, in the scope of other modules??.  The new TCRS definition should be available to the custom element when it gets defined, I think, so I'm not sure what the right way to do this is....

image

ahmadayubi:
That's strange because the exact same code works fine for me.

ahmadayubi:
Try getting rid of the async in the first script

prushforth:
The error is gone!  No map yet though.

ahmadayubi:
You have to assign the cProjection to a map element:
```js
let map = document,querySelector("mapml-viewer");
map.projection = myProjection;
```

prushforth:
I thought there was a problem with assigning an object to a property that expects a string?

ahmadayubi:
Not a problem really but it just didn't make sense, so now you're assigning a string to it, the cProjection is type string.

prushforth:
Why is the declaration of the mapml-viewer not sufficient:

`<mapml-viewer projection="basic"...`

?

ahmadayubi:
It could be, the only issue I saw with that is if a projection takes longer to calculate/fetch than it does to load mapml-viewer.js you would get errors as the projection is not yet defined. By assigning the projection you let the map know that it's appropriate to start rendering with the custom projection as it's been defined.

prushforth:
It was in discussing the original issue that we noted that the `map.projection` / `map@projection` accepting both objects and strings might be a problem.  I *think* we should support accepting **only** strings via both/either, but I am wary of the property/attribute dichotomy.  Let's say in my example I had left off the attribute; ...seconds pass... the map does construct, and the projection attribute is generated with the value that was assigned to the property.  If, leaving the script to select and set the map.projection property in place, I change the projection on the element to:

`<mapml-viewer projection="CBMTILE"`

the map displays in CBMTILE, no errors.  It just seems to me we should be supporting the attribute approach consistently.  I could be wrong.  It should be perfectly possible to construct and manipulate the element via the DOM though.

ahmadayubi:
Would you be able to explain that comment on Google Hangouts? I'm not fully understanding what you mean.

prushforth:
I wonder if mapml-viewer.defineCustomProjection should also set mapml-viewer.projection, if the process succeeds?  WDYT?

PR #230, opened 2021-01-01
[debug mode] Fix issue with debug grid overlapping the tooltip

Malvoz:
Fix an issue with the debug grid rendering on top of the tooltip by decreasing `mapml-debug-grid`'s `z-index`.

### Before

grid-debug-z-index-before

### After

grid-debug-z-index-after

### DIFF
--- a/src/mapml/layers/DebugLayer.js
+++ b/src/mapml/layers/DebugLayer.js
@@ -28,7 +28,7 @@ export var DebugOverlay = L.Layer.extend({
     this._grid = debugGrid({
       className: "mapml-debug-grid",
       pane: map._panes.mapPane,
-      zIndex: 10000,
+      zIndex: 400,
     });
     map.addLayer(this._grid);
 


src/mapml/layers/DebugLayer.js
```diff
@@ -28,7 +28,7 @@ export var DebugOverlay = L.Layer.extend({
     this._grid = debugGrid({
       className: "mapml-debug-grid",
       pane: map._panes.mapPane,
-      zIndex: 10000,
+      zIndex: 400,
```
Malvoz, line 31: I found that `400` is the lowest `z-index` value that seemingly doesn't visibly break anything _and_ fixes the issue with borders overlapping the tooltip.

ahmadayubi (approved):
400 seems to work great for all the different layer types.

prushforth:
Thanks!

PR #241, opened 2021-01-12
context menu: hide decorative character from ATs

### LINKED ISSUE #240: context menu: decorative character should be hidden from ATs
The decorative character ">" should be hidden from assistive technology.

When focusing the menu item, the [NVDA](https://www.nvaccess.org/) screen reader announces <q>Copy coordinates C greater menu button collapsed submenu</q>.

Decorative character not hidden from screen readers

Malvoz:
Fix #240.


prushforth:
Thanks!

PR #243, opened 2021-01-15
Template Layer ability to default to meta element before projection extent, also allows extent to be defined using different units

### LINKED ISSUE #156: Layers should default to <meta name=extent> before defaulting to projection extent
Currently a MapMLFeatures layer's extent are set using `<meta name="extent">` and if that element is missing, it defaults to the projections bounds/extent.

It was suggested by @prushforth  that the  `<meta name="extent">` should be extended to other layers aswell, to provide a default for templated layers for example.

This addition shows it's utility when a layer has multiple templates, rather than setting a min and max attribute on all the inputs a user can set the extent using a single meta element.

### LINKED ISSUE #242: Allowing TemplatedTileLayer bounds to be defined using units other than tilematrix
It might be useful to have a way of defining the bounds of templated tile layer using units other than tilematrix. Currently you're only able to use tilematrix as such:
```html
<extent units="WGS84">
  <input name="zoomLevel" type="zoom" min="1" max="1" value="0" />
  <input name="row" type="location" axis="row" units="tilematrix" min="0" max="2" />
  <input name="col" type="location" axis="column" units="tilematrix" min="0" max="2" />
  <link rel='tile' type='text/mapml' title='layer' tref='data/wgs84/{zoomLevel}/r{row}_c{col}.mapml' />
</extent>
```

ahmadayubi:
This PR adds a check to extent definition, it allows the inputs min and max attributes to first default to a meta element if one exists before defaulting to the projection extent. With this addition it also allows the min and max attribute to be defined using different units that are then converted to the appropriate type.

For example:
```xml
<mapml>
  <head>
    <title>Canada Base Map - Transportation (CBMT)</title>
    <meta http-equiv="Content-Type" content="text/mapml;projection=CBMTILE"/>
    <meta charset="utf-8"/>
    <base href=""/>
    <link rel="license" href="https://www.nrcan.gc.ca/earth-sciences/geography/topographic-information/free-data-geogratis/licence/17285" title="Canada Base Map © Natural Resources Canada"/>
    <link rel="zoomin" href="/mapml/en/cbmtile/cbmt/?z=18" type="text/mapml"/>
    <meta name="extent" content="zoom=17,top-left-easting=3426966.71,top-left-northing=4244700.70,bottom-right-easting=3171754.82, bottom-right-northing=-781069.77" />
  </head>
  <body>
    <extent units="CBMTILE">
    <input name="z" type="zoom" value="17" min="0" max="17"/>
    <input name="y" type="location" units="tilematrix" axis="row"/>
    <input name="x" type="location" units="tilematrix" axis="column"/>
    <link rel="tile" tref="cbmt/{z}/c{x}_r{y}.png" />
    </extent>
  </body>
</mapml>
```
This will work appropriately even though the meta[name=extent] element content is defined using pcrs.
Closes #156 also closes #242

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3704,9 +3704,9 @@
       "dev": true
     },
     "ini": {
-      "version": "1.3.8",
-      "resolved": "https://registry.npmjs.org/ini/-/ini-1.3.8.tgz",
-      "integrity": "sha512-JV/yugV2uzW5iMRSiZAyDtQd+nxtUnjeLt0acNdw98kKLrvuRVyB80tsREOE7yvGVgalhZ6RNXCmEHkUKBKxew==",
+      "version": "1.3.5",
+      "resolved": "https://registry.npmjs.org/ini/-/ini-1.3.5.tgz",
+      "integrity": "sha512-RZY5huIKCMRWDUqZlEi72f/lmXKMvuszcMBduliQ3nnWbx9X/ZBQO7DijMEYS9EhHBb2qacRUMtC7svLwe0lcw==",
       "dev": true
     },
     "ip-regex": {
--- a/src/mapml/index.js
+++ b/src/mapml/index.js
@@ -573,6 +573,8 @@ window.M = M;
   });
 }());
 
+M.convertPCRSBounds = Util.convertPCRSBounds;
+M.axisToXY = Util.axisToXY;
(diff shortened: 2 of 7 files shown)


ahmadayubi:
There's a strange issue where if the map zoom attribute is out of bounds of the first layer loaded in the map freezes.

Issue #191, opened 2020-10-20
Debug tiles shouldn't use non-standard HTML attributes


Malvoz:
`col`, `row` and `zoom` are invalid (and as such non-standard) attributes on `<div>` elements. Authors may be flagged for using invalid HTML when [validating](https://validator.w3.org/nu/) their web pages that has the component embedded. Use [`data-` attributes](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes) instead?

Malvoz:
May also be motivated by https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/176.

prushforth:
> Authors may be flagged for using invalid HTML when validating their web pages that has the component embedded. 

Doubtful of this; the `<div>`s in question are found in the shadow root, and as such are dynamic, and created by code at run time.  I don't think a validator looks inside the shadow root for a `<video>` controls, for instance.

I'm not saying it would hurt to use data- attributes, just that I don't it's a big deal / internal concern is all.

Malvoz:
I think this is a similar consideration to https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/174. If we create `customattr` today and an attribute with the same name is added to the web platform tomorrow they would conflict (even if in Shadow DOM). It's probably quite unlikely that these particular attributes would ever be added as global attributes (applies to `<div>`) to the HTML spec though.

Malvoz:
There's a related <span title="Technical Architecture Group">TAG</span> issue here: https://github.com/w3ctag/webcomponents-design-guidelines/issues/2.

Issue #96, opened 2020-05-08
[bug] <layer src> should only accept MapML documents


Malvoz:
![Recording-#12](https://user-images.githubusercontent.com/26493779/81450521-e4493380-9182-11ea-814f-c5fb62758128.gif)

prushforth:
Thanks!

PR #249, opened 2021-01-19
Ability to reorder layer in control panel, reflect changes in DOM and map

### LINKED ISSUE #248: Drag and drop behaviour of layers in control
Currently, the [layer control does nothing with drop events](https://geogratis.gc.ca/mapml/en/cbmtile/cbmt/), they are passed on to the map.  So, if you drag a layer in the layer control to another position in the stack, it simply copies the layer on top of the map.  Useful sometimes, but not if you don't intend for it.

It would be nice to be able to re-order layers by dragging their layer control entry up or down.  This would require updating the position of the `<layer->` element in the DOM as a consequence of the event, and preventing the event from further propagation (to the map) and consequent node (`<layer->`).

Your thoughts appreciated below, especially with regard to the keyboard operation of this facility (is that possible even).

ahmadayubi:
This PR allows the user to drag the control elements in order to reorder them, this reordering also changes the `<layer->` elements orders within the `<mapml-viewer>`.

captured (1)

Closes #248

### DIFF
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -354,10 +354,3 @@ summary {
 .leaflet-container .leaflet-control-container {
   visibility: unset!important;
 }
-
-.drag-active {
-  opacity: 0;
-  cursor: grabbing;
-  cursor: -moz-grabbing;
-  cursor: -webkit-grabbing;
-}
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -138,6 +138,7 @@
   border: none;
   margin: 0;
   padding: 0;
+  cursor: grab;
 }
 
 .leaflet-control-layers-overlays > fieldset > .mapml-control-layers > summary,
@@ -277,12 +278,6 @@
 /*
  * User interaction.
  */
-
-/* Disable dragging of controls. */
-.leaflet-control :not([draggable="true"]),
-.mapml-contextmenu :not([draggable="true"]) {
-  -webkit-user-drag: none;
-}
 
 /* Disable text selection in controls. */
 .leaflet-control,
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -356,7 +356,8 @@ summary {
 }
 
 .drag-active {
-  background: transparent;
-  color: transparent;
-  border: 1px solid #4ca1af;
+  opacity: 0;
+  cursor: grabbing;
+  cursor: -moz-grabbing;
+  cursor: -webkit-grabbing;
 }
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -354,3 +354,7 @@ summary {
 .leaflet-container .leaflet-control-container {
   visibility: unset!important;
 }
+
+.mapml-drag-active {
+  cursor: row-resize;
+}
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -355,6 +355,6 @@ summary {
   visibility: unset!important;
 }
 
-.mapml-draggable {
+.mapml-draggable * {
   cursor: row-resize;
 }
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -354,3 +354,9 @@ summary {
 .leaflet-container .leaflet-control-container {
   visibility: unset!important;
 }
+
+.drag-active {
+  background: transparent;
+  color: transparent;
+  border: 1px solid #4ca1af;
+}
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -138,7 +138,6 @@
   border: none;
   margin: 0;
   padding: 0;
-  cursor: grab;
 }
 
 .leaflet-control-layers-overlays > fieldset > .mapml-control-layers > summary,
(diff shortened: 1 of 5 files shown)


src/mapml/layers/MapLayer.js
```diff
@@ -521,19 +521,54 @@ export var MapMLLayer = L.Layer.extend({
...
+        fieldset.ondragend = (e) => {
+          e.target.classList.remove("drag-active");
+          e.target.setAttribute("aria-grabbed", "false");
+          e.target.removeAttribute("aria-dropeffect");
+          let controls = e.target.parentNode.children,
+              layers = map.getPane("overlayPane").children,
+              zIndex = 1;
+          for(let control of controls){
+            let layerEl = control.querySelector("span").layer._layerEl;
+            layerEl.setAttribute("moving","");
```
Malvoz, line 558: `moving` as opposed to e.g. `data-moving` has the same concern as https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/191. Just an FYI (and also to create a reference in that issue, in case we decide to follow the recommendation we know what to change).

src/mapml.css
```diff
@@ -278,12 +279,6 @@
  * User interaction.
  */
 
-/* Disable dragging of controls. */
-.leaflet-control :not([draggable="true"]),
-.mapml-contextmenu :not([draggable="true"]) {
-  -webkit-user-drag: none;
-}
-
```
Malvoz, line 286: Keeping this shouldn't cause any issues as it doesn't target elements that are `draggable="true"`, and I like it, as it prevents dragging of e.g. zoom controls (as showcased in https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/96), at least in some browsers (there's no support data for `-webkit-user-drag` unfortunately). Maybe controls should just default to `draggable="false"` instead?

ahmadayubi: Alright, I'll re-add that snippet, do you know why the cursor isn't the grab cursor? There's a drag-active class I added that should set the cursor to 'grabbing' but it doesn't seem to work, it keeps the 'not-allowed' cursor. If you have any suggestions let me know.

src/mapml/layers/MapLayer.js
ahmadayubi: You suggest using `data-moving` instead? Either works, I just need a way to prevent making unneeded requests when moving the layer element.

Malvoz: Yes, authors are encouraged to used [`data-`](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes) prefixed attributes, this allows the web platform to be extensible by not breaking sites when adding new (non-prefixed) attributes.

src/mapml.css
Malvoz: It seems that `preventDefault()` is needed on the `dragover` event to allow drop, the intended behavior can be seen in [this example](https://developer.mozilla.org/en-US/docs/Web/API/Document/dragover_event#examples).

btw, since we're pulling in leaflet.css we could save a few lines of code by utilizing [`leaflet-grab`](https://github.com/Leaflet/Leaflet/blob/436430db4203a350601e002c8de6a41fae15a4bf/dist/leaflet.css#L211-L215) and not duplicate that CSS in a new `drag-active`. 👍🏼

ahmadayubi: I think I'll still need the `drag-active` as it also changes the opacity of what's being dragged, this makes clear where the dragged element is being moved, but I can get rid of the cursor related CSS in `drag-active` as it is duplicated. Thanks

Malvoz: I just realized Leaflet sets `leaflet-dragging` on the document `<body>`,  so the selector `.leaflet-dragging .leaflet-grab` wont work because that's outside the shadow boundary... I suppose we might as well have a custom class for this then, alternatively set `leaflet-dragging` on some parent element of `leaflet-grab`.

Malvoz: @ahmadayubi I have a fix to allow for using `class="leaflet-grab"` in https://github.com/Malvoz/Web-Map-Custom-Element/commit/c47db9d7a4016607373935132ad2c5e1ea22900b, do you want to incorporate that in this PR, or should I create a separate PR for that? **We'll need that regardless**, because the issue described in https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/249#discussion_r561320063 means that dragging the map display currently doesn't show the `grabbing` cursor (only `grab`).

ahmadayubi: Maybe keep it a separate PR or delay adding it to this one, I'm currently working on using a custom dragging implementation to limit the axis the user can drag on, so the control item can only be moved up and down within the controls overlay rather than anywhere the mouse goes.

Malvoz: Okay I opened a PR, but please ignore my last https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/249#discussion_r561346566, although the PR fixes `cursor: grabbing` for dragging the map display, it relies on `<mapml-viewer class="leaflet-drag-target">`. So going back to using a custom `drag-active` class is probably the easiest way, even if it means duplicating CSS.

Issue #250, opened 2021-01-20
Grabbing cursor not showing


Malvoz:
Leaflet.css uses a `.leaflet-dragging .leaflet-grab` selector to set `cursor: grabbing`, this is currently not working because `leaflet-dragging` is set on the document `body` element (cross shadow boundary).

https://user-images.githubusercontent.com/26493779/105248730-36967d00-5b77-11eb-8b73-2017f7df2566.mp4

prushforth:
Why is leaflet-dragging set on body do you suppose?

Malvoz:
@prushforth https://github.com/Leaflet/Leaflet/issues/2164#issuecomment-27944701 suggests that they're using it to show the grabbing cursor when the mouse pointer goes outside the map view, however that doesn't seem to work when testing here: https://leafletjs.com/examples/quick-start/example.html.


And people continue to report [performance issues](https://github.com/Leaflet/Leaflet/issues/2164#issuecomment-45768405) due to the reflow/repaint triggered by setting/removing the class on the body element. 😕

PR #251, opened 2021-01-20
Fix to allow .leaflet-grab for a "grabbing" cursor

Malvoz:
Fix: #250.

## Preview

https://user-images.githubusercontent.com/26493779/105249264-20d58780-5b78-11eb-890b-4b0e2fa59a5f.mp4


prushforth:
Is it possible to put the css into mapml.css? If not, should this fix be applied to web-map.js as well?

Malvoz:
@prushforth Good catch, these aren't critical for the initial rendering and so should definitely be in mapml.css. Thanks! Fixed in https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/251/commits/35e01e2cde73d8459baad7994a0c0cbfb23ba489. 👍🏼

ahmadayubi (approved):

prushforth:
Thanks!

PR #259, opened 2021-01-26
Add animations to layer control reordering

### LINKED ISSUE #258: Dragging layers in layer control missing fluid animation
When reordering the layers in the layer controls, the items jump positions rather than smoothly dragging.

ahmadayubi:
https://user-images.githubusercontent.com/55214462/105867221-d5c9d300-5fc2-11eb-90e7-8c2f5e5b0de9.mp4

Closes #258

### DIFF
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -358,6 +358,7 @@ summary {
   cursor: pointer;
 }
 
+.mapml-draggable,
 .mapml-draggable * {
   cursor: row-resize;
   background-color: white;
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -370,4 +370,5 @@ summary {
 
 .mapml-draggable * {
   cursor: row-resize;
+  background-color: white;
 }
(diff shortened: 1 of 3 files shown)


ahmadayubi:
@Malvoz you may notice the cursor isn't row-resize when the mouse is over only the `mapml-draggable[aria-grabbed=true]` element. I was wondering if you knew why that is? Shouldn't 
```css
.mapml-draggable * {
  cursor: row-resize;
  background-color: white;
}
```
set the cursor? Or does `pointer-event:none` override that?

prushforth:
I noticed that you can't drag a disabled layer.  It's not a problem, because you can grab other layers and move them around the disabled layer.  Are you happy with this now?  I can merge it when you are satisfied.

ahmadayubi:
I hadn't considered that but I guess if it turns into a user issue we submit a new issue for it then? I think it's good to merge now though.

Issue #247, opened 2021-01-15
Tracking accessibility issues upstream


Malvoz:
Using code from Leaflet (and plugins) means this map component is sometimes inheriting accessibility issues.

We do have comments inline in code where fixes have been made, sometimes with links for upstream issues. Collecting them here in one spot helps get an overview of the state of accessibility in this component. This issue is to track upstream issues (some TBD) for discussion/fixes.

## Map component
- [x] Should be a programmatically identifiable element. Fixed in https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/151/commits/7c86ddcbb8b39328b90930ffa1c5992f891f3df7. Upstream: https://github.com/Leaflet/Leaflet/issues/7193.
- [x] Use legible font-sizes. Fixed in https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/210/commits/f7d705fe793c14fb409d07a3b4d020587af0f709. Upstream: https://github.com/Leaflet/Leaflet/pull/7346.
- [ ] The controls container should come before other author provided content such as popups in the DOM (i.e. `.leaflet-control-container` should come before `.leaflet-map-pane`). This reduces confusion by letting users form a consistent mental model of the content, across different maps (perhaps created by different authors, with different content). Relates to SC [1.3.2 Meaningful Sequence](https://www.w3.org/TR/WCAG21/#meaningful-sequence), [2.4.3 Focus Order](https://www.w3.org/TR/WCAG21/#focus-order) and [3.2.3 Consistent Navigation](https://www.w3.org/TR/WCAG21/#consistent-navigation). Upstream Leaflet: TBD.

## Controls
- [ ] `<a role="button">` should behave as `<button>`. Maybe re-open: https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/19? Upstream: https://github.com/Leaflet/Leaflet/issues/7368 + https://github.com/Leaflet/Leaflet/issues/7369.
- [ ] Fullscreen control should be conveyed as a button, not a link. Upstream: https://github.com/runette/Leaflet.fullscreen/pull/2/ + https://github.com/Leaflet/Leaflet.fullscreen/pull/114.
- [ ] Popup's close button is missing an accessible name. Upstream: https://github.com/Leaflet/Leaflet/issues/7399.
[...]

PR #270, opened 2021-01-29
Adding tabbing navigation and accessibility to map

### LINKED ISSUE #260: Focus outline on map and query target indicator / cross-hairs
@Malvoz I was looking at our mapml-viewer today while tabbing, and I noticed that there is no focus outline on the map when focused.  Should we fix this? 

Also, I set it up so that when you hit the spacebar with the map focused, you issue a query against the center of the map, if there is a query layer in effect.  I was thinking we might reasonably add a cross-hair or some visible indicator of the query location before the query gets issued.  The cross-hair would only show up when the map is focused, perhaps.

Any advice you have about making that facility more accessible would be appreciated.

### LINKED ISSUE #268: Allow features on a map to be tab navigable
It was suggested that the features should be tab navigable by users if they are within the map viewport.

### LINKED ISSUE #280: Feature tabbing bypass blocks
> That's a convention sometimes referred to as "skip links": https://webaim.org/techniques/skipnav/.
> Here's an example: https://genericcomponents.netlify.app/generic-skiplink/demo/index.html.
> 
> It works by using a CSS hack to visually hide the links and unhide them as they recieve focus. There is no standard way to hide something that's intended to be focusable (related discussions, FWIW: https://github.com/w3c/csswg-drafts/issues/560, https://github.com/w3c/css-a11y/issues/13).
> 
> It's indeed helpful, it relates to <span title="Success Criterion">SC</span> [2.4.1 Bypass Blocks](https://www.w3.org/TR/WCAG21/#bypass-blocks). I suppose I can see a standard implementation of it (both as a CSS property and as a default facility in maps) but I think that's also based on assumptions of how users currently interact with maps (and what is keyboard accessible, under which circumstances).

_Originally posted by @Malvoz in https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/264#issuecomment-768606392_

Was discussing with Ahmad, we think having a forward/backward bypass block mechanism built into the (popup) UI would be helpful / essential.  When you tab to a feature, then tab into its (popup) UI, the tab sequence could include a fast forward/backward affordance to get to the end of the feature tab sequence quickly.  So the "normal" tab-through-features sequence would be interruptible by drilling into one feature and activating its bypass block. 

image

ahmadayubi:
https://user-images.githubusercontent.com/55214462/106287470-d35eb780-6214-11eb-934a-cd80c6026e05.mp4

Allow features to be tab navigable and add crosshair to map when query layer exists + map is tab focused

Closes #268 also closes #260 

Edit:
Now adds navigation bypass

https://user-images.githubusercontent.com/55214462/107054722-7d52bc80-679e-11eb-9141-a7be686fde49.mp4

Closes #280


Malvoz:
The features are indistinguishable to someone relying on a screen reader as they don't have an accessible name. `<path tabindex="0">` is announced as <q>graphic</q> in Chrome (<q>blank</q> in Firefox):

<table>
<td width="400">
<video src="https://user-images.githubusercontent.com/26493779/106403524-6c691a80-642f-11eb-8bb8-62b633059214.mp4"></video>
</td></table>

Malvoz:
When using the NVDA screen reader it's not possible to open popups by pressing <kbd>Enter</kbd> or <kbd>Space</kbd> when a feature has focus. This works on other elements such as controls, supposedly because they're standard interactive elements.

prushforth:
> The features are indistinguishable to someone relying on a screen reader as they don't have an accessible name.

I see that the `<path>` element supports a child `<title>` element that provides an accessible name.  If we can come up with a convention for providing a similar accessible name for `<feature>`, we could generate the `<path><title>` from it.

> When using the NVDA screen reader it's not possible to open popups by pressing <kbd>Enter</kbd> or <kbd>Space</kbd> when a feature has focus. 

That's a problem.  Open to suggestions to achieve this. Maybe we should open the popup when the feature receives focus, and dismiss it on blur?

ahmadayubi:
> Maybe we should open the popup when the feature receives focus, and dismiss it on blur?

I think opening a pop up on focus alone might be an annoying interaction to have with a webpage. Adding a title element seems like a good solution based off: https://www.w3.org/TR/SVG11/struct.html#DescriptionAndTitleElements .

prushforth:
> I think opening a pop up on focus alone might be an annoying interaction to have with a webpage.

I'm not saying you're wrong, just want to think about this a bit here.  

When a user clicks or touches a feature when the map isn't in "focus mode", the feature either pops  up a balloon or if there's a query available it pops up a balloon after receiving the response.  Presumably (I am presuming) tab-to-focus a feature is the equivalent operation to touching it for a sighted person, so the user **wants** to hear the properties if they're using a screen reader.

Given the extremely generic properties model that `<feature>` has, I wonder if we should *add* a `<title>` element child of `<properties>` as a special case for accessible names purposes similar to svg.  In the polyfill we can map that to <path><title> or what have you. I'll open an issue on this latter discussion in the MapML and UCR document repos.

prushforth:
I think in the short term, we could generate an accessible name on the <path> element that has the value `<path><title>feature</title></path>`.  I think if a feature isn't provided an accessible name in the future (even if we add `<title>`), we'll still have to generate something that screen readers can announce (they might not need a fake `<title>` if they ever come to support `<feature>`).

Malvoz:
> > I think opening a pop up on focus alone might be an annoying interaction to have with a webpage.
>
> I'm not saying you're wrong, just want to think about this a bit here.
>
> When a user clicks or touches a feature when the map isn't in "focus mode", the feature either pops up a balloon or if there's a query available it pops up a balloon after receiving the response. Presumably (I am presuming) tab-to-focus a feature is the equivalent operation to touching it for a sighted person, so the user wants to hear the properties if they're using a screen reader.

Tab-to-focus is more of a discovery phase (perhaps tedious, depending on the amount of features), where e.g. `<title>` would provide the accessible name for discernibility (discernibility is also an issue present for regular focusable markers!). Touching is equivalent to pressing <kbd>Enter</kbd> or <kbd>Space</kbd> once an element has focus. I agree that popups shouldn't open on focus.

"focus mode" is a term to describe one of few modes in screen readers but I take it you're referring to when the reticle is visible? This mode currently has some usability issues:
1. The user doesn't necessarily know they can press <kbd>Space</kbd> to get details for features. [`aria-keyshortcuts`](https://www.w3.org/TR/wai-aria-1.1/#aria-keyshortcuts) may be useful here, but visual users would need this info on screen. Of course, if this was a standard behavior people would know this by heart.
2. Features that have other features stacked on top of them cannot be interacted with, I believe.
3. Particularly screen reader users:
    - don't know which feature the reticle is on at any given moment (which feature is available to them). Although if features are in the tab order then they can reach them by sequentially navigating instead.
    - don't know that they're actively panning (and in which direction) when pressing the arrow keys as there is no such feedback (but I suppose this is a stand-alone issue, which is also true in regards to zooming).

(Also popups aren't announced when opened. This is an issue with Leaflet popups in general, we're tracking the upstream issue in https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/247.)

Esri's [a11y-map](https://github.com/Esri/a11y-map) (yes I'm referring to them _again_! 😋) doesn't have these issues because each feature within the "scope" is:
1. Announced to the user via ARIA [live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) when panned to.
2. Assigned a keyboard shortcut that can be pressed, to expand details.<br><br><br><br>pressing <kbd>1</kbd> reveals:<br><br>

I think the a11y-map's interaction model is intuitive, it doesn't _need_ to include features in the tab order (unlike regular documents).

Malvoz:
Getting back to popups not opening after pressing <kbd>Enter</kbd> or <kbd>Space</kbd> when a feature has focus (when using a screen reader), [WebAIM says](https://webaim.org/techniques/keyboard/tabindex):

> ### Important
> If `tabindex="0"` is used to make elements focusable, the keyboard interaction must be correct and intuitive for that element. This means that elements that are presented as buttons, for example, must respond to both <kbd>Enter</kbd> and <kbd>Spacebar</kbd> key presses. In some browsers this may require keypress detection with scripting. ARIA roles (such as `role="button"`) and attributes (such as `aria-expanded="true"`), etc. must often be used to inform screen reader users of the function of the navigable element.

Supposedly we're not meeting these criteria, or it is a screen reader bug.

ahmadayubi:
a11y's querying is similar to Google's, and it works well when you have a discrete number of features. In mapml from what I can tell there can be a continuous number of queryable points on the map. There's querying a layer and querying a feature. Querying a layer is continuous but "querying" features is discrete. If you had to use the square model of a11y or google maps you would only show that conditionally as you still have the issue of continuous querying. 

I guess you could introduce a new crosshair + square hybrid? Where numbers query a feature but space queries the center?

Malvoz:
I'm getting a strange bug when opening a popup by pressing <kbd>Enter</kbd>,  _only_ when using NVDA:

https://user-images.githubusercontent.com/26493779/106805628-3c0bc100-6667-11eb-947e-585c9f68525a.mp4

I've never encountered so many oddities when using a screen reader vs not using one as I have in this project. For example, I know that regular markers (``, ugh..) cannot be opened using NVDA, that upstream issue is also tracked in #124 ("marker missing `role="button"`"~).

src/mapml/layers/MapLayer.js
```diff
@@ -94,7 +94,7 @@ export var MapMLLayer = L.Layer.extend({
                 if (properties) {
                   var c = document.createElement('div');
                   c.insertAdjacentHTML('afterbegin', properties.innerHTML);
-                  geometry.bindPopup(c, {autoPan:false});
+                  geometry.bindPopup(c, {autoPan:false, closeButton: false});
```
Malvoz, line 97: Why are close buttons not useful in feature's popups?

prushforth: I think @shepazu mentioned they might be problematic, or it might have been me misinterpreting something...

prushforth:
I think the reticle / graticule could be about 2x as big and bold, personally. I find it easy to not notice.  WDYT?

ahmadayubi:
> I'm getting a strange bug where the popup is misplaced when opening it by pressing Enter, only when using NVDA:

@Malvoz when I use it I don't even get a popup? I put a breaker on leaflet code responsible for creating the popup and it seems with NVDA running that code is never reached, Certain keyboard events might be getting hijacked?

ahmadayubi:
> I think the reticle / graticule could be about 2x as big and bold, personally. I find it easy to not notice. WDYT?

For reference this is what 2x would look like, that's not considering bolding either. It goes from being a map with a UI to being a UI with a map, if that makes sense, but you're right in that it might make seeing it easier. Pros and cons. 
mapoversized

src/mapml/layers/MapLayer.js
ahmadayubi: It was suggested that it be removed, should it be added back?

shepazu: I think having a close button is fine, with the following caveats:
- the popup should not be a modal (that is, it shouldn't be a focus trap)
- if possible, we shouldn't force screen reader users to "close" each feature popup to move on to another feature

This seems like somewhat competing experiences between voice-control and screen-reader users, in exposing the right amount of detail with the fewest possible number of forced interactions. My preference would be to have the popup **not** exposed to screen readers on focus, but only read the accessible name (the popup "title"), with notification of additional information (e.g. `aria-expanded`) to access the other popup details only on demand (at which point they might encounter the close button, but that's fine, because they chose to expand it). For voice-control users, I think having the popup on focus should be good, with scrolling for long details, and a close button.

Does this seem right to you, @Malvoz?

Malvoz: > I think having a close button is fine, with the following caveats:
>   - the popup should not be a modal (that is, it shouldn't be a focus trap)

The popups are (generally*) non-modal, so focus traps are mostly not an issue.

*when the popup for the last feature in the sequence is opened tabbing _is_ currently trapped to some extent (<kbd>shift</kbd> + <kbd>tab</kbd> can be used to move back to the previous feature):

<table>
<td width="250">
<video src="https://user-images.githubusercontent.com/26493779/107206251-1122c380-69ff-11eb-87ab-f30b5d33c7dd.mp4"></video>
</td>
</table>

> if possible, we shouldn't force screen reader users to "close" each feature popup to move on to another feature

Because focus (generally) isn't trapped inside popups users don't have to explicitly close popups before moving on to other features/focusable elements.

> This seems like somewhat competing experiences between voice-control and screen-reader users, in exposing the right amount of detail with the fewest possible number of forced interactions.

I'm not super familiar with voice-control so I'm not sure about how there are competing experiences here. I do know that some people use voice-control to move their cursor around to interact with elements, in that case I think a close button could be appreciated if they just wanted to close a popup without opening a new one.

> My preference would be to have the popup not exposed to screen readers on focus, but only read the accessible name (the popup "title"), with notification of additional information (e.g. `aria-expanded`) to access the other popup details only on demand (at which point they might encounter the close button, but that's fine, because they chose to expand it).

Yes I think we agreed that popups should not be opened (and thus exposed to ATs) when a feature recieves focus, and that each feature itself should be labelled for discernibility. (discussion from https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/270#issuecomment-770489852)



I agree that features that are interactive and have popups should expose the state of the popup using `aria-expanded`.

PR #285, opened 2021-02-09
Allow cycling through features returned from queryable layer.

### LINKED ISSUE #271: query link response without <extent> element doesn't draw
Noticed when updating geoserver code, if I remove the <extent> element from a query response mapml document, the  feature doesn't draw (apparently).

### LINKED ISSUE #282: Allow adding queried features in turn using the "previous" and "next" bypass block buttons
When a query is made it returns mapml features, there can be multiple features. @prushforth suggested that it should be possible to add one of those features at a time by using the "previous" and "next" bypass block buttons.

Behavior example:
If a query returns 4 features, only the first feature is added to the map.
To view the other 3 features, the user needs to click next, this removes the current feature and adds the next one.
Only 1/n features (n being all the features returned) from the query would be visible/on the map at a time. 
Only the `<properties>` content of the current feature i of n (where i is the current feature iteration, n being all features returned) would be the content of the popup.

### LINKED ISSUE #284: iframes for query responses can execute scripting
[CSP Embedded Enforcement](https://w3c.github.io/webappsec-cspee/) (defines the iframe `csp` attribute) [isn't supported by all major browsers](https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement/csp#browser_compatibility), so we fail to prevent scripts from executing in unsupporting browsers. Additionally, scripting is also executed in Chrome due to a [bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1163174) where about:srcdoc fails to inherit the policy.

There are at least 2 ways to go about blocking scripts with wide support:
1. Prepend `srcdoc` with `<meta http-equiv="content-security-policy" content="script-src 'none';">`
2. [`sandbox`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox) the iframes

I think option 1 is more straight forward, and doesn't have any further implications other than blocking scripts, if that's what we want to do.

ahmadayubi:
This PR allows the user to cycle through the different features when a query returns a mapml document. It also updates the popup content to the feature's `<properties>` inner html. 

https://user-images.githubusercontent.com/55214462/107396281-a1c6d580-6acb-11eb-9d7b-7a7b62a2b55e.mp4

Also prepends `<meta http-equiv="content-security-policy" content="script-src 'none';">` to iframe src content as per #284 
Also allows features to be drawn without `<extent>`, I guess it closes #271 
Closes #282 and closes #284


src/mapml/handlers/QueryHandler.js
```diff
@@ -104,74 +104,72 @@ export var QueryHandler = L.Handler.extend({
...
+        });
+        f.addTo(map);
 
-              var c = document.createElement('iframe');
-              c.csp = "script-src 'none'";
-              c.style = "border: none";
-              c.srcdoc = mapmldoc.querySelector('feature properties').innerHTML;
+        let div = L.DomUtil.create("div", "mapml-popup-content"),
+            c = L.DomUtil.create("iframe");
+        c.csp = "script-src 'none'";
```
Malvoz, line 151: I don't think there's a need to set the  `csp` attribute anymore, I suggest removing this line.

src/mapml/handlers/QueryHandler.js
```diff
@@ -104,74 +104,72 @@ export var QueryHandler = L.Handler.extend({
...
-              var c = document.createElement('iframe');
-              c.csp = "script-src 'none'";
-              c.style = "border: none";
-              c.srcdoc = text;
-              layer.bindPopup(c, popupOptions).openPopup(loc);
-          });
+      function handleOtherResponse(text, layer, loc) {
+        let div = L.DomUtil.create("div", "mapml-popup-content"),
+            c = L.DomUtil.create("iframe");
+        c.csp = "script-src 'none'";
```
Malvoz, line 168: I don't think there's a need to set the `csp` attribute anymore, I suggest removing this line.

PR #292, opened 2021-02-12
Follows updated keyboard interaction rules

### LINKED ISSUE #291: Query handling keyboard access
A few issues noticed with the the query handler specifically

- reticle is [too small](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/270#issuecomment-773359216) and can get lost on a busy / large map
- when the focus is on the zoom in button (+) and the user shift-tabs to put focus on the map, the focus indicator / reticle appears, then instantly disappears when the shift key is released, forcing the user to do shift-tab x 2, then tab, to get focus onto the map
- when focus is on the map and the reticle is visible, Enter OR Spacebar should trigger a query, if possible. At least it should be consistent with the tab-through-features key stroke to activate the popup (in that case it's Enter, in this case it's Spacebar; we could make it easy by enabling either keystroke for both scenarios, if that won't result in other problems).
- when the popup appears, focus is on the div surrounding the iframe.  IMHO it should be on the "Next" affordance by default, so that the user can efficiently 'scroll' through all the query results (i.e. or e.g. by repeatedly hitting 'Enter') but is that possible.  We haven't yet dealt with figuring out how to get focus into the contents of the iframe. 
- When leaving a popup, via Esc or otherwise, the focus should return to the map with the reticle displayed, so that they can continue their investigations.

ahmadayubi:
This PR:

- [x] Doubles size of crosshair
- [x] Focuses `+` (Zoom control) instead of parent div when clicking `>|`
- [x] Doesn't hide the crosshair on keypresses as the crosshair is meant to indicate focus 
- [x] `Space` and `Enter` now query top layer
- [x] Pressing `Esc` or closing using the close button now focuses the map afterwards with the crosshair visible
- [x] Tabbing out of query popup focuses controls
- [x] Shift + Tabbing out of query popup focuses map
- [x]  Shift + Tabbing out of feature popup focuses the current feature
- [x] Re-adds close button to features popup 

Closes #291

### DIFF
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -391,9 +391,9 @@ summary {
 }
 
 .mapml-crosshair {
-  margin: -18px 0 0 -18px;
-  width: 36px;
-  height: 36px;
+  margin: -36px 0 0 -36px;
+  width: 72px;
+  height: 72px;
   left: 50%;
   top: 50%;
   content: '';
--- a/src/mapml/handlers/QueryHandler.js
+++ b/src/mapml/handlers/QueryHandler.js
@@ -22,13 +22,17 @@ export var QueryHandler = L.Handler.extend({
         }
     },
     _queryTopLayerAtMapCenter: function (event) {
-        if (event.originalEvent.key === " ") {
+      setTimeout(() => {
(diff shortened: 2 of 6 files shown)


prushforth:
Travis has run out of runway, disabling

PR #302, opened 2021-02-17
Adds outline to map

### LINKED ISSUE #273: Enable focus outline on the map element(s)
Separating the relevant parts from the discussion in https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/260 about enabling the focus outline for map elements:

@prushforth wrote https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/260#issue-794537387:

> @Malvoz I was looking at our mapml-viewer today while tabbing, and I noticed that there is no focus outline on the map when focused. Should we fix this?

@Malvoz wrote https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/260#issuecomment-768128229:
> Showing an outline on `.leaflet-container` on focus would be ideal, but Leaflet tends to do strange things with focus management in general, it produces an inconsistent or "flashing" outline when operating the zoom controls with a mouse pointer:
> 
> https://user-images.githubusercontent.com/26493779/105957381-7975ba80-6079-11eb-9620-576718196ae1.mp4
> 
> A second problem is that some content inside the map such as the attribution container and the map graphics are overlapping the outline.
> 
> >  Should we fix this?
> 
> If we can, yes. It is still a failure of SC [2.4.7 Focus Visible](https://www.w3.org/TR/WCAG21/#focus-visible) even though we've [fixed outlines](https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/24245084ca66488605ad0e7ced1de621fc4d491f/src/mapml.css#L328-L335) for all the other focusable components.
> 
> Edit: One could argue that the issues described above shouldn't dissaude us from displaying a visible focus indicator. Should we just enable it?

@ahmadayubi wrote https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/260#issuecomment-768357363:
> Could [focus-within](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within) be used?

@Malvoz wrote https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/260#issuecomment-768398149:

> @ahmadayubi focus is still triggered the same way, producing the same behavior as `:focus` unfortunately. I also tested the relatively new [`:focus-visible`](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) which can be used to only show outlines when the user is using a keyboard, but it seems the UA heuristics determine a focus outline should be drawn anyway, supposedly due to programmatic focus.

@ahmadayubi wrote https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/260#issuecomment-768405962:

> Ah okay, I recently added the ability to focus vector features so I was thinking `focus-within` could be used to outline the map while the feature was focused as additional feedback to let the user know they are tabbing through the elements within the map.

@Malvoz wrote https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/260#issuecomment-768409870:

> Why isn't outline set to the vector feature directly in that case? That way the user would know which feature as focus. Is it because features outside the map view can be focused?

@ahmadayubi wrote https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/260#issuecomment-768410947:

> No the feature is outlined but it might be helpful to have both the feature focused outlined and the map outlined at the same time. Similar to how google maps outlines the entire map when you are navigating with keyboard and tabs.

@prushforth wrote https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/260#issuecomment-768607093:

> > One could argue that the issues described above shouldn't dissuade us from displaying a visible focus indicator. Should we just enable it?
>
> I think so. I asked Ahmad to put the cross-hair on when the map is focused. We can deal with the focus outline as a separate issue, since we may have to figure out how to make it work better than is currently possible, it seems.

ahmadayubi:
Uses JavaScript to create outline as the outline created using `:focus` flickers, this way the outline is smooth and only shows on keyboard tabbing into and out of the map. 

Edit: The outline color in the video is blue, but has since been changed to black.

https://user-images.githubusercontent.com/55214462/108253576-f00d5180-7127-11eb-9c8a-576e1f63d348.mp4

Closes #273

### DIFF
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -427,3 +427,12 @@ summary {
   text-align: center;
   padding: 2px;
 }
+
+.mapml-outline {
+  outline-style: auto;
+  outline-offset: -1px;
+  outline-color: #44A7CB;
+  z-index: 1000;
+  pointer-events: none;
+  position: relative;
+}
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -431,7 +431,6 @@ summary {
 .mapml-outline {
   outline-style: auto;
   outline-offset: -1px;
-  outline-color: #44A7CB;
   z-index: 1000;
   pointer-events: none;
   position: relative;
--- a/src/mapml/layers/Crosshair.js
+++ b/src/mapml/layers/Crosshair.js
@@ -63,6 +63,20 @@ export var Crosshair = L.Layer.extend({
     }
   },
 
+  _addOrRemoveMapOutline: function (e) {
+    let mapContainer = this._map._container;
+    if (this._map.isFocused && !this._outline) {
+      this._outline = L.DomUtil.create("div", "mapml-outline", mapContainer);
+      this._outline.style.width = mapContainer.style.width;
+      this._outline.style.height = mapContainer.style.height;
+      //mapContainer.style.outlineStyle = "auto";
+      //.mapContainer.style.outlineColor = "#44A7CB";
+    } else if (!this._map.isFocused && this._outline) {
+      L.DomUtil.remove(this._outline);
+      delete this._outline;
+    }
+  },
+
   _hasQueryableLayer: function () {
     let layers = this._map.options.mapEl.layers;
     if (this._map.isFocused) {
@@ -84,6 +98,7 @@ export var Crosshair = L.Layer.extend({
     } else {
       this._map.isFocused = false;
     }
+    this._addOrRemoveMapOutline();
     this._addOrRemoveCrosshair();
   },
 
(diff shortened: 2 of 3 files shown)


src/mapml.css
```diff
@@ -426,4 +426,13 @@ summary {
   white-space: nowrap;
   text-align: center;
   padding: 2px;
+}
+
+.mapml-outline {
+  outline-style: auto;
+  outline-offset: -1px;
+  outline-color: #44A7CB;
```
Malvoz, line 434: I would prefer if this didn't set a particular `outline-color`, to keep the component as unopinionated as possible, but maybe that was that problematic for some reason?

Malvoz (changes requested):

src/mapml.css
ahmadayubi: No just thought the default color (black) looked bad, just a color preference, it's also a similar color Google Map's uses to outline but it can be removed.

Malvoz (approved):

Issue #263, opened 2021-01-27
HTML anatomy of items in the layers list needs refactoring


Malvoz:
## HTML anatomy of a layer item

```html
<fieldset aria-grabbed="false">
    <details class="mapml-control-layers">
        <summary>
            <label>
                <input checked="" type="checkbox" class="leaflet-control-layers-selector" />
                <span style="font-style: normal;"> Canada Base Map - Transportation (CBMT)</span>
            </label>
        </summary>
        <details class="mapml-control-layers">
            <summary>
                <label for="o82">Opacity</label>
            </summary>
            <input id="o82" type="range" min="0" max="1.0" value="1.0" step="0.1" />
        </details>
    </details>
</fieldset>
```

### Nesting interactive elements inside `<summary>` is problematic:

- `<summary>` is not keyboard operable if it has descendant `<label>`/`<input>` _when using a screen reader_. This is a critical issue. I filed NVDA bug: https://github.com/nvaccess/nvda/issues/12008.
- Nesting interactive elements is generally [adviced against](https://github.com/w3c/aria/issues/1285).
- Interactive content in `<summary>` specifically may be disallowed in the future: https://github.com/whatwg/html/issues/2272.
- `<summary>` isn't clickable using a mouse pointer due to nested `<label>`, this required disabling `pointer-events` of `<label>` (using a perhaps not so robust selector) in [mapml.css#L321](https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/15e8157501d47dadc5b473ed8171879b366252ae/src/mapml.css#L321).

Refactoring the layer items is not a small task (it may inadvertently fix #90). First we need to decide on a new layout that does not entail nesting interactive elements in `<summary>`, sigh.

Malvoz:
Legend links are also problematic since they're interactive elements, inside summary.

Malvoz:
We should also ensure that the `<fieldset>` elements are properly labelled, which can be accomplished using [`<legend>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend). However there are limitations with `<legend>` (and `<fieldset>`):

  1. Required to be the first child of `fieldset` (so imposes a certain visual design)
  2. May be problematic if wanting/needing to use CSS flexbox/grid specifically, since the rendering/layout model was only recently revamped (which allows for these `display` styles) https://github.com/whatwg/html/pull/3934.

[...]

PR #306, opened 2021-02-22
Adds a remove layer button to controls

### LINKED ISSUE #253: Allow user to remove layer from (DOM) map / layer control
This is just an idea for discussion. 

It might be useful to allow the user to not just remove a layer from the map (checked), but to remove the layer from the DOM altogether, e.g. click X to delete it from the layer control. It might be confusing though too, so maybe not.  Please add your  thoughts below.

Does such an idea affect requirements, or specifications?

ahmadayubi:
Closes #253

### DIFF
--- a/src/mapml/layers/MapLayer.js
+++ b/src/mapml/layers/MapLayer.js
@@ -494,6 +494,18 @@ export var MapMLLayer = L.Layer.extend({
         opacityControlSummaryLabel = document.createElement('label'),
         mapEl = this._layerEl.parentNode;
 
+        let removeButton = document.createElement('a');
+        removeButton.href = '#';
+        removeButton.role = 'button';
+        removeButton.title = 'Remove Layer';
+        removeButton.innerHTML = '&#10006;';
+        removeButton.classList.add('mapml-context-close-button');
+        L.DomEvent.disableClickPropagation(removeButton);
+        L.DomEvent.on(removeButton, 'click', L.DomEvent.stop);
+        L.DomEvent.on(removeButton, 'click', (e)=>{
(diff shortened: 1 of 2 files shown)


src/mapml/layers/MapLayer.js
```diff
@@ -494,6 +494,18 @@ export var MapMLLayer = L.Layer.extend({
         opacityControlSummaryLabel = document.createElement('label'),
         mapEl = this._layerEl.parentNode;
 
+        let removeButton = document.createElement('a');
+        removeButton.href = '#';
+        removeButton.role = 'button';
+        removeButton.title = 'Remove Layer';
+        removeButton.innerHTML = '&#10006;';
```
Malvoz, line 501: This `✖` character should be hidden from ATs (same as https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/241/). NVDA announces <q>multiply</q> on hover and <q>multiply remove layer, button</q> on keyboard focus. (This should also be done for characters in `.mapml-popup-button`.)

Malvoz (changes requested):

Malvoz:
There is a problem with the layer labels being cut off by the absolutely positioned close button.

Setting a `margin-right` of 15 or 20px on the `<label>` could be a quick fix for now, although it may break if a user changes the font-size in their browser settings (since the icon is an HTML entity and not an img/svg). Otherwise flexbox may be a better solution. We should also try to ensure minimum dimensions of 44px (per https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/215). But I don't think it's worth spending too much time on these things until https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/263 fixed since we'd probably end up rewriting that CSS anyways.

Malvoz (approved):

PR #321, opened 2021-02-25
Check disabled on layer remove, fix re-adding of controls overlay

### LINKED ISSUE #310: Disabled layer items get enabled after toggling and deleting a non-disabled layer item
<table><td width="400"><video src="https://user-images.githubusercontent.com/26493779/109058384-3b54d080-76e3-11eb-9847-6bcf90772cd3.mp4"></video></td></table>

### LINKED ISSUE #318: Layer control reappears after toggling controls off and on after all layers have been removed
<table><td width="450"><video src="https://user-images.githubusercontent.com/26493779/109156599-50784080-7771-11eb-8c3d-86d3e8c1bfa8.mp4"></video></tr></table>

ahmadayubi:
https://user-images.githubusercontent.com/55214462/109171607-f31fc780-774f-11eb-8559-7f4bc5a360ea.mp4

Closes #310  and closes #318

### DIFF
--- a/src/mapml-viewer.js
+++ b/src/mapml-viewer.js
@@ -254,7 +254,7 @@ export class MapViewer extends HTMLElement {
         }
       }
 
-      if (!this.controlslist.toLowerCase().includes("nolayer") && !this._layerControl){
+      if (!this.controlslist.toLowerCase().includes("nolayer") && !this._layerControl && this.layers.length > 0){
         this._layerControl = M.mapMlLayerControl(null,{"collapsed": true, mapEl: this}).addTo(this._map);
         //if this is the initial setup the layers dont need to be readded, causes issues if they are
         if(!setup){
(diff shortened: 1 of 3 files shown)


Malvoz:
This is great @ahmadayubi, though it seems to me that the Remove Layer button shouldn't be clickable if `fieldset` is `disabled` per https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fe-disabled (quoting the directly relevant parts):

> A form control is **disabled** if any of the following conditions are met:

> 2. The element is a descendant of a **fieldset** element whose **disabled** attribute is specified

> A form control that is **disabled** must prevent any `click` events that are queued on the user interaction task source from being dispatched on the element.

But, if it's important that layers can be removed even if a layer item (`<fieldset>` as currently implemented) is `disabled` then perhaps that should be a consideration (for later) when refactoring the layer items' anatomy in https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/263.

Malvoz (approved):

Malvoz:
Maybe I'm being a bit too strict about conforming to the HTML spec's behavior in this matter.

PR #322, opened 2021-02-25
Add aria-expanded to features

### LINKED ISSUE #317: Features that bind to popups should have appropriate ARIA properties
There's currently nothing indicating to users of assistive technology that a feature can be interacted with.

Features (rendered as `<path tabindex="0">`) that bind to popups should:

- convey the `aria-expanded="true|false"` state of the corresponding popup (as suggested by @shepazu)
- potentially have `role="button"`

ahmadayubi:
Closes #317

### DIFF
--- a/src/mapml/layers/FeatureLayer.js
+++ b/src/mapml/layers/FeatureLayer.js
@@ -80,6 +80,7 @@ export var MapMLFeatures = L.FeatureGroup.extend({
               title.innerText = "Feature";
               path._path.appendChild(title);
             }
+            path._path.setAttribute("aria-expanded", "false");
           }
         }
       }
--- a/src/mapml/layers/MapLayer.js
+++ b/src/mapml/layers/MapLayer.js
@@ -1265,6 +1265,7 @@ export var MapMLLayer = L.Layer.extend({
       if(path) {
         // e.target = this._map
         // Looks for keydown, more specifically tab and shift tab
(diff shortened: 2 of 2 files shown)


Malvoz:
I just tested this using NVDA in Chrome, `aria-expanded` isn't conveyed unless `role="button"` (for example) is used.

ahmadayubi:
Should paths have `role="button"`? @prushforth. It sounds a little odd but I guess they do behave as a button would.

Malvoz:
It's indeed a bit strange that features are exposed as "merely" buttons, perhaps [`aria-roledescription`](https://www.w3.org/TR/wai-aria-1.1/#aria-roledescription) with a value of "feature" can be used to express that this is something more than a button? Maybe @shepazu has some thoughts on this.

prushforth:
> Should paths have `role="button"`? @prushforth. It sounds a little odd but I guess they do behave as a button would.

I don't believe so.  I hope they'll have role=feature one day, but they certainly don't have button semantics.

Malvoz:
If `role="feature"` was the implicit role of `<feature>` and browsers and ATs supported MapML then this polyfill wouldn't be in use and because `role="button"` is seemingly required (at least for NVDA, which is one of the most popular screen readers) to expose `aria-expanded` I think there's no reason to not add `role="button"`.

PR #328, opened 2021-03-01
Add tooltip on focus + hover

### LINKED ISSUE #327: Add "tooltip" display on focus / hover event
In testing the first [experiment](https://maps4html.org/experiments/index.html) with @shepazu, we found that (due to my misunderstanding earlier comments) the user experience is not great for sighted keyboard users, since there is no display of the feature's accessible name on focus.   The default <title>Feature</title> is being generated in paths today (which if features existed in HTML would not be necessary because of a default that we'll establish), and is presumably read by screen readers (I tried to download NVDA but their download is out of order at the moment). 

In reviewing the svg [`<title>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title) element, it looks like it would be reasonable to add `<title>` as an optional child of `<feature>`, which would have the `<properties>` and `<geometry>` siblings.   

I added a [feature request](https://github.com/Maps4HTML/geoserver/issues/27) to the [GeoServer MapML extension](https://github.com/Maps4HTML/geoserver) to allow the user to map content to the generated `<feature><title>` so that there is a meaningful experience available there.

Aside: `<properties>`  is starting to look a lot like svg's [`<desc>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/desc), which is OK; we can have the conversation about that element's name in a working group.  "properties" comes from GeoJSON and is meant to be familiar to developers familiar with that format.

ahmadayubi:
https://user-images.githubusercontent.com/55214462/109526912-1a90d000-7a81-11eb-9987-c1a2dc741ba5.mp4

The first square is created using:
```html
<feature zoom="2" class="refDiff">
  <featurecaption>Accessible Square</featurecaption>
  <properties>
    <h1>Test</h1>
    <a href="www.example.com">test</a>
  </properties>
  <geometry cs="tilematrix">
    <polygon>
      <coordinates>11 11 12 11 12 12 11 12</coordinates>
    </polygon>
  </geometry>
</feature>
```

Closes #327

### DIFF
--- a/.github/workflows/sync.yml
+++ b/.github/workflows/sync.yml
@@ -43,3 +43,15 @@ jobs:
         user_name: 'ahmadayubi'
         commit_msg: '[AUTO] Sync MapML Build'
         destination_branch: master
+    - name: Pushes to experiments repo
+      uses: ./.github/actions/sync-repo
+      env:
+        API_TOKEN_GITHUB: ${{ secrets.API_TOKEN_GITHUB }}
+      with:
+        source_folder: 'dist'
+        destination_repo: 'Maps4HTML/experiments'
+        destination_folder: 'static/dist'
+        user_email: aayub041@uottawa.ca
+        user_name: 'ahmadayubi'
+        commit_msg: '[AUTO] Sync MapML Build'
+        destination_branch: main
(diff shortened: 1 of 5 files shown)


prushforth:
I believe the `<title>` only is presented if it is found in inline layer DOM, not found in `<layer- src=...>` layers. Let's wait until we've figured out what this element will be called during upcoming meeting hopefully.

prushforth:
Darn I just noticed that when you use the ">" or "<" affordances you don't get the tooltip when focus goes to the next/previous feature. Sorry I should have checked more thoroughly.

Issue #299, opened 2021-02-15
Implement the "Leaflet API" on top of <mapml-viewer> and <layer-> etc


prushforth:
This is more of a thought experiment than an actual implementation idea.  

What would it take to run Leaflet plugins if the Web browser actually implemented a `<map>` and [<layer>](url) (etc) element set? 

It seems reasonable that simple maps would be simple using declarative HTML, but could these browser mapping facilities be used to underpin (implement) a script library that provided backwards compatibility with the Leaflet 1.x API, such that [plugins](https://leafletjs.com/plugins.html) would run without modification?

The idea comes from the notion of jQuery, which is no longer needed in a browser, because element.querySelector or querySelectorAll are natively implemented.  But if your code uses jQuery, jQuery itself could be shimmed, with the underlying functionality provided by the native implementation.  Of course it would be smarter to just remove dependence on jQuery, but maybe some abstractions that are implemented by Leaflet would become conspicuous by our inability to provide them easily.  Those abstractions might be necessary, but they also might be non-low hanging fruit, which would necessarily be progressive enhancements (script-only).

Malvoz:
For reference (and correct me if I'm wrong), I think this describes one of the [Alternative approaches](https://github.com/Maps4HTML/MapML-Proposal#alternative-approaches) in the MapML proposal:

> APIs: Leaflet, OpenLayers and others, (albeit others without any notion of cross-origin resource sharing) provide excellent map scripting APIs and events. Can these or similar APIs be built on top of the proposed HTML infrastructure? Would life be simpler for authors with the proposed HTML?

Without giving this too much thought I think my opinion is that this is not worth it.

The time (and "_pie_", as Brian calls it) invested by implementers into - essentially - supporting old plugins, could be spent on better things. Not to mention the added complexity of doing so.

Plugins go out of date quickly, and are generally inaccessible to users (I've personally raised 20+ accessibility-related issues in 15 of the most popular Leaflet plugins. I think 1 has been resolved in 6 months~). Many plugins even suffer from not being compatible with the latest Leaflet version. Heck, even Leaflet is unmaintained nowadays.

PR #334, opened 2021-03-02
Adds SVG marker

### LINKED ISSUE #307: Double check point features, might be referencing incorrect directory for marker PNG
The viewer keeps the leaflet markers in an images folder in dist, but when a point is created it may be referencing the incorrect directory internally. Resulting in a 404 error when trying to load in the marker png.

### LINKED ISSUE #325: Tabbing through features has a hiccup - skip block gets stuck between set of polygon and point features
Steps to reproduce

Visit this [experiment](https://maps4html.org/experiments/index.html), which has a set of features in the DOM. Press Tab twice, the first polygon feature (quite small) should be focused. Hit Enter.  Hit Tab 3x to focus the ">" affordance. Hit Enter.  The second polygon feature should be focused. Hit Enter, then Tab 3x, then Enter. The third polygon feature should be focused. Hit Tab 3x, then Enter.  The focus **should** go to the fourth feature (a point) but it does not, it stays on the Prescott Hotel, a very fine Italian eatery and tavern.  You need to hit Tab to get to the fourth feature.

Edit: In playing with the ">" affordance more, when the set of features in focus contains points, the "<" and ">" don't seem to focus the next or previous feature reliably, you always need to add a Tab or Shift-Tab to get focus to where you want.

ahmadayubi:
https://user-images.githubusercontent.com/55214462/109718052-d7178e00-7b74-11eb-9814-3b76111c271c.mp4

Closes #325 also closes #307


Malvoz (approved):

Malvoz:
Are Leaflet markers expected to be used at all after this change? for example to support https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/299?

Otherwise I think this also closes https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/307, and we can also remove markers from:

https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/5a21d3a94872e5f4d7fed48a7326574912358aae/Gruntfile.js#L131

ahmadayubi:
That's a good point, those can now be removed.

PR #341, opened 2021-03-04
Changes to feature interaction

### LINKED ISSUE #337: Change popup behaviour so that the map pans when the popup extends past the viewport.
Mentioned as a problem in #335.  Also, if possible via Leaflet popup options, make the popup stay on screen until dismissed (don't disappear on pan).

### LINKED ISSUE #338: Set geometry colour to strong, bold value when focus is on it.
Mentioned in #335.

### LINKED ISSUE #339: Suggest using aria-label attribute, not SVG <title> element
From #335: Map the `<feature><featurecaption>` to `<path aria-label>` instead of `<path><title>` to avoid native tooltip presentation issue

ahmadayubi:
https://user-images.githubusercontent.com/55214462/109994030-e7497d80-7cda-11eb-8eb8-f7848998793e.mp4

Closes #337 and closes #338  and closes #339

### DIFF
--- a/src/mapml/handlers/QueryHandler.js
+++ b/src/mapml/handlers/QueryHandler.js
@@ -48,7 +48,7 @@ export var QueryHandler = L.Handler.extend({
           crs = layer.crs,
           tileSize = map.options.crs.options.crs.tile.bounds.max.x,
           container = layer._container,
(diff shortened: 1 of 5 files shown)


Malvoz (approved):

prushforth:
Good work, looks great.

PR #353, opened 2021-03-15
Consolidate feature rendering between MapMLFeatures and Templated Tile Layer

### LINKED ISSUE #360: Transform query response <featurecaption>value</featurecaption> to <path aria-label="value" ...></path>
I have updated [GeoServer](https://github.com/Maps4HTML/geoserver/commit/3d971a3450312723d736cb8314202a38b21e9edc)to enable the serialization of `<featurecaption>` when querying the service, so we should update the viewer to use the new feature renderer which supports this transformation.

### LINKED ISSUE #361: Error during processing of <feature> with no <geometry> element in query response
To reproduce, get a running copy of [GeoServer](https://build.geoserver.org/geoserver/main/geoserver-main-latest-bin.zip), with the MapML [extension](https://build.geoserver.org/geoserver/main/ext-latest/geoserver-2.20-SNAPSHOT-mapml-plugin.zip) [installed](https://docs.geoserver.org/latest/en/user/extensions/mapml/index.html#installation).  (At the time of writing this, those links don't work.  Hopefully just a weekend problem???)

Open the Layer Preview page, click on the "MapML" link in the row for the "mosaic" imagery demo.

Open browser console

Click on / query the image.  You'll see this is in the console:

image
image

It's legit that features have null geometry, such as with a query response to an image.  I think we can / should use the click location and generate a (temporary) point geometry there.  I think that was the old behavior, at least.

ahmadayubi:
This PR introduces a new Feature type that takes MapML and generates the appropriate feature points, outlines, etc. It also adds a new FeatureRenderer which takes those features and their internal structure and sets the appropriate d attribute on the paths.

Summary:
- [x] `<span>` functionality for static features and templated features
- [x] Preserves `<span>` functionality for templated tile layer
- [x] Reduces duplicate code
- [x] Adds a new `data-fid` attribute on paths that takes the value of `<feature>` id 
- [x] Enables `<geometrycollection>` 
- [x] Allows templated tile layer features to be defined in any coordinate system (might have already existed, not sure)
- [x] M.Features now has a built in SVG marker, enabling the removal of M.SVGMarker

Also closes #360 and closes #361


prushforth:
Pretty impressive, I will have a closer look tonight. Getting out for a walk while the sun shines, suggest you do likewise 😎

prushforth:
Probably on the to-do list, but fwiw: `multi ( point | line | polygon)` are treated like `{point | line |  polygon) `as far as focus goes. I think they should still be a single entity as far as tabbing goes.  For example, Hawaii or Alaska in us_pop_density.mapml

prushforth:
Looks like the svg is not in the layer container, that might be affecting rendering etc (?):

image

prushforth:
Seem to be "ghost points" which generate markers when you markup a polygon like so:

```html
<polygon>
  <coordinates><span class="test">-101.812942 36.501861 -100.000075 36.501861 -100.000075 34.563024 -99.923398 34.573978 -99.698843 34.382285 -99.57835 -101.812942 36.501861</span></coordinates>
</polygon>
```
you get a "ghost marker": 
image

ahmadayubi:
@prushforth I'll leave the tab focusing to another issue as this one mainly focuses on just rendering, as for that svg I believe that's either the debug layer or the crosshair, I don't believe those are "features", it exists in the current version of mapml-viewer, I believe. The ghost point is odd though, good digging.

ahmadayubi:
> Probably on the to-do list, but fwiw: `multi ( point | line | polygon)` are treated like `{point | line | polygon) `as far as focus goes. I think they should still be a single entity as far as tabbing goes. For example, Hawaii or Alaska in us_pop_density.mapml

Should geometry collection be treated the same way or should you be able to tab through the parts of that?

prushforth:
Geometry Collection should be treated as a single object like any other geometry.

prushforth:
I noticed that the `<featurecaption>` value, while being used for the tooltip, is not copied to the `aria-label` value which remains at the default of `Feature` (see picture).

image

prushforth:
Also, I noticed that the stylesheets for templatedTileLayer aren't being picked up:

master:
image


vs customRenderer:

image

prushforth:
btw I believe the `<svg>` in this [comment](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/353#issuecomment-799869423) is actually the debug layer (you may have said that originally).  It might be an idea to tag it with a class or otherwise identify it (make a pane for it??) ?

ahmadayubi:
Could the style sheet issue be related to how the css is defined? Maybe the new structure affects the selectors because the stylesheets and styles are being added to the containers from what I can see.

ahmadayubi:
style
feature

Do these align? I'm not too familiar with css.

prushforth:
@ahmadayubi IIRC the "." is illegal in class names, so I vaguely recall replacing it with "_" in some code.  Maybe that's the problem?

ahmadayubi:
> @ahmadayubi IIRC the "." is illegal in class names, so I vaguely recall replacing it with "_" in some code. Maybe that's the problem?

Is it appropriate to hard code replacing . with _ or should that be the authors responsibility?

prushforth:
There should be no hard coding.  If I did that it was a hack.

ahmadayubi:
> btw I believe the `<svg>` in this [comment](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/353#issuecomment-799869423) is actually the debug layer (you may have said that originally). It might be an idea to tag it with a class or otherwise identify it (make a pane for it??) ?

I'll tag it with a class.

Malvoz:
Re https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/353#issuecomment-802845461

> style

The CSS `svg path.fclass._20` matches the following HTML:
`<svg><path class="fclass _20">` (so cannot match the `feature` element)

feature

The exact match in a CSS selector for this HTML would be:
`feature#fclass\.20.fclass` (have to escape `.20` in the id or else it'd match `id="fclass" class="20 fclass"` which is invalid because class names cannot start with numbers) (more info https://mathiasbynens.be/notes/css-escapes).

Not sure if this is a useful or mostly confusing comment. But if we can avoid using dots in ID and class names that'd cause less confusion.

prushforth:
I noticed that the paths within a g, for example, the polygons in Quebec (a multipolygon), are correctly rendered as a set of paths, however (some of) these paths are given the aria-label from the parent (i.e. the `<featurecaption>` value).  That seems incorrect.

Also, I notice that there are paths that get the class="mapml-feature-outline" and these are shaded and stroked.  I don't think these should be stroked, only filled / shaded, and then the outline drawn (stroked, but not filled) on top of the filled-but-not-stroked polygon that gave rise to it, further down in the DOM / group.  The reason is tiles: when a polygon is truncated by a tile, you don't want to draw the border of the thing all the way around, you want to stroke only part of it, and hide the tile boundary.  I noticed the tile boundaries are still drawn on the test mapml tiles; I imagine this is why.

ahmadayubi:
> however (some of) these paths are given the aria-label from the parent (i.e. the <featurecaption> value). That seems incorrect.

I thought it might be useful as those parts are focusable when there is a span, should it still be removed?

ahmadayubi:
@prushforth 
> I notice that there are paths that get the class="mapml-feature-outline" and these are shaded and stroked. I don't think these should be stroked, only filled / shaded, and then the outline drawn (stroked, but not filled) 

The paths with class `mapml-feature-outline` are the outlines drawn, they're given a fill by default but it also has a style="fill:none;" which overrides it, but

prushforth:
> it would be better to not have a fill in the first place, doing so just saved lines of code slightly.

Because the code for rendering features is now shared by (MapMLFeaturesLayer, MapMLTileLayer, and hopefully eventually everything that needs to render MapML features e.g. query handler), it should be possible to trigger behaviour/processing differences with an option, or flag or something like that. Specifically, it's desirable (I *think*) to not duplicate paths when you don't have to i.e. you're not processing tiles and / or when a span does not exist in an un-tiled geometry.

Malvoz:
> > however (some of) these paths are given the aria-label from the parent (i.e. the value). That seems incorrect.
>
> I thought it might be useful as those parts are focusable when there is a span, should it still be removed?

Ideally they should be uniquely labelled by the author _if_ they're focusable, however it doesn't seem right that the paths are focusable (I couldn't find any such cases when inspecting any of the changed HTML files). Anyways here's NVDA screen/audio capture for keyboardInteraction.html between the master and customRenderer branches:

<table><td width="300"><video src="https://user-images.githubusercontent.com/26493779/112007484-f9be1680-8b24-11eb-8be8-11a1d2df82e4.mp4"></video></td></table>

<table><td width="300"><video src="https://user-images.githubusercontent.com/26493779/112008807-1c046400-8b26-11eb-8c72-cea59569bfb2.mp4"></video></td></table>

src/mapml/features/feature.js
```diff
@@ -0,0 +1,270 @@
...
+  /**
+   * Creates the <g> conditionally and also applies event handlers
+   * @private
+   */
+  _createGroup: function(){
+    if(this.options.multiGroup){
+      this.group = this.options.multiGroup;
+    } else {
+      this.group = L.SVG.create('g');
+      this.group.setAttribute('role', 'region');
```
Malvoz, line 64: Although [`role="region"`](https://www.w3.org/TR/wai-aria/#region) sounds like a good role because these are also geographic regions, that's not what the role is intended for. `region` is one of few [landmark roles](https://www.w3.org/TR/wai-aria/#landmark_roles) that may be exposed to users in an interface such as this:

document-regions

(image credit: https://www.youtube.com/watch?v=7Rs3YpsnfoI&t=157s)

Accessibility expert Scott O'Hara [states](https://www.scottohara.me/blog/2018/03/03/landmarks.html#region):
<q>The use of region landmarks should be limited, as too many landmarks can dilute their usefulness in quickly navigating to important areas of a page.</q>

And, in testing `role="region"` also does not seem to expose `aria-expanded="true|false` whereas `role="button"` does. IMO, `role="button"` in conjunction to `aria-roledescription="Feature"` makes the most sense.

Malvoz: OTOH maybe it's best to discuss this in a separate issue.

ahmadayubi: It was a recommendation given by @shepazu , it was suggested as a way to enable aria-label on `<g>` elements.

prushforth:
Nice work @ahmadayubi ! Let's merge this and process changes if and as required.

Issue #358, opened 2021-03-18
Small crisis: the summary/details marker has disappeared in the layer control in all browsers


prushforth:
![image](https://user-images.githubusercontent.com/2437285/111560526-0a5b3f00-8769-11eb-8698-e7b8cecf35ae.png)

The details can still be exposed by clicking on where the triangle should be.   Would be ideal to fix this so I can rush it into the GeoServer 2.19.0 release if I can get it in there in time.

Malvoz:
Changing the `display` of `<summary>` (https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/312) removes the native marker icons, I wonder why I didn't catch that in that PR. I can work on a quick fix if you want, but the alignment probably wont be as pretty.

prushforth:
Ok thanks. It's not really a crisis, I can always ask for a backport for the next release in June

Malvoz:
> I wonder why I didn't catch that in that PR, or why the screenshot does not reflect that

Chrome landed this "[feature](https://www.chromestatus.com/feature/6730096436051968)" (removing markers through changing `display` 🙄) in v89 which was released a week ago.

ahmadayubi:
Would adding our own custom ► button work?

prushforth:
> Would adding our own custom ► button work?

I don't think it's necessary because the triangle is pretty much a web standard for summary/details.  The issue doesn't arise because of something we did, but because of a change to the default [display](https://chromestatus.com/feature/6730096436051968) property in chrome. I'm not sure why I didn't notice that in Firefox earlier, since it appears that they had done the change a bit sooner than chrome.

PR #359, opened 2021-03-18
Remove flexbox in layer items

Malvoz:
Close https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/358.

## Preview

summary-markers

### DIFF
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -442,17 +442,22 @@ summary {
   position: relative;
 }
 
+/* Quick fix for https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/358,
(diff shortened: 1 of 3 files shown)


Malvoz:
This doesn't always look great though:

x

prushforth:
Thanks for looking at this.  I'm not sure how removing flexbox results in the markers appearing??  And still the [deprecation warning](https://chromestatus.com/feature/6730096436051968) in the console appears: 
image

Maybe the warning is due to Leaflet using that proprietary selector?  Anyway, the result looks good. It's ok by me to squash it.

prushforth (commented):
👍

PR #369, opened 2021-03-24
Update renderer

### LINKED ISSUE #368: SVG Rendering behavior changes needed
Comment changes that are needed for the renderer below.

ahmadayubi:
Test case from issue #368 now looks like so, it's also rendered using a single path element as they don't have any spans:
feat

Closes #368

### DIFF
--- a/src/mapml/features/feature.js
+++ b/src/mapml/features/feature.js
@@ -40,8 +40,11 @@ export var Feature = L.Path.extend({
     this._markup = markup;
     this.options.zoom = markup.getAttribute('zoom') || this.options.nativeZoom;
 
-    this._generateOutlinePoints();
     this._convertMarkup();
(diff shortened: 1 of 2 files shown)


prushforth:
Looks good. I had a look at  Leaflet and it seems that a [long time ago](https://github.com/Leaflet/Leaflet/commit/38476615e61ab709b3a54e3b0665f4a93bb57303), evenodd was  made the default value for paths. I believe it is because Leaflet's geometries may not necessarily be self-closed, or non-intersecting.  A Simple Features conforming geometry should work with the SVG [default of nonzero](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fill-rule).  I don't think it's actually a problem, so let's go with it until we have more information.

PR #370, opened 2021-03-25
Connect first and last points of a outline using modulo

### LINKED ISSUE #368: SVG Rendering behavior changes needed

ahmadayubi:
Test cases renders as:
newfea

Cause of issue:
Needed to preform modulo when generating outline, without it the first and last coordinate pairs in a coordinates element didn't behave appropriately. Coordinates for a polygon are better thought of as a circular linked list rather than a sequence with a start and an end. 

On a side node, the behavior of `%` operator in JavaScript is not consistent at all, occasionally -1 % 4 = 3 and other times -1 % 4 = -1, there's a work around in this PR for it but it's odd nonetheless. 

Closes #368

### DIFF
--- a/src/mapml/features/feature.js
+++ b/src/mapml/features/feature.js
@@ -185,14 +185,15 @@ export var Feature = L.Path.extend({
       for (let n of nodes) {
         let line = [];
         if (!n.tagName) {  //no tagName means it's text content
-          let c = '';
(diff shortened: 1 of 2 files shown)


prushforth:
This looks good, sorry for the painful experience!  Good work!

PR #371, opened 2021-03-26
Adds ability to wrap geometry subtypes with span and a elements + a element functionality within coordinates

### LINKED ISSUE #367: Links (and spans) in <geometry> content
An odd way to write a spec, but here are a few ideas, for discussion.  Please comment @shepazu and @Malvoz !

```html
<mapml lang="en" xmlns="http://www.w3.org/1999/xhtml/">
    <head>
        <meta charset="UTF-8" />
        <title>BasicPolygons</title>
        <meta content="text/mapml" http-equiv="Content-Type" />
        <meta name="extent" content="top-left-longitude=-2.000000,top-left-latitude=6.000000,bottom-right-longitude=2.000000,bottom-right-latitude=-1.000000" />
        <meta name="cs" content="gcrs" />
        <meta name="projection" content="MapML:EPSG:4326" />
    </head>
    <body>
        <!-- links here would make no sense -->
        <feature id="BasicPolygons.1" class="BasicPolygons">
        <!-- links here would make no sense -->
        <geometry>
            
        <!-- links here would be drawn around the entire geometry. Each polygon in a multipolygon would
                be interactive; same for any other geometry type. The holes in polygons would be non-interactive; 
                if there was another feature in
               the hole (an island, say), the island could itself be wrapped in a link, so we would want the
               island's link's behaviour to take over. A span here might not be too useful, but would not be
            harmful, I think; you already have the root geometry element to hang attributes on (e.g. class, and so on).
        FWIW I believe that some markup is not supposed to be found in the content of <a> (see Permitted Content): 
            https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#properties -->

            <!-- Regarding behaviour of links, I think we should use the existing "target" attribute vocabulary:
            https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-target
            and interpret that value as follows for maps:
        
            IFF the linked resource mime type is text/mapml:

                _self: replace the current layer (i.e. replace the current map layer; same behaviour as named style links) (Default)
                _blank: usually, add a new layer on top of map, but users can configure browsers to open a new window instead.
                _parent: replace all layers in the map with the linked layer. If no other layers, behaves as _self.
                _top: the topmost browsing context (the page) should navigate to the map resource, synthesize an HTML document
                around it, and load the map resource into the synthesized map and process any URL fragments to control the
                initial location of the map view. If no fragment, set zoom to lowest value, centered on extent of map doc.
                (the "highest" context that’s an ancestor of the current one). 
                If no ancestors, behaves as _self.  

            ELSE 
                do what browsers do with similarly tagged links of today's Web.



        -->
            
            <a href="https://example.org">
                <span class="harmlessButNotUseful">
                <multipolygon>
                    <!-- a link here would be drawn around the the entity or entities which it wrapped.  Say there were
                    three polygons in this multipolygon; a link could wrap two of the three. Similarly,
                    say you wanted to style two of the three in a different way than the third. You should
                    be able to wrap selected ones in a <span class="twoOfThreeStyle"> - that would 
                    prevent you from having to put duplicate spans to get the styles you want. -->

                    <polygon>
                        <!-- links in coordinates would be processed like spans are, except that
                        on output they are styled and behave as links.  Same deal, just copy them and
                    their attributes to the appropriate output. -->
                        <coordinates>-1.0 0.0 <a href="./door">0.0 1.0 1.0 0.0</a> 0.0 -1.0 -1.0 0.0</coordinates>
                    </polygon>
                </multipolygon>
              </span>
            </a>
            </geometry>
            <properties>
                <table>
                    <thead>
                        <tr>
                            <th role="columnheader" scope="col">Property name</th>
                            <th role="columnheader" scope="col">Property value</th>
                        </tr>
                    </thead>
                    <tbody>
                        <tr>
                            <th scope="row">ID</th>
                            <td itemprop="ID" />
                        </tr>
                    </tbody>
                </table>
            </properties>
        </feature>
    </body>
</mapml>
```

ahmadayubi:
Current Functionality:

- [x] ability to wrap multiple `<span>` (& `<div>`) around geometry subtypes (polygon, linestring, point, etc.), adding classes using appropriate precedence
- [x] ability to wrap `<a>`around geometry subtypes and have click functionality
- [x] enable `<a>` within `<coordinates>`
- [x] follow `<a>` target rules set in issue #367 

Closes #367


src/mapml/features/feature.js
```diff
@@ -53,36 +53,61 @@ export var Feature = L.Path.extend({
...
+      case "_parent":
+        for(let l of this._map.options.mapEl.querySelectorAll("layer-"))
+          if(l._layer !== this.options.featureLayer.options._leafletLayer) this._map.options.mapEl.removeChild(l);
+        this._map.options.mapEl.appendChild(layer);
+        this._map.options.mapEl.removeChild(this.options.featureLayer.options._leafletLayer._layerEl);
+        break;
+      case "_top":
+        window.location.href = link;
+        break;
+      case "_self":
```
prushforth, line 107: ^ Expected a 'break' statement before 'default (reported by jshint)

PR #372, opened 2021-03-27
Downgrade proj4 to 2.3.14. At ^2.4.0 leaflet starts to break.

prushforth:
There are failing tests because of this change, maybe we can review together next week.  The custom projection that we're working on for Ben starts to work though.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "leaflet": "^1.7.1",
     "path": "^0.12.7",
     "playwright": "^1.2.1",
-    "proj4": "^2.6.2",
+    "proj4": "^2.3.14",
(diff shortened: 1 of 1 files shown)


prushforth:
The old proj4 works well with this [custom projection](https://maps4html.org/experiments/custom-projections/esri-53032/)

ahmadayubi:
Which tests fail? They seem to pass on here

prushforth:
Yeah they only failed on my machine. It was the custom projection tests, unfortunately. Until we can figure out why the new version of proj4 doesn't work with theses polar projections in particular, I think we can just leave this PR open as a reminder. I prefer to not go backwards.

PR #373, opened 2021-03-29
Fix NaN errors when unprojecting coordinates that don't return location

prushforth:
on Earth.

Patch the version of proj4leaflet during the grunt build process. TBD how
this can be incorporated in proj4leaflet upstream, so that the patching 
isn't necessary.


prushforth:
@ahmadayubi what do you think of this patching business? I'll submit a PR to proj4leaflet later to see if we can get it fixed upstream, but in the meantime...

ahmadayubi (approved):
Looks good, I don't think they'll like it/merge it upstream since NaN may seem like the more appropriate returned value but it works for the custom projections in the experiments case.

PR #379, opened 2021-04-01
Replace code in custom projection definition with projection name

prushforth:


ahmadayubi (approved):
Seems to be working, kind of odd how the internal `code` value that proj4leaflet uses can just be arbitrary and still work.

prushforth:
It's a different philosophy of library.  They prioritize the parameters, not the code.  They spit on codes.

PR #380, opened 2021-04-04
Prefix the generic `web-map` class with `mapml-`

Malvoz:
I named this class quite generically at some point, it's safer (and more consistent) to prefix it with `mapml-` to reduce the odds of clashing with authors' own perhaps generically named divs.

### DIFF
--- a/debug/vector-tile-test.css
+++ b/debug/vector-tile-test.css
@@ -1,7 +1,7 @@
 
 path.fclass { stroke: blue; }
 
-svg.web-map.style-scope path.fclass._1 { stroke: blue; fill: #FF8C00;}
+svg.mapml-web-map.style-scope path.fclass._1 { stroke: blue; fill: #FF8C00;}
 path.fclass._2 { stroke: blue; fill: #F5DEB3;}
 path.fclass._3 { stroke: blue; fill: #7CFC00;}
 path.fclass._4 { stroke: blue; fill: #E9967A;}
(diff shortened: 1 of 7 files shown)


test/e2e/core/styleParsing.css
```diff
@@ -10,7 +10,7 @@ svg path.fclass {
   stroke: blue;
 }
 
-svg.web-map.style-scope path.fclass._1 {
+svg.mapml-web-map.style-scope path.fclass._1 {
```
Malvoz, line 13: I'm not sure how/where this SVG selector is used (it seems it's not to be in use atm). Should it also change to match the new `mapml-web-map` class, or do I just revert this change?

debug/vector-tile-test.css
```diff
@@ -1,7 +1,7 @@
 
 path.fclass { stroke: blue; }
 
-svg.web-map.style-scope path.fclass._1 { stroke: blue; fill: #FF8C00;}
+svg.mapml-web-map.style-scope path.fclass._1 { stroke: blue; fill: #FF8C00;}
```
Malvoz, line 4: I'm not sure how/where this SVG selector is used (it seems it's not to be in use atm). Should it also change to match the new `mapml-web-map` class, or do I just revert this change?

test/e2e/core/styleParsing.css
prushforth: I think that style-scope stuff is a holdover from the polymer days, but not sure.  I believe that was how they pretended that shadow dom was real (they called it "shady dom").  I might just delete the whole style if it's not in use.

Malvoz: Okay, I reverted the SVG selector change just in case. Feel free to merge.

PR #381, opened 2021-04-04
CSS selection/dragging of controls + outline improvements

Malvoz:
- [x] https://github.com/Maps4HTML/Web-Map-Custom-Element/commit/5c0db77406b126b03f16383c0d23e65331303012: Add  `mapml-focus-buttons` and `leaflet-popup-close-button` to the selectors that disable dragging/selection of controls.
- [x] https://github.com/Maps4HTML/Web-Map-Custom-Element/commit/9dae9096adaa499a2a71aa0d3a9407797413f6e1: Ensure outline is always 100% the size of the container, as it currently may not span the entire container after resize (which in best case is perhaps only a developer annoyance):<br><br>

### DIFF
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -442,7 +442,9 @@ summary {
   outline-offset: -1px;
   z-index: 1000;
   pointer-events: none;
-  position: relative;
+  position: absolute;
+  height: 100%;
+  width: 100%;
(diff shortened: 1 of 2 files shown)


prushforth:
LGTM

Issue #10, opened 2015-08-11
Allow points to be styled similar to LineString, Polygon etc.


prushforth:
Currently MapML/Leaflet converts points to L.Markers, which are stored the the leaflet-markers-pane.  Such points should not be rendered as markers, but as points transformed to SVG according to the CSS supplied, with possibly a default point symbol (that isn't a giant blue marker).

prushforth:
This is an issue with the MapML-Leaflet-Client library, not with the custom element.  Moving the issue over there.

PR #387, opened 2021-04-08
Add zooming when layer is added using <map-a>

### LINKED ISSUE #378: Zoom to fragment of <map-a href> else zoom to extent of linked resource
A page can have > 1 map on it, but it would be useful to be able to capture the URL fragment and set the map center or bounds based on the parameters found there.  

Maybe it could be in the form of a script that is distributed with the custom elements set, that you could configure easily to "do something" with a page's URL fragment if doing something is possible i.e. the fragment parses as something recognizable as a zoom,lon,lat or bbox etc.

ahmadayubi:
Zooms map to a link url fragment, if that doesn't exist then zoom to the layer extent. 
If `inplace` attribute is present then it doesn't zoom.

Closes #378

### DIFF
--- a/src/layer.js
+++ b/src/layer.js
@@ -302,4 +302,40 @@ export class MapLayer extends HTMLElement {
               {target: this}}));
         },this);
   }
+  focus(){
+    let map = this._layer._map,
+      tL = this.extent.topLeft.pcrs,
+      bR = this.extent.bottomRight.pcrs,
+      layerBounds = L.bounds(L.point(tL.horizontal, tL.vertical), L.point(bR.horizontal, bR.vertical)),
+      center = map.options.crs.unproject(layerBounds.getCenter(true)),
+      currentZoom = map.getZoom();
+
+    map.setView(center, currentZoom, {animate:false});
+    let mapBounds = M.pixelToPCRSBounds(
+      map.getPixelBounds(),
+      map.getZoom(),
+      map.options.projection);
+
+    //fits the bounds to the map view
+    if(mapBounds.contains(layerBounds)){
+      while(mapBounds.contains(layerBounds) && (currentZoom + 1) <= this.extent.zoom.maxZoom){
+        currentZoom++;
+        map.setView(center, currentZoom, {animate:false});
+        mapBounds = M.pixelToPCRSBounds(
+          map.getPixelBounds(),
+          map.getZoom(),
+          map.options.projection);
+      }
+      if(currentZoom - 1 >= 0) map.flyTo(center, (currentZoom - 1));
+    } else {
+      while(!(mapBounds.contains(layerBounds)) && (currentZoom - 1) >= this.extent.zoom.minZoom){
+        currentZoom--;
+        map.setView(center, currentZoom, {animate:false});
+        mapBounds = M.pixelToPCRSBounds(
+          map.getPixelBounds(),
+          map.getZoom(),
+          map.options.projection);
+      }
+    }
+  }
 }
--- a/src/layer.js
+++ b/src/layer.js
@@ -303,6 +303,7 @@ export class MapLayer extends HTMLElement {
         },this);
   }
   focus(){
+    if(!this.extent) return;
     let map = this._layer._map,
       tL = this.extent.topLeft.pcrs,
       bR = this.extent.bottomRight.pcrs,
--- a/src/layer.js
+++ b/src/layer.js
@@ -15,12 +15,19 @@ export class MapLayer extends HTMLElement {
     if (val) {
       this.setAttribute('src', val);
       if (this._layer) {
+        let focusOnLoad;
+        if(this._layer._events.extentload)
+          for(let e of this._layer._events.extentload)
+            if(e.name === "focusOnLoad")
+              focusOnLoad = e;
+
         // go through the same sequence as if the layer had been removed from
         // the DOM and re-attached with a new URL source.
         this.disconnectedCallback();
         var base = this.baseURI ? this.baseURI : document.baseURI;
         this._layer = M.mapMLLayer(val ? (new URL(val, base)).href: null, this);
         this._layer.on('extentload', this._onLayerExtentLoad, this);
+        if(focusOnLoad) this._layer._events.extentload.push(focusOnLoad);
         this._setUpEvents();
         if (this.parentNode) {
           this.connectedCallback();
--- a/src/layer.js
+++ b/src/layer.js
@@ -15,19 +15,12 @@ export class MapLayer extends HTMLElement {
     if (val) {
       this.setAttribute('src', val);
       if (this._layer) {
-        let focusOnLoad;
-        if(this._layer._events.extentload)
-          for(let e of this._layer._events.extentload)
-            if(e.name === "focusOnLoad")
-              focusOnLoad = e;
-
         // go through the same sequence as if the layer had been removed from
         // the DOM and re-attached with a new URL source.
         this.disconnectedCallback();
         var base = this.baseURI ? this.baseURI : document.baseURI;
         this._layer = M.mapMLLayer(val ? (new URL(val, base)).href: null, this);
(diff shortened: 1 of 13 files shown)


prushforth:
I don't see the zoomTo functionality that I was expecting with non-fragment text/mapml links (actually haven't tested fragments yet)?

For example, this link:`<map-a href="http://localhost:8081/geoserver/mapml/spearfish/cbmtile/">` , which matches the `_self` spec below, is not zooming to the spearfish layer extent when link is activated


_self: (Default if no target attribute) replace the current layer - **not zooming afaict**
_blank: add a new layer on top of map - **not zooming to layer extent afaict**
_parent: replace all layers in the map with the linked layer. If no other layers, behaves as _self. - **not zooming afaict**
_top: the topmost browsing context (the page) should navigate to the map resource, synthesize an HTML document
                around it, and load the map resource into the synthesized map and process any URL fragments to control the
                initial location of the map view. If no fragment, set zoom to lowest value, centered on extent of map doc.
                (the "highest" context that’s an ancestor of the current one). 
                If no ancestors, behaves as _self.  - **not zooming afaict**

ahmadayubi:
That's odd, I got the behavior on my end:
https://user-images.githubusercontent.com/55214462/114193446-87b74100-991c-11eb-84cf-425307b5f440.mp4

Could you try using the featureLink.html file and see if it works there? Run the test server `node test/server.js` and go to http://localhost:30001/featurelinks.html

ahmadayubi:
There was an error with inplace but it should be focusing spearfish when you add it, I'll try using geoserver and see if I can recreate the issue.

prushforth:
There's this error in the console actually, sorry I didn't look before: 
image

ahmadayubi:
Seems to be an issue with timing, the focus() function is called on a layer before the layer's extent is fully processed, although it's strange that the geogratis ones loaded in but geoserver ones didn't.

prushforth:
> Seems to be an issue with timing, the focus() function is called on a layer before the layer's extent is fully processed, although it's strange that the geogratis ones loaded in but geoserver ones didn't.

This was a problem in early iterations. It may have been affected by this PR, perhaps: https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/366

ahmadayubi:
Maybe there should be an event that fires when a layers extent is set, I don't believe there's an event fired from the map that lets you know the layer is fully added, do you know of any?

prushforth:
I used to have the extentloaded event, not sure if that still happens.

prushforth:
I also remember hacking with a labelchanged event.  tbh I have not fully grasped asyn programming techniques, which is probably a major issue in this space.

ahmadayubi:
Probably has to do with the `extentload` event being fired on the first request rather than on when it eventually negotiates the correct projection.

prushforth:
First check of a fragment url such as:

`<map-a href="http://localhost:8081/geoserver/mapml/topp:states/cbmtile/#7,-98,38.5" type="text/mapml" target="_self">`

It seems to zoom to 7, lon=98, lat=38.5, but it doesn't load the states layer.

prushforth:
The latest seems to work better for waiting on projection changes, but still getting the error below for url fragments:

 `<map-a href="http://localhost:8081/geoserver/mapml/topp:states/cbmtile/#6,-96.36158755253881,39.668588667742846">`

image

prushforth:
Seems like this link doesn't work like I was expecting, although it does something (zoom/pan to the specified location):

`<map-a href="#10,-76.474059,45.345151">`

would correspond to the _self, below, I think, with the URL defaulting to the "URL" of the current layer, whether or not it is an actual document is another matter, I guess, since the content could be inline.  We should be able to use fragments like the above with each of the target types below, I think.

_self: (Default if no target attribute) replace the current layer
_blank: add a new layer on top of map
_parent: replace all layers in the map with the linked layer. If no other layers, behaves as _self. 
_top: the topmost browsing context (the page) should navigate to the map resource, synthesize an HTML document
around it, and load the map resource into the synthesized map and process any URL fragments to control the
initial location of the map view. If no fragment, set zoom to lowest value, centered on extent of map doc.
(the "highest" context that’s an ancestor of the current one).
If no ancestors, behaves as _self.

ahmadayubi:
I think in a conversion we had I believe you mentioned that if the type=text/mapml and target=_self it should "simply reposition the viewer with zoomTo()". It was decided that the fragments would only work with type=text/mapml, in which case using the URL of the layer in the case of inline features doesn't work since the URL of the layer is pointing to an html page. In the case of templated features, what value do you get from replacing the layer with itself vs just repositioning?

prushforth:
> It was decided that the fragments would only work with type=text/mapml

Right, but the default is text/mapml, correct? 

>  the URL of the layer is pointing to an html page

That's true, but I don't think it affects the use of that URL as the context for zoom/pan to operations?  I may be missing something though.   I believe the intent was "simply reposition the viewer with zoomTo()". The current behaviour as I mentioned does move the map but not to the coordinates specified.  There's a console error as well:

image

>  In the case of templated features, what value do you get from replacing the layer with itself vs just repositioning?

I would not replace the layer with itself - for urls that begin with #, I would not try to resolve against a base url, I would just invoke zoomTo and hope for the correct result.  In a browser if you put a fragment that doesn't have a corresponding id, the browser doesn't complain, it does nothing, which seems to be the appropriate response in case of error here to.

prushforth:
`<map-a href="#z,lon,lat">` could be considered a declarative API that exposes `zoomTo(zoom, longitude, latitude)`.

ahmadayubi:
So in the case of `<map-a href="#3,12,43">`, the viewer should consider that localhost:3000/mypage.html#3, 12, 34 (for example), try to follow the link according to the target type rules (and assume its text/mapml), and even in the case of a failed loading of the layer the fragment should be honored? Is that correct?

ahmadayubi:
Another note is that the `zoomTo(lat, lon, zoom)`, accepts parameters in the order of `lat, lon, zoom`, and not `zoom, lon, lat`. There's probably a reason why, but that's how it's currently implemented.

prushforth:
> So in the case of <map-a href="#3,12,43">, the viewer should consider that localhost:3000/mypage.html#3, 12, 34 (for example), try to follow the link according to the target type rules (and assume its text/mapml), 

In the case of `href="#..."` I *think* it would be ok to just **assume** the current document, be it an HTML page or a MapML document is the search environment, and if the fragment can be understood (parsed), invoke zoomTo on it, end of story.

> and even in the case of a failed loading of the layer the fragment should be honored? Is that correct?

I would not try to load anything in this case.  Could it be treated as a shortcut?  The idea behind fragment identifiers is that they don't identify anything on a server at all, they are simply processed client side.  So when there is a URL with a fragment, the URL parsing and resolution and loading takes place  and ** then ** the fragment is processed.  So if there is no url, I think it might be safe to assume / perform a zoomTo without otherwise attempting to process the url.

> Another note is that the zoomTo(lat, lon, zoom), accepts parameters in the order of lat, lon, zoom, and not zoom, lon, lat. There's probably a reason why, but that's how it's currently implemented.

OK, that's important to remember and document especially when documenting `<map-a href="...">`

I think I might have had the idea that you might not know the zoom, but you wanted to position the map over a location, but it does look a little funny.

ahmadayubi:
During a discussion it was decided that `<map-a href="#...">` would behave as target="_self" for all target values (_blank, _top, etc.).

If the `href="#..."` then simply `zoomTo()` the location
If the `herf="../#..."` (relative URL with fragment) then resolve the URL, replace the current layer with the resolved URL and finally `zoomTo()` the location

prushforth:
LGTM - great work @ahmadayubi, let's go!

prushforth:
I can't get this URL to load properly, it replaces the whole page:

`<map-a href="https://geogratis.gc.ca/mapml/en/cbmtile/bedrockgeology/#4,-97.4376874,54.2737257" target="_self" type="text/mapml">`

PR #413, opened 2021-04-19
Adds opacity setter & getter & preserves opacity on _self links

### LINKED ISSUE #393: Enable opacity retention / setting on layer change
When a layer in the layer control is replaced, for example by the user selecting a different style, or by following a _self link, it should retain the opacity value through that state change.  

Also, could we make it settable via a property or CSS api ?  What is the "right" design for this?  

/cc @Malvoz

ahmadayubi:
Works with _self links, retains it's opacity, although I'm not 100% sure it's appropriate to retain since you're replacing the layer not modifying it.
Closes #393

### DIFF
--- a/src/layer.js
+++ b/src/layer.js
@@ -59,6 +59,16 @@ export class MapLayer extends HTMLElement {
       this.removeAttribute('hidden');
     }
   }
+
+  get opacity(){
+    return this._layer._container.style.opacity || this._layer.options.opacity;
+  }
+
+  set opacity(val) {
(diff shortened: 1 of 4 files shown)


prushforth:
@ahmadayubi sorry this has got conflicts on it, probably because I left it sit. Can you resolve the conflicts, I will merge.

prushforth:
Does this maintain opacity when following a style link?  That might seem less weird to you, since the user is expecting a simple style change, and a change in opacity is annoying through that.

PR #415, opened 2021-04-20
Sandbox iframes

### LINKED ISSUE #284: iframes for query responses can execute scripting

ahmadayubi:
Closes #284

### DIFF
--- a/src/mapml/handlers/QueryHandler.js
+++ b/src/mapml/handlers/QueryHandler.js
@@ -166,8 +166,8 @@ export var QueryHandler = L.Handler.extend({
         let div = L.DomUtil.create("div", "mapml-popup-content"),
             c = L.DomUtil.create("iframe");
         c.style = "border: none";
-        c.srcdoc = `<meta http-equiv="content-security-policy" content="script-src 'none';">` + mapmldoc.querySelector('feature properties').innerHTML;
-        c.setAttribute("sandbox","allow-same-origin");
+        c.srcdoc = mapmldoc.querySelector('feature properties').innerHTML;
+        c.setAttribute("sandbox","allow-same-origin allow-forms");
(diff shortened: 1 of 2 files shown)


Malvoz (commented):
With `sandbox` in place we can remove all `<meta http-equiv="content-security-policy" content="script-src 'none';>` as there's no added benefit in keeping it, neither from a security standpoint nor browser support.

Malvoz (approved):
We may want/need to add more values to the allowlist, for example: `allow-forms` would allow the author to add a `<form>` in a popup that the user could submit for example, otherwise the form submission would be blocked.

And _not_ adding `allow-top-navigation-by-user-activation` may interfere with `target` of `<map-a>`, I suppose.

We can add more values as needed if/when we notice stuff not working as intended. Other than that, LGTM.

PR #422, opened 2021-04-22
Add width, height, alt and presentation to img

### LINKED ISSUE #421: Tile images are missing some attributes
In `mapml-templated-tile-container` images are missing `width` and `height` attributes (the attributes are set on a parent `<div>` which has no effect):

tile-img-no-width-height-attr

In `mapml-static-tile-layer` images are missing `width` and `height` attributes:

tile-img-no-role-presentation
This image should also be hidden from screen readers (same as the above image), `aria-hidden="true"`  should be sufficient but for consistency with the other images it can have `role="presentation"` and empty `alt` instead (the legacy way of hiding images before `aria-hidden="true"` was broadly supported I believe). Otherwise screen readers may announce the URL of the image when the user clicks on it.

ahmadayubi:
Closes #421

### DIFF
--- a/src/mapml/layers/TemplatedTileLayer.js
+++ b/src/mapml/layers/TemplatedTileLayer.js
@@ -65,9 +65,6 @@ export var TemplatedTileLayer = L.TileLayer.extend({
           tileSize = this.getTileSize();
       L.DomUtil.addClass(tileGroup, "mapml-tile-group");
       L.DomUtil.addClass(tileGroup, "leaflet-tile");
-      
-      tileGroup.setAttribute("width", `${tileSize.x}`);
-      tileGroup.setAttribute("height", `${tileSize.y}`);
 
       this._template.linkEl.dispatchEvent(new CustomEvent('tileloadstart', {
         detail:{
--- a/src/mapml/layers/TemplatedTileLayer.js
+++ b/src/mapml/layers/TemplatedTileLayer.js
@@ -62,12 +62,12 @@ export var TemplatedTileLayer = L.TileLayer.extend({
     },
     createTile: function (coords) {
       let tileGroup = document.createElement("DIV"),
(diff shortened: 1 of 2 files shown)


src/mapml/layers/TemplatedTileLayer.js
```diff
@@ -62,12 +62,12 @@ export var TemplatedTileLayer = L.TileLayer.extend({
...
       let tileGroup = document.createElement("DIV"),
-          tileSize = this._map.options.crs.options.crs.tile.bounds.max.x;
+          tileSize = this.getTileSize();
       L.DomUtil.addClass(tileGroup, "mapml-tile-group");
       L.DomUtil.addClass(tileGroup, "leaflet-tile");
       
-      tileGroup.setAttribute("width", `${tileSize}`);
-      tileGroup.setAttribute("height", `${tileSize}`);
+      tileGroup.setAttribute("width", `${tileSize.x}`);
+      tileGroup.setAttribute("height", `${tileSize.y}`);
```
Malvoz, line 70: Browsers don't do anything with HTML `width` and `height` attributes on `<div>` elements so I think these can be removed.

Malvoz (approved):

src/mapml/layers/TemplatedTileLayer.js
ahmadayubi: I believe they were added just to let whoever know the tileSize if they ever looked into the shadow dom, since tiles can contain things other than `` with the tileloadstart event.

Malvoz: Okay well, the size is already visible in the style attribute on that same `<div>`, e.g. `style="width: 256px; height: 256px;"`. I don't think it's correct to set invalid attributes for developer convenience.

PR #424, opened 2021-04-23
Link preview

### LINKED ISSUE #392: Add map control to simulate hyperlink URL destination output in lower left of map
Browsers implement a similar control when you hover over a link, they display the destination of the link in a control in the lower left of the browser viewport.  Programs don't appear to have access to that, so we can't write to it, but we could simulate that control with a text display of the URL of a map link in the lower left of the map viewport.

ahmadayubi:
https://user-images.githubusercontent.com/55214462/115907064-59665500-a436-11eb-9810-352813c19efa.mp4

Closes #392

### DIFF
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -507,3 +507,17 @@ svg.leaflet-image-layer.leaflet-interactive g {
   color: #fff;
   text-shadow: 1px 1px 1px #000, 1px 1px 1px #000;
 }
+
+.mapml-link-preview {
+  position: absolute;
+  left: 0;
+  bottom: 0;
+  background-color: rgb(222,225,230);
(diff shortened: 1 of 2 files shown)


Malvoz:
I just want to note that with standard links the link previews are not only displayed on hover but focus too.

prushforth:
I noticed that when you follow a link which replaces the current map via target="_top" for example to a map on another origin, the link preview doesn't display on links in that new map.

https://user-images.githubusercontent.com/2437285/115925935-f5e92100-a44f-11eb-84c9-05eb0dbd29d0.mp4

prushforth:
ok that comment was incorrect, deleted.

PR #426, opened 2021-04-26
Point fix

### LINKED ISSUE #409: Point generated when no / null geometry in feature returned by query is not near click / touch location
The handler should generate a point at the click location. It generates a point, but it seems to be far from the click location.

ahmadayubi:
I believe this fixes the issue, I tried it with the demo provided in issue #409 and it seemed to fix it.

- [x] Has test

Closes #409


prushforth:
Nicely done!

PR #433, opened 2021-04-27
Small refactor + addition of unit tests for util functions

ahmadayubi:


prushforth:
@ahmadayubi small merge conflict; going to need you to resolve as I can't tell which is the correct one to choose

PR #440, opened 2021-04-28
Allow linked feature's stroke to be customized, change linked feature style once visited.

### LINKED ISSUE #431: Allow stroke styling of linked features + adding different styling for visited features
Currently the blue stroke for linked features is hard coded, it would be better to allow full customization of linked feature styles.

Also when a feature is clicked, the style should be adjusted to reflect the fact it was clicked.

ahmadayubi:
https://user-images.githubusercontent.com/55214462/116460742-610e6b00-a835-11eb-8f5e-fd3d9b9e4341.mp4

When a feature is visited it is tagged with a `map-a-visited` class. The first example is the default styling, the second is the an example of custom styling of links before and after being clicked.
Closes #431

### DIFF
--- a/src/mapml/features/feature.js
+++ b/src/mapml/features/feature.js
@@ -60,6 +60,7 @@ export var Feature = L.Path.extend({
     container.classList.add('mapml-link-preview');
     container.appendChild(p);
     elem.classList.add('map-a');
+    if (link.visited) elem.classList.add("map-a-visited");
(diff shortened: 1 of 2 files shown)


prushforth:
I'm not sure why but the default blue links have disappeared from this [example](https://maps4html.org/experiments/custom-projections/Arctic-SDI/).  Some people may not like the blue hyperlink, but I think I should get a hyperlink blue tatoo.

PR #441, opened 2021-04-29
Add copying of mapml content

ahmadayubi:
Copying extent is done in mapml instead of CSV format, copying mapml of map is also a new menu item in this PR. 

items


prushforth:
I noticed that the layer extent copy is slightly off:

`<meta name="extent" content="top-left-easting,-3263369, top-left-northing,4046262, bottom-right-easting,3823752, bottom-right-northing,-1554448">`

should be:

`<meta name="extent" content="top-left-easting=-3263369, top-left-northing=4046262, bottom-right-easting=3823752, bottom-right-northing=-1554448">`

(equals signs missing).

Issue #274, opened 2021-02-01
`role="application"`


Malvoz:
I wonder if [`role="application"`](https://www.w3.org/TR/wai-aria-1.1/#application) should replace `role="region"` on the leaflet-container. See https://github.com/w3c/csswg-drafts/issues/5275#issuecomment-770008199 for background information.

Screen readers have different modes of operation, I believe the default mode is usually "browse mode" (this is true for NVDA), in browse mode you can use the arrow keys to read text in a certain direction, for example. Because browse mode intercepts keyboard commands, it is currently not possible to:

- Pan the map display
- Open the context menu

There's also an issue with [misplaced popups](https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/423).

With the implications on [keyboard interactions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Application_Role#keyboard_interactions) from `role="application"` we wouldn't have these issues, because:

> Keyboard interaction is completely under the web author's control and can be anything associated with the particular widget being implemented.

But that can cause confusion for the user, as they may not be able to operate their screen reader as they typically do with keyboard commands (see [NVDA default keyboard shortcuts](https://dequeuniversity.com/screenreaders/nvda-keyboard-shortcuts#nvda-nvda_shortcut_keys) for reference). Of course, a user _can_ switch between modes, but how would they know to do so, to enable panning etc.? Does the typical screen reader user even know how to manually switch modes? I obviously don't have a good enough understanding of how every day screen reader users use their tool. @nchan0154 do you happen to have any insight to this?

Esri's [a11y-map](https://github.com/Esri/a11y-map) uses `role="application"` (and their interaction model for keyboard users is superb by the way, panning is the primary interaction method as opposed to sequentially navigating through focusable features).

nchan0154:
I believe generally those who use screen readers due to disability consider themselves highly proficient ([68% would classify themselves as advanced](https://webaim.org/projects/screenreadersurvey8/#proficiency), though obviously those who would respond to a survey may be a self selecting group) and would be capable of manually switching modes!

[...]

PR #466, opened 2021-05-18
Add role='application' to FeatureGroup <g> element

### LINKED ISSUE #423: Popups are misplaced when using a screen reader
I'm getting a strange bug where the popup is misplaced when opening it by pressing <kbd>Enter</kbd>,  _only_ when using a screen reader (NVDA in this case):

<table><td width="400"><video src="https://user-images.githubusercontent.com/26493779/106805628-3c0bc100-6667-11eb-947e-585c9f68525a.mp4"></video></td></table>

_Originally posted by @Malvoz in https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/270#issuecomment-772808780_

prushforth:
Fixes #423

### DIFF
--- a/src/mapml/features/featureGroup.js
+++ b/src/mapml/features/featureGroup.js
@@ -27,6 +27,8 @@ export var FeatureGroup = L.FeatureGroup.extend({
     }
 
     this.options.group.setAttribute('aria-label', this.options.accessibleTitle);
+    // fix issue https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/423
+    this.options.group.setAttribute('role', 'application');
     if(this.options.featureID) this.options.group.setAttribute("data-fid", this.options.featureID);
   },
 


Malvoz:
It seems that NVDA announces everything on the page after pressing a feature to open a popup. At this point, I'm starting to think we're doing the user a disservice by not just making the entire map `role="application"`.

prushforth:
Ok I'll have a look at that when I get a chance.

Malvoz:
Maybe we should just go ahead and replace `role="region"` of the maps (in web-map.js as well) with `role="application"` (and close out https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/274).

I'll go ahead and do some testing with that approach, see how it behaves.

prushforth:
For some reason I didn't experience NVDA  reading the whole page. I only tried it on the restaurant page though.

PR #467, opened 2021-05-18
Set `role="application"`

Malvoz:
Set `role="application"` on the map element.

Close https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/274, Close https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/466.

Ideally `role="region"` on the Leaflet container would simply have been replaced by `role="application"`, however doing so results in <q>application</q> not being announced by NVDA.

I've tested numerous combinations, e.g. moving both `aria-label` and `role` to the map element, or only setting `role` on the map element and `aria-label` and/or `aria-roledescription` on the Leaflet container, and vice-versa, etc. This approach seems to be the most consistent one.

### DIFF
--- a/src/mapml-viewer.js
+++ b/src/mapml-viewer.js
@@ -219,7 +219,9 @@ export class MapViewer extends HTMLElement {
     
           this.setControls(false,false,true);
           this._crosshair = M.crosshair().addTo(this._map);
-    
+          
+          // https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/274
+          this.setAttribute('role', 'application');
           // Make the Leaflet container element programmatically identifiable
           // (https://github.com/Leaflet/Leaflet/issues/7193).
           this._container.setAttribute('role', 'region');
--- a/src/web-map.js
+++ b/src/web-map.js
@@ -254,6 +254,8 @@ export class WebMap extends HTMLMapElement {
             this.poster.style.display = 'none';
           }
           
+          // https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/274
+          this.setAttribute('role', 'application');
           // Make the Leaflet container element programmatically identifiable
           // (https://github.com/Leaflet/Leaflet/issues/7193).
           this._container.setAttribute('role', 'region');


Malvoz:
@prushforth as you noticed popups are no longer announced, which is caused by this PR.

If we revert this and apply https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/466 instead it seems everything works as intended. I think it's strange to give featuregroups `role="application"` but the results are better than this.

Malvoz:
Either way, we're just working around the root issue.

Malvoz:
To be safe it's probably best to go back to square 1 by reverting this, and re-opening these:

> Close #274, Close #423, Close #466.

`role="application"` on the map element(s) may have other implications we don't know about yet.

If the web had 1 accessible map library we could build on, that'd be cool.

prushforth:
Huh I thought we were making progress, and we might be able to tweak something to get the popups read. All your changes seem to have worked, and as far as I recall, popups weren't being read even with limiting the role="application" to the featuregroup.

Malvoz:
Yes we make progress in certain areas and introduce issues in others...

Setting `role="application"` has the following benefits:
- _Only_ the `role`s and `aria-label` of the map element and leaflet container is announced when the map is tabbed to (as opposed to everything in the map container being announced, including all controls, all attribution, etc.). I think this is an improvement.
- The first marker _is_ announced when tabbed to (remember <q>Hung Sum Restaurant</q> in https://maps4html.org/experiments/screenreader/restaurants/ was never announced on the first tab)
- Popups are not misplaced

But as noted popups are now not announced, I can confirm they were announced before applying `role="application"`.

> we might be able to tweak something to get the popups read

Yes... https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/274#issuecomment-771901816:

> If we were to do this, we could perhaps set `role="document"` for things like author provided HTML (e.g. popup content) so that they don't inherit the application role's behavior

I've tested setting `content.setAttribute("role", "document")` in MapLayer.js locally, it works, the content of the popup is announced and the screen reader user can use keyboard shortcuts to navigate that content.

Issue #354, opened 2021-03-16
Popups should be programmatically determinable


Malvoz:
The `leaflet-popup` is a plain `<div>` (click any marker to view a popup here: https://maps4html.org/experiments/index.html). I believe it is a failure of SC [1.3.1 Info and Relationships](https://www.w3.org/TR/WCAG21/#info-and-relationships) to not convey its structure (as visually presented) to <span title="Assistive Technology">ATs</span>. This issue is on the list of [upstream accessibility issues (for popups)](https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/247#popups).

I'd like to think that [`role="dialog"`](https://www.w3.org/TR/wai-aria-1.2/#dialog) is appropriate for the element, but I'm not sure (we can hopefully get some feedback on this per https://github.com/w3c/aria/issues/1421#issuecomment-791645912). WDYT? Is `dialog` appropriate for popups?

/cc @shepazu and @nchan0154, in case you have any opinions on this. 🤗

Malvoz:
Accessibility expert Scott O'Hara stated in https://github.com/w3c/aria/issues/1421#issuecomment-824152117:

> per the provided example, i could see that as a non-modal dialog.

(the example provided was https://maps4html.org/experiments/index.html)

With that response I do think we should use `role="dialog"` and [`aria-modal="false"`](https://www.w3.org/TR/wai-aria-1.2/#aria-modal) for popups. I'm assuming we'll continue to use popups to some extent despite https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/403.

nchan0154:
Apologies for missing this earlier. I agree that `role="dialog"` and `aria-modal="false"` are most appropriate here (feeling validated by this as I had decided this in my head before I saw Scott's comment 😂) In the specific case here I do think the appropriate accessible name is the `name` as the content of the tooltip, and it would also be great if Leaflet let you pass this in as an option.

I'm also noticing this very interesting bug on NVDA + Chrome/Firefox where opening a popup with the keyboard causes the popup to open very left and very up of the actual point, so much so that the map scrolls. I'm not sure how much of this is MapML and how much of it is Leaflet since last I checked, Leafalet didn't even have popups that were keyboard accessible, but possibly worth noting.

Malvoz:
Thanks Nic.

> I do think the appropriate accessible name is the name as the content of the tooltip

[...]

PR #468, opened 2021-05-18
Improve controls in popups

Malvoz:
Fixes the following item in https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/247#popups:

> - [ ] Popup's Close button (and controls in `mapml-focus-buttons`) should be at least 44 by 44 CSS pixels (relates to #215).

Also replaces the `<div>` container for `mapml-focus-buttons` with `<nav>`.

### DIFF
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -442,10 +442,8 @@ summary {
 	padding: 0 4px 0 4px;
 	border: none;
 	text-align: center;
-	width: 18px;
-	height: 14px;
 	font: 16px/14px Tahoma, Verdana, sans-serif;
-	color: #c3c3c3;
+	color: inherit;
 	text-decoration: none;
 	font-weight: bold;
   background: transparent;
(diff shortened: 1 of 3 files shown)


Malvoz:
Also comitted https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/468/commits/96127831d3f9ecb1b64405c3033efe5d52f4e4af to set `role="dialog"` on popups (does not entirely fix but is one step of https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/354) to improve the UX for screen reader users.

prushforth:
I notice that the content of the popup is not read by NVDA. But it's starting to work better  I think.  Thanks for this!

Malvoz:
> I notice that the content of the popup is not read by NVDA.

That is caused by https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/467...

PR #470, opened 2021-05-19
Set inner `role="document"` in popups

Malvoz:
Fixes issue with popups not being announced after https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/467.

Edit: To clarify, [`role="document"`](https://www.w3.org/TR/wai-aria-1.2/#document) is used to escape the `role="application"`s behavior.

### DIFF
--- a/src/mapml/layers/MapLayer.js
+++ b/src/mapml/layers/MapLayer.js
@@ -1184,6 +1184,7 @@ export var MapMLLayer = L.Layer.extend({
 
       popup._container.setAttribute("role", "dialog");
       content.setAttribute("tabindex", "-1");
+      // https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/467#issuecomment-844307818
       content.setAttribute("role", "document");
       popup._count = 0; // used for feature pagination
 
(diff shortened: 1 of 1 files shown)


prushforth:
This works well, as far as I can tell. I noticed in ff the first restaurant was announced as a diagram.  Anyway, it looks reasonable from what I can see.  I wonder if there's an automated testing framework for NVDA/JAWS etc.

Malvoz:
> I noticed in ff the first restaurant was announced as a diagram.

That is an issue separate from any of the latest changes. Chrome announce markers as <q>graphic</q>, this is because markers are SVG `<g>` elements with no explicit `role` set.

We've discussed markers' `role`s before and I thought they should be `role="button"`, perhaps with `aria-roledescription="feature"` (in which case `aria-expanded="false"` is correctly exposed, and announced as <q>collapsed</q>).

Malvoz:
> I wonder if there's an automated testing framework for NVDA/JAWS etc.

[axe-core](https://github.com/dequelabs/axe-core) (which has contributors from ARIA spec editors) powers the accessibility tests in Lighthouse and Webhint. But there's a limit to what automated tests can find, so manual testing for WCAG compliance is a must.

PR #471, opened 2021-05-20
Conditional role link

prushforth:
Make role="application" or role="link" attribute of featureGroup conditional on whether the graphic is a link or simply a popup.

@Malvoz I noticed that links were not being announced.  I think this resolves it, or improves the experience a little.


Malvoz:
Where can I test links? I know we have featureLinks.html but that uses `<map>` so that's not going to be accessible no matter our efforts.

prushforth:
A couple here, I'll add another example

http://maps4html.org/experiments/painting/

src/mapml/features/featureGroup.js
```diff
@@ -19,8 +19,11 @@ export var FeatureGroup = L.FeatureGroup.extend({
       if(layers.length === 1 && firstLayer.options.link) this.options.link = firstLayer.options.link;
       if(this.options.link){
         M.Feature.prototype.attachLinkHandler.call(this, this.options.group, this.options.link, this.options._leafletLayer);
+        this.options.group.setAttribute('role', 'link');
       } else {
         this.options.group.setAttribute("aria-expanded", "false");
+        // fix issue https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/423
+        this.options.group.setAttribute('role', 'application');
```
Malvoz, line 26: I think still that `role="button"` is better suited in this case. I'm not going to object to using `application`, but just saying... for the record.

Malvoz:
> I noticed that links were not being announced.

In my testing NVDA announces: <q><i>Graphic</i>, Girl zoom equals 2, <i>Link</i></q>.

If `role="none"` (and space separated/or `presentation`, these are synonyms, latter is legacy role) is set on the parent `<svg>`, it is not announced as <q>Graphic</q> and the output is: <q>Girl zoom equals 2, <i>Link</i></q> which I think is more appropriate.

Malvoz (approved):

src/mapml/features/featureGroup.js
Malvoz: Not sure if `role="button"` fixes https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/423 though. So maybe that's why. OTOH since we're currently using `role="application"` on the map element that fixes that issue.

prushforth: Updated

src/mapml/features/featureGroup.js
```diff
@@ -19,8 +19,11 @@ export var FeatureGroup = L.FeatureGroup.extend({
       if(layers.length === 1 && firstLayer.options.link) this.options.link = firstLayer.options.link;
       if(this.options.link){
         M.Feature.prototype.attachLinkHandler.call(this, this.options.group, this.options.link, this.options._leafletLayer);
+        this.options.group.setAttribute('role', 'link');
       } else {
         this.options.group.setAttribute("aria-expanded", "false");
+        // fix issue https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/423
```
Malvoz, line 25: I think we should remove this comment, I'm not sure it holds true since this is no longer `role="application"` (instead relies on the entire map to be `application` (I also noted this in https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/471#discussion_r636193258).

Malvoz (approved):
Just noted that there's a comment that probably should be removed. Other than that LGTM!

src/mapml/features/featureGroup.js
prushforth: How about linking to https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/471#discussion_r636193258, or maybe I'm misunderstanding?

prushforth: I mean https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/423#issuecomment-824951896, sorry

Malvoz: That comment talks about Leaflet's `` markers not being operable without `role="button"`, but for some reason that's not an issue with `<g tabindex="0">` . The main reason we're using `role="button"` is to convey that this component operates as such. And potentially overriding any announcements such as <q>Graphic</q> for `<g>` (we know that's true for `<svg>`, don't remember the case for `<g>`). I don't think we need to link to anything there actually.

Issue #297, opened 2021-02-15
Create User Preferences API and UI


shepazu:
- Allow users to select UI preferences
  - e.g. color of highlight selection
- Honor browser-set preferences
  - `prefers-contrast` 
- Enable user stylesheets

See https://github.com/Maps4HTML/HTML-Map-Element-UseCases-Requirements/issues/43

prushforth:
Should this also actually be a Use Case / Requirement, apart and aside from the linked issue?

prushforth:
See also: #494

PR #478, opened 2021-06-25
native-like styles for interactive features

Malvoz:
- [x] Add `stroke` to interactive features on `:hover` (in conjunction to `:focus`). This enables users to see the entire hit area of the interactive feature that is being hovered.
- [x] Remove the fixed `stroke-width: 5` (currently applied on `:focus`) such that the highlighting inherits (and does not override) the initial stroke-width of a feature.
- [x] Remove `outline` on `:focus` as imposed by our [outline reset](https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/e1264b86b897bbdeb7a3dd2ff7da31c3d53aa313/src/mapml.css#L369-L376) because `stroke` takes care of the highlighting of interactive features, no need to also have an additional outline.
- [x] Remove `stroke` animations. In the spirit of the W3C TAG guidelines for custom components we should keep styles to a bare minimum. If authors want animations they should implement it themselves (even though I'm aware author-provided CSS is not yet supported).

## Comparison
Interaction steps as displayed in the videos below:
1. _Hover_ Norway, Sweden, Finland, Russia with the mouse
2. _Focus_ Russia with the keyboard

<table border="0">
<tr><td width="800"><h3>Before</h3><video src="https://user-images.githubusercontent.com/26493779/123456764-81fbfe00-d5e3-11eb-934d-5ed5beef71bf.mp4"></video></td></tr>
<tr><td width="800"><h3>After</h3><video src="https://user-images.githubusercontent.com/26493779/123456746-7c061d00-d5e3-11eb-9ae2-c2abe4414021.mp4"></video></td></tr>
</table>

Thoughts? Personally, I'm not sure about:
> - [x] Add `stroke` to interactive features on `:hover` (in conjunction to `:focus`). This enables users to see the entire hit area of the interactive feature that is being hovered.

I think it's a nice addition and makes it clear where the clickable area of an interactive feature is, but it may also feel "chatty". Either way, the highlighting of features may be subject to user-preferences: https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/297.

### DIFF
--- a/src/mapml/layers/FeatureLayer.js
+++ b/src/mapml/layers/FeatureLayer.js
@@ -13,16 +13,21 @@ export var MapMLFeatures = L.FeatureGroup.extend({
         // info: https://github.com/Leaflet/Leaflet/pull/4597
         L.DomUtil.addClass(this._container, 'leaflet-pane mapml-vector-container');
         L.setOptions(this.options.renderer, {pane: this._container});
-        let anim = L.DomUtil.create("style", "mapml-feature-animation", this._container);
-        anim.innerHTML = `@keyframes pathSelect {
-          0% {stroke: white;}
-          50% {stroke: black;}
-        }
-        g:focus > path,
-        path:focus {
-          animation-name: pathSelect;
-          animation-duration: 1s;
-          stroke-width: 5;
+        let style = L.DomUtil.create("style", "mapml-feature-style", this._container);
+        style.innerHTML = `
+        .mapml-vector-container g[role="link"]:focus,
+        .mapml-vector-container g[role="link"]:hover,
+        .mapml-vector-container g[role="button"]:focus,
+        .mapml-vector-container g[role="button"]:hover,
+        .mapml-vector-container g[role="link"] path:focus,
+        .mapml-vector-container g[role="link"] path:hover,
+        .mapml-vector-container g[role="button"] path:focus,
+        .mapml-vector-container g[role="button"] path:hover,
+        .mapml-vector-container g[role="link"]:focus path,
+        .mapml-vector-container g[role="link"]:hover path,
+        .mapml-vector-container g[role="button"]:focus path,
+        .mapml-vector-container g[role="button"]:hover path {
+          outline: 0!important;
           stroke: black;
         }`;
       }
@@ -310,7 +315,7 @@ export var MapMLFeatures = L.FeatureGroup.extend({
     _removeCSS: function(){
       let toDelete = this._container.querySelectorAll("link[rel=stylesheet],style");
       for(let i = 0; i < toDelete.length;i++){
-        if(toDelete[i].classList.contains("mapml-feature-animation")) continue;
+        if(toDelete[i].classList.contains("mapml-feature-style")) continue;
         this._container.removeChild(toDelete[i]);
       }
     },
--- a/src/mapml/layers/FeatureLayer.js
+++ b/src/mapml/layers/FeatureLayer.js
@@ -28,7 +28,7 @@ export var MapMLFeatures = L.FeatureGroup.extend({
         g[role="button"]:focus path,
         g[role="button"]:hover path {
           outline: 0!important;
-          stroke: #000;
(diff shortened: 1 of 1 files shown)


Malvoz:
I've committed https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/478/commits/769e145be944ce3d63ad0180ca5ac7dc4f0616d0 to use [`LinkText` system color](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#system_colors) (if supported, else fall back to black), so the videos above are not accurate in terms of the stroke color on focus/hover. This makes features more "linky" or "buttony" in terms of highlighting, and the colors may vary depend on the OS.

It'd be nice if we could _ensure_ that these "highlight strokes" always have sufficient contrast, but I'm not sure there's a proper way to do that. [`color-contrast()`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-contrast()) may help, but that's not supported in any browsers atm I think, so I'm not sure it'd work as I imagine.

prushforth:
I think this works. As you say, we'll need to support prefs, somehow.

prushforth:
I'll merge 'soon', but the PE demo will break if I merge now.

src/mapml/layers/FeatureLayer.js
```diff
@@ -13,17 +13,23 @@ export var MapMLFeatures = L.FeatureGroup.extend({
...
+        .mapml-vector-container g[role="button"]:focus,
+        .mapml-vector-container g[role="button"]:hover,
+        .mapml-vector-container g[role="link"] path:focus,
+        .mapml-vector-container g[role="link"] path:hover,
+        .mapml-vector-container g[role="button"] path:focus,
+        .mapml-vector-container g[role="button"] path:hover,
+        .mapml-vector-container g[role="link"]:focus path,
+        .mapml-vector-container g[role="link"]:hover path,
+        .mapml-vector-container g[role="button"]:focus path,
+        .mapml-vector-container g[role="button"]:hover path {
```
Malvoz, line 29: Scoping these styles to `.mapml-vector-container` may be problematic, for example they wouldn't apply to the SVG `path`/`g` elements in https://maps4html.org/experiments/vector-tiles/mapml/mapml-viewer/ because that SVG is created inside `.mapml-templatedlayer-container`.

I'll remove that scope, perhaps it's better to scope the styles `.leaflet-pane` (alternatively `.mapml-layer`?), otherwise these styles may apply inside things like `.leaflet-controls-container` which we probably don't want.

Malvoz: > they wouldn't apply to the SVG `path`/`g` elements in https://maps4html.org/experiments/vector-tiles/mapml/mapml-viewer/

While those `path`/`g` elements don't seem to be interactive, I suppose they could be. So it's safer to remove the `.mapml-vector-container` scope for now. Done in https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/478/commits/bc604f014e24ac69b8f9e04e8da947a695daf2e3.

Malvoz:
Pushed https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/478/commits/bd668e845050027713fc50438c660fce6d31c11e to use the `#0000EE` color as a fallback instead of `#000000` as it is the color specified for `:link` in [phrasing content](https://html.spec.whatwg.org/multipage/rendering.html#phrasing-content-3) under [UA presentational hints](https://html.spec.whatwg.org/multipage/rendering.html#the-css-user-agent-style-sheet-and-presentational-hints).

prushforth:
This is good work. I'm pretty sure I pushed Ahmad to include the thick black highly visible stroke on focus (maybe not the animation, though).  I do like the addition to hover. Fully agree we don't need both the stroke and the outline; the stroke alone is good information, although sometimes you might miss a teeny outlier piece of the geometry without the outline.  I also appreciate that each `<feature>` hover/focus stroke is rendered with the link color unless otherwise specified by the author, giving us the natural idea that there is something potentially link-ish about all features.

PR #486, opened 2021-07-18
Add `<featurecaption>`s to `us_pop_density.mapml`

Malvoz:
Not sure this is useful to anyone else, but I like to inspect/learn from the test files, and missing `<featurecaption>`s results in everything defaulting to "Feature" which isn't ideal.

There is a [similar file in the demo folder](https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/master/demo/us_pop_density.mapml), which I have not updated because I don't think that's used anywhere. I don't mind updating that as well if it is to be used (or is used elsewhere).

There are many other mapml files that could need some `<featurecaption>` love, may fix in separate PRs if desired.

### DIFF
--- a/test/e2e/data/tiles/cbmt/us_pop_density.mapml
+++ b/test/e2e/data/tiles/cbmt/us_pop_density.mapml
@@ -13,6 +13,9 @@
   </head>
   <body>
     <feature class="all  _50-100">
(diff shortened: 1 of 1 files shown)


prushforth:
Agree featurcaption should be standard in our examples.

Issue #376, opened 2021-03-30
Layout Shift due to no intrinsic size before elements are upgraded


Malvoz:
https://user-images.githubusercontent.com/26493779/113009091-eee43100-9177-11eb-89e0-156e328dd10c.mp4

(This Layout Shift is more notable when the map is larger, and on slower connections.)

We should give the element a (default) intrinsic size (300x150) _before_ it is upgraded (`:not(:defined)`). These styles should only apply when Custom Elements is supported _and_ JavaScript enabled such that we don't impose the styles on the map element in the graceful degradation/noscript scenario, in which case it may act as a container for any potential fallback content. And we also want to avoid overriding any author-provided CSS `width` and `height` declarations - shiny new [`contain-intrinsic-size`](https://developer.mozilla.org/en-US/docs/Web/CSS/contain-intrinsic-size) to the rescue?

[Who said styling custom elements is easy?](https://github.com/whatwg/html/issues/6231) 😋

PR #487, opened 2021-07-18
Pre-style to avoid Layout Shift

Malvoz:
Fix https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/376.

https://user-images.githubusercontent.com/26493779/126076213-e19fd082-1450-4676-aa8e-df1609facc1d.mp4

I'm only adding these styles to the "main" HTML documents, and not every HTML file in this repo because that's a bit tedious. And these are ("advanced") styles to progressively enhance the map viewer, nothing essential.

How/why to pre-style to avoid Layout Shift is to be documented, per https://github.com/Maps4HTML/web-map-doc/issues/26.

### DIFF
--- a/index-map-area.html
+++ b/index-map-area.html
@@ -36,6 +36,14 @@
       map[is="web-map"]:not(:defined) > :not(area):not(.mapml-web-map) {
         display: none;
       }
+      
+      /* Pre-style to avoid Layout Shift. */
+      map[is="web-map"]:not(:defined) {
+        display: inline-block;
+        contain: size;
+        contain-intrinsic-size: 304px 154px;
+      }
+      
       /* Ensure inline layer content is hidden if custom/built-in elements isn't
       supported, or if javascript is disabled. This needs to be defined separately
       from the above, because the `:not(:defined)` selector invalidates the entire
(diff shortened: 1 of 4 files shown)


index-map-area.html
```diff
@@ -36,6 +36,14 @@
       map[is="web-map"]:not(:defined) > :not(area):not(.mapml-web-map) {
         display: none;
       }
+      
+      /* Pre-style to avoid Layout Shift. */
+      map[is="web-map"]:not(:defined) {
+        display: inline-block;
+        contain: size;
+        contain-intrinsic-size: 304px 154px;
```
Malvoz, line 44: Just a note:

`contain-intrinsic-size: 304px 154px` may look strange, one might expect `300px 150px` (same as the default size of `width: 300px` and `height: 150px`), it is used because the map viewers have a 2px `border-width`.

I.e. `contain-intrinsic-size: 300px 150px` would still cause a small but noticable layout shift.

Malvoz:
[Firefox does not support `contain-intrinsic-size`](https://bugzilla.mozilla.org/show_bug.cgi?id=1597529) btw, which is fine. Again, this is to progressively enhance the map element(s).

index-map-area.html
prushforth: Nice work!

prushforth:
Thank you.

prushforth:
I'll have a look at the documentation issue.  Good idea.

Issue #473, opened 2021-05-21
Change `<a role="button">` controls to `<button type="button">`


Malvoz:
It's recommended to use `<button>` instead of `<a role="button">`. We should change the use of `<a>` in all places we've introduced them:
- Context menu items
- Popup navigation controls
- Reload button
- Remove Layer button
- ...?

This would require a CSS reset to make buttons look seamless by default, e.g.:

```CSS
.mapml-button {
  background-color: transparent;
  border: none;
  border-radius: 0;
  color: inherit;
  font: inherit;
  line-height: inherit;
  margin: 0;
  padding: 0;
  overflow: hidden;
  text-align: left;
  text-align: inline-start;
  text-transform: none;
  -webkit-appearance: none;
     -moz-appearance: none;
          appearance: none;
}
```

(This generic button class was originally proposed in https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/263#issuecomment-805854911.)

Info on `<a>` vs `<button>`:
- https://marcysutton.com/links-vs-buttons-in-modern-web-applications
- https://bitsofco.de/anchors-vs-buttons/
- https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20160317/examples/button/button.html

This issue relates to https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/19 (same problem but keeping the issues separate as that one pertains to Leaflet's controls).

Issue #19, opened 2016-05-03
Make zoom controls button elements since they are buttons. Buttons are for actions, links are to take you to another page.


prushforth:
This was the suggestion of @powersurge on a11yslackers.  He is correct, and perhaps this issue should be raised over at the Leaflet.js project, since the zoom control comes from there.

Malvoz:
To my understanding, anchors are perfectly valid as buttons in this instance because `role="button"` is used, the buttons have accessible labels, and they respond correctly to keyboard input. Although I'm not certain, I think the usage of `<a>` is because buttons and other form elements are generally harder to style, e.g. they don't inherit font styles by default, and CSS [appearance](https://developer.mozilla.org/en-US/docs/Web/CSS/appearance) doesn't have (or didn't have at the time) wide enough browser support to warrant its use.

Anyhow, there's a leaflet issue to tackle some accessibility issues in https://github.com/Leaflet/Leaflet/issues/3210, including moving from links to buttons.

prushforth:
I think that's good enough for me at this time. Closing; feel free to discuss if necessary.

Malvoz:
https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/19#issuecomment-578974225:
> they respond correctly to keyboard input

That statement is incorrect, see upstream issue https://github.com/Leaflet/Leaflet/issues/7369.

Malvoz:
This issue should consider not only Zoom controls but all standard Leaflet controls, such as a popup's Close button, Fullscreen control, Layer control.

PR #488, opened 2021-07-19
Change introduced `<a>` controls to use `<button>`

Malvoz:
Fix https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/473.

This change makes these particular controls operable using both <kbd>Enter</kbd> _and_ <kbd>Space</kbd>, and screen readers appropriately announce them as <q>button</q> as opposed to <q>link</q>.

Note: It seems that Firefox is using different `outline` colors for buttons and links (whereas e.g. Chrome uses the same color for both elements), as such, keyboard users may find it strange that different controls have different outline colors (because Leaflet is using `<a>` elements for controls, to be fixed separately: https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/19).

### DIFF
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -46,6 +46,29 @@
  * Controls.
  */
  
+ /* Generic class for seamless buttons */
+ .mapml-button {
+  background-color: transparent;
+  border: none;
+  border-radius: 0;
+  color: inherit;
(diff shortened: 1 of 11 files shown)


prushforth:
Nice work, thank you!

PR #489, opened 2021-07-20
Correct HTML namespaces

Malvoz:
(See https://github.com/Maps4HTML/MapML/pull/198)

@prushforth This PR only corrects current namespace declarations, but there are quite a few .mapml files that have `<mapml>` elements without a namespace declaration. Should I fix that before we proceed to merge this PR?

### DIFF
--- a/test/e2e/data/tiles/wgs84/0/r0_c0.mapml
+++ b/test/e2e/data/tiles/wgs84/0/r0_c0.mapml
@@ -1,4 +1,4 @@
-<mapml xmlns="https://www.w3.org/1999/xhtml/">
+<mapml xmlns="http://www.w3.org/1999/xhtml">
    <head>
       <title>ne_10m_admin_0_countries</title>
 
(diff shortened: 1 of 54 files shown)


test/e2e/data/tiles/wgs84/0/r0_c0.mapml
```diff
@@ -2176,15 +2176,15 @@
                   <tr><th scope="row">subunit</th><td itemprop="subunit">Ivory Coast</td></tr>
                   <tr><th scope="row">su_a3</th><td itemprop="su_a3">CIV</td></tr>
                   <tr><th scope="row">brk_diff</th><td itemprop="brk_diff">0</td></tr>
-                  <tr><th scope="row">name</th><td itemprop="name">C�te d'Ivoire</td></tr>
-                  <tr><th scope="row">name_long</th><td itemprop="name_long">C�te d'Ivoire</td></tr>
+                  <tr><th scope="row">name</th><td itemprop="name">C�te d'Ivoire</td></tr>
+                  <tr><th scope="row">name_long</th><td itemprop="name_long">C�te d'Ivoire</td></tr>
```
Malvoz, line 2180: Many "modifications" like these can be seen in this PR, I in fact did not modify these lines (and nothing actually changed AFAICT). This is certainly due to the "�" characters. Should be fixed, separately from this PR.

prushforth: That was careless of me.

prushforth:
During processing of https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/515 I discovered that by correcting the namespace on GeoGratis resources e.g. [this one](https://geogratis.gc.ca/mapml/en/cbmtile/cbmtgeom/?alt=xml), I can no longer 'force' the browser to parse the response with the XML parser, it sniffs the namespace and parses with the (X)HTML(?) parser, with the result being that it displays a very simple web page with only the text content rendered (the map-title). Not a big deal, better to have the correct namespace, but interesting.

PR #490, opened 2021-07-21
Use `hidden` attribute for image maps

Malvoz:
Looks cleaner when inspecting the DOM, and isn't (potentially) blocked by the authors' CSP directives that govern styling.

### DIFF
--- a/src/web-map.js
+++ b/src/web-map.js
@@ -248,10 +248,12 @@ export class WebMap extends HTMLMapElement {
             }
           }
 
-          // undisplay the img in the image map, because it's not needed now
-          // gives a slight fouc, not optimal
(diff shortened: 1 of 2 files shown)


prushforth:
Thanks, every little bit helps.  Were you encountering failures with the keyboardInteraction test? Curious...

Malvoz:
Yes, the failing tests of the initial commit can be seen here https://github.com/Maps4HTML/Web-Map-Custom-Element/runs/3128752752#step:7:631. I didn't think to run the tests for this simple change. _All_ other instances of `await page.waitForTimeout()` except for these failing tests were set to 1000, so changed accordingly.

PR #496, opened 2021-07-23
A few modifications encountered while working toward npm distribution.

erictheise:
Now available at https://www.npmjs.com/package/@maps4html/web-map-custom-element


prushforth:
Thanks!

PR #499, opened 2021-07-28
Inherit `box-sizing` into `<map>`

### LINKED ISSUE #350: How can we get rid of the scrollbars on the map
For some reason, scrollbars show up on some maps, but not others, and I'm wondering if we can banish them, since the primary means of navigating the map is pan/zoom provide by other  means.

Malvoz:
Per https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/350#issuecomment-888361985. Close #350.

### DIFF
--- a/src/web-map.js
+++ b/src/web-map.js
@@ -130,6 +130,7 @@ export class WebMap extends HTMLMapElement {
     `width: 300px;` +
     `border-width: 2px;` +
     `border-style: inset;` +
+    `box-sizing: inherit;` + // https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/350#issuecomment-888361985
     `}` +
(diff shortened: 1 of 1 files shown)


Malvoz:
> Per [#350 (comment)](https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/350#issuecomment-888361985).

To clarify further, and for you to see it in action, compare the experiment maps: 

https://maps4html.org/experiments/screenreader/restaurants/ (`<mapml-viewer>`) and  
https://maps4html.org/experiments/progressive-enhancement/ (`<map>`)

Both these maps share the same stylesheet, despite that, the latter has scrollbars, if you add `box-sizing: inherit` to `<map>` (as this PR suggests) they now behave the same and no scrollbars appear.

PR #500, opened 2021-07-29
Modifications to allow npm publication and unpkg content delivery.

erictheise:
The end result is that the root level of the npm tarball contains `package.json` and the `.md` files; these are not carried over into `/dist`, which should just contain `.js` and `.css` files.

Apologies for mixing whitespace-only changes in with substantive changes; they're good for consistency but I admit my IDE is aggressive about chomping trailing whitespace.


README.md
```diff
@@ -8,7 +8,19 @@ The `<mapml-viewer>` custom element is a prototype implementation of the
 defined in the MapML (Map Markup Language)
 [specification](https://maps4html.org/MapML/spec/).
 
-<!-- ## Installation -->
+## Installation
+
+Beginning with `v0.8.2`, &lt;mapml-viewer&gt; can be loaded from the [UNPKG content delivery network](https://unpkg.com/) –
+```bash
```
Malvoz, line 14: This should be JS syntax highlighting.

Malvoz (changes requested):

README.md
Malvoz: I mean, HTML!

erictheise: Good catch, thanks.

Malvoz (approved):
LGTM. This is great work @erictheise, thanks!

erictheise:
Thanks @Malvoz, I'm very happy to be contributing to this project.

Issue #262, opened 2021-01-27
CSS resize issues


Malvoz:
Setting CSS `resize` on the map element enables users to resize it:

<table>
<tbody>
<td width="450"><video src="https://user-images.githubusercontent.com/26493779/105969717-db89ec00-6088-11eb-8b7b-e5d678fb2ba4.mp4"></video>
</td>
</body>
</table>

1. The resizer handle is not always visible (perhaps due to a bug with composited descendant elements: https://bugs.chromium.org/p/chromium/issues/detail?id=1135676)
2. Should probably `invalidateSize()` after resize

I'd say this is low priority, but is unfortunate, authors would expect this to work without issues on a native element.

PR #510, opened 2021-09-09
Allow context menu to overflow the map container

### LINKED ISSUE #509: Allow context menu to overflow the map container?
If the map element is too small, the context menu is clipped which may be a usability issue. It also doesn't look great:

contextmenu-overflow-contained

Most maps aren't this small since developers override the default size of the map element, so shouldn't be an issue most of the times. But I think it'd be good to fix it, and allow the context menu to overflow the map container.

Malvoz:
Fixes #509.

Caveats:
1. Removes `overflow: hidden` on the map element. This shouldn't be a problem in terms of elements overflowing the map element because it's not really in use: all light DOM children of the map element except for the shadow host are set to `display: none` anyways, and the leaflet-container - which holds the content of the map - is itself set to `overflow: hidden`.
2. Not possible to use CSS `resize` on the map element as it requires `overflow: hidden` (though `resize` is already problematic https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/262). However I suspect the use of CSS resize on maps in general is very low.
3. Lose performance benefits of [`strict`](https://www.w3.org/TR/css-contain/#valdef-contain-strict) containment on the leaflet-container.

## Preview

contextmenu-overflowing

### DIFF
--- a/src/mapml-viewer.js
+++ b/src/mapml-viewer.js
@@ -116,12 +116,11 @@ export class MapViewer extends HTMLElement {
     mapDefaultCSS.innerHTML =
     `:host {` +
     `all: initial;` + // Reset properties inheritable from html/body, as some inherited styles may cause unexpected issues with the map element's components (https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/140).
-    `contain: size;` + // Contain size calculations within the map element.
+    `contain: layout size;` + // Contain layout and size calculations within the map element.
     `display: inline-block;` + // This together with dimension properties is required so that Leaflet isn't working with a height=0 box by default.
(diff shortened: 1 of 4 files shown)


Malvoz:
With these caveats I'm not certain this is worth it. What do you think?

prushforth:
> With these caveats I'm not certain this is worth it. What do you think?

I don't think the caveats are that big a deal, but experience may prove me wrong eventually.  I think we need to keep thinking in terms of what we want a user agent to do in the same situation. So, go for it!

Malvoz:
A native map element would probably both be resizable and allow the native context menu to overflow the element. That's the case for iframes, but to my surprise (testing in Chrome) `resize` does not seem to work on video. So this change would bring us closer to the behavior of video at least.

PR #513, opened 2021-09-21
Update fullscreen to v1.0.5

Malvoz:
See the [changelog](https://github.com/runette/Leaflet.fullscreen/blob/gh-pages/CHANGELOG.md#105).

Fixes the following item in https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/247#fullscreen-control:

> ## Fullscreen control
> - [ ] Should have `role="button"`


prushforth:
LGTM - Thanks!

Issue #511, opened 2021-09-12
Refactor MapML elements as custom elements


prushforth:
Per https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/174, it seems like a best practice of Web platform speculative polyfills is to generate proposed new APIs in a separate namespace, so that the name of the polyfills doesn't collide with the eventual standard name. 

For elements, that implies today that we should use custom elements, and if I understand correctly, probably also data- prefixed attributes. Customised built-in elements would be ideal for implementing proposed extensions of existing elements, while autonomous custom elements would better serve the use case of new elements.  There is apparently a viable polyfill for customized built-in (is=), but unsure how it works. It may not be worth the trouble of using the polyfill, opting instead for a apec explanation of the intended name of the eventual element.

This issue is for discussion of how to approach this refactoring.

We'll also need to decide how to do this in the draft spec, so that it's clear whether a proposed.element is 'new', or an extension of an existing element.

Also see https://github.com/Maps4HTML/MapML/issues/215#issuecomment-917656635

Malvoz:
> probably also data- prefixed attributes

Just noting that IMO the situation regarding "custom" attributes vs `data-` prefixed attributes (related: https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/191) is much less important than converting the MapML elements to custom elements containing a hyphen. Especially so if the attributes are used only on custom elements, since that wont "pollute" standard HTML elements. So perhaps this issue can focus solely on the elements aspect?

prushforth:
Yes, for now we'll focus on elements.  What's your opinion on the `<mapml>` element?

Malvoz:
All MapML elements must contain a hyphen, so it'd be `<mapml->`. In the spec and the explainer the elements should be referred to without hyphens.

prushforth:
Ok, I went with `<mapml->`, but everything else (except `mapml-viewer` and `layer-`) I put into the `map-` "namespace".  ok?

We'll need to explain this somehow, perhaps in the MapML spec?

Malvoz:
I don't think there's a need to use a `map-` prefix. The MapML spec should name elements as they're intended to be standardized. But the polyfill must follow the HTML custom components rules, so they're required to include a hyphen.

[...]

PR #515, opened 2021-09-22
Team mapml refactoring

prushforth:
Fixes issue #511


src/mapml/layers/MapLayer.js
```diff
@@ -1120,15 +1146,15 @@ export var MapMLLayer = L.Layer.extend({
       let extent = this._extent;
       if(!extent) return FALLBACK_PROJECTION;
       switch (extent.tagName.toUpperCase()) {
-        case "EXTENT":
+        case "MAP-EXTENT":
           if(extent.hasAttribute('units'))
             return extent.getAttribute('units').toUpperCase();
           break;
         case "INPUT":
```
ahmadayubi, line 1153: I believe this has to be updated as well since the test files use map-input rather than input now.

ahmadayubi, line 1153: I believe this has to be updated as well since the test files use map-input rather than input now.

prushforth:
@Malvoz @erictheise fyi this is a big change, hopefully we've covered it all.  Let us know if you have concerns.  We'll push the change on geogratis and experiments when this gets merged.

ahmadayubi:
Should `<layer->` be changed to `<map-layer>`?

prushforth:
> Should `<layer->` be changed to `<map-layer>`?

I believe it would be quite disruptive, and the gain would be a bit more coherence. I tend towards "no, not enough benefit", but remain open to opinions.

Malvoz:
If we want to be consistent we'll also have to change `<mapml->` to `<map-mapml>` which reads a bit funny. I personally squinted at `<layer->` the first time I saw it, while it is valid HTML it looks a bit awkward before getting accustomed to. We should consider if it may affect how developers, especially beginners, perceive them. And if we're to make the changes, they should probably happen in this PR.

I don't have a strong opinion on this, I'm mostly happy the polyfill is heading towards web platform compatibility. 😊

ahmadayubi:
Line 965, it's creating input elements for an extent, I believe they should be `<map-input>` rather than `<input>`. Although the fact that it didn't cause errors may mean it's either not a function being used anymore or it's not tested for.

prushforth:
> Although the fact that it didn't cause errors may mean it's either not a function being used anymore or it's not tested for.

I think it's this, but I'll remove that code once we've got this issue closed.  Should have done it before but tempes fugit :-)

test/e2e/data/cbmt.mapml
```diff
@@ -0,0 +1,17 @@
+<mapml- xmlns="http://www.w3.org/1999/xhtml/">
+  <map-head>
+    <map-title>Canada Base Map - Transportation (CBMT)</map-title>
+    <map-meta http-equiv="Content-Type" content="text/mapml;projection=OSMTILE"></map-meta>
+    <map-meta charset="utf-8"></map-meta>
+    <map-link rel="license" href="https://www.nrcan.gc.ca/earth-sciences/geography/topographic-information/free-data-geogratis/licence/17285" map-title="Canada Base Map © Natural Resources Canada"></map-link>
```
Malvoz, line 6: @prushforth I'm looking to reflect these changes in https://github.com/Maps4HTML/MapServer-documentation/tree/mapml, I'm wondering whether it was intended to change the `title` attribute on `<map-link>`(s) to `map-title` as well.

Malvoz: I'm assuming that was not intended, as it seems no other attributes use the `map-` prefix.

prushforth: That's correct. We only changed the elements' prefix, I'll await further information as to whether it's necessary to change attributes. Fwiw I think the link title attribute would be appropriate if there's a conflict there that might be motivation to adopt map-title. We haven't done that yet, so I think it's best to keep with title for now.  Btw I can't find the mapserver MapML code, I was looking to update it too. Lete know if you come across it.

Malvoz: MapServer doesn't actually support MapML yet, here's the implementation ticket: https://github.com/MapServer/MapServer/issues/6024.

Issue #519, opened 2021-09-27
Context menu keyboard interactions in Firefox do not work


ben-lu-uw:
i.e. right clicking on the map and pressing D does not toggle debug mode in Firefox but it does in Chrome.
Perhaps similar to issue https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/465.
event.keyCode is deprecated but is still supported by browsers.

ben-lu-uw:
The usage of event.path is unsupported in Firefox. Using event.composedPath is the standard and fixes this issue and https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/465.

Issue #465, opened 2021-05-18
Escape key does not dismiss popup in Firefox


prushforth:
... but it does in Chrome.  

Possibly related to one of [these](https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/247#popups), or a regression/omission perhaps.

Malvoz:
Yes that'd be this item:

> - [x] When focus is inside of a popup, pressing <kbd>Esc</kbd> should close the popup. Fixed in #292.

I must've only tested in Chrome...

Issue #469, opened 2021-05-19
Focus is not delegated after opening context menu in Chrome


Malvoz:
Pressing <kbd>Shift</kbd> + <kbd>F10</kbd> to open the context menu does not send focus to the context menu (or first menu item within) in Chrome.

ben-lu-uw:
https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/4c452ef9879d636e843cd3174ce3bfde4695fc95/src/mapml/handlers/ContextMenu.js#L436
event.button is 0 in Firefox as expected when pressing shift + F10 but -1 in Chrome. I am unsure as to why they are different. 
https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button

prushforth:
This issue didn't automatically close by merging the PR, so closing manually.  I know there's a way to make that happen.  Maybe with "Close #469" or something like that.

Malvoz:
> This issue didn't automatically close by merging the PR

> Maybe with "Close #469"

Indeed, the syntax is `KEYWORD #ISSUE-NUMBER`, keywords are listed here: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #520, opened 2021-09-28
Use e.composedPath in addition to e.path for firefox compatibility

ben-lu-uw:
Fix for issues https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/519 and https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/465

### DIFF
--- a/src/mapml/handlers/ContextMenu.js
+++ b/src/mapml/handlers/ContextMenu.js
@@ -529,7 +529,10 @@ export var ContextMenu = L.Handler.extend({
     if(!this._mapMenuVisible) return;
 
     let key = e.keyCode;
-    if(key !== 16 && key!== 9 && !(!this._layerClicked && key === 67) && e.path[0].innerText !== "Copy Coordinates (C) >")
+    let path = e.path || e.composedPath();
+    console.log(path + e.path + e.composedPath());
+
+    if(key !== 16 && key!== 9 && !(!this._layerClicked && key === 67) && path[0].innerText !== "Copy Coordinates (C) >")
       this._hide();
     switch(key){
       case 32:  //SPACE KEY
--- a/src/mapml/handlers/ContextMenu.js
+++ b/src/mapml/handlers/ContextMenu.js
@@ -537,9 +537,6 @@ export var ContextMenu = L.Handler.extend({
       this._hide();
     switch(key){
       case 13:  //ENTER KEY
-        if(this._map._container.parentNode.activeElement.parentNode.classList.contains("mapml-contextmenu"))
-          this._map._container.parentNode.activeElement.click();
-        break;
       case 32:  //SPACE KEY
         if(this._map._container.parentNode.activeElement.parentNode.classList.contains("mapml-contextmenu"))
           this._map._container.parentNode.activeElement.click();
--- a/src/mapml/handlers/ContextMenu.js
+++ b/src/mapml/handlers/ContextMenu.js
@@ -531,9 +531,15 @@ export var ContextMenu = L.Handler.extend({
     let key = e.keyCode;
     let path = e.path || e.composedPath();
 
+    if(key === 13)
+      e.preventDefault();
     if(key !== 16 && key!== 9 && !(!this._layerClicked && key === 67) && path[0].innerText !== "Copy Coordinates (C) >")
       this._hide();
     switch(key){
+      case 13:  //ENTER KEY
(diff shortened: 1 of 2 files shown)


Malvoz:
I'm seeing an issue in mapml-viewer.html where <kbd>Enter</kbd> does not activate menu items (whereas <kbd>Space</kbd> does), although it works in featureLinks.html (`<map>`).

Additionally, in both maps <kbd>Space</kbd> and the <kbd>V</kbd> shortcut to View source is now blocked with the message: "Firefox prevented this site from opening a pop-up window.", although <kbd>Enter</kbd> does work to View source.

Malvoz (commented):
I'm seeing an issue in mapml-viewer.html where <kbd>Enter</kbd> does not activate menu items (whereas <kbd>Space</kbd> does), although it works in featureLinks.html (`<map>`).

Additionally, in both maps <kbd>Space</kbd> and the <kbd>V</kbd> shortcut to View source is now blocked with the message: "Firefox prevented this site from opening a pop-up window.", although <kbd>Enter</kbd> does work to View source.

prushforth:
> > Additionally, in both maps Space and the V shortcut to View source is now blocked with the message: "Firefox prevented this site from opening a pop-up window.", although Enter does work to View source.
> 
> * I guess Firefox is more pick about window.open and popups compared to Chrome. Will try to find a work around

That is pretty odd that ff allows the "popup" (new tab) when one uses "Enter" vs the shortcut "V". Could it be that they are mapped to different ways of doing the action?

I actually have another issue in Chrome: when you activate the context menu via Shift-F10 (or right-click), the first menu item is not focused. In FF, the first menu item **is** focused when you Shift-F10, but not when you right-click.   Maybe that's worthy of a separate issue.

Malvoz:
> I actually have another issue in Chrome: when you activate the context menu via Shift-F10 (or right-click), the first menu item is not focused. In FF, the first menu item is focused when you Shift-F10, but not when you right-click. Maybe that's worthy of a separate issue.

See https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/469. I don't think the expectation is for focus to be delegated on right click however.

ahmadayubi:
I also had the issue where `Space` activated the menu items but `Enter` did not

ahmadayubi:
One solution to this is to manually configure the behavior for `Enter` (keyCode 13), it'd have the same behavior of `Space`. You'd have to preventDefault though. (This is just one of many solutions, feel free to go with whatever you find best)

In ContextMenu.js line 528
```js
let key = e.keyCode;
let path = e.path || e.composedPath();
if(key === 13) e.preventDefault(); //Add this line
if(key !== 16 && key!== 9 && !(!this._layerClicked && key === 67) && path[0].innerText !== "Copy Coordinates (C) >")
  this._hide();
switch(key){
  case 13: //And this line
```

ben-lu-uw:
Giving the enter key the same behavior as the space button resolves the menu issue however, hitting the enter button over the view source option leads to the "popup" being blocked, just like space and v. I believe its because its under the keydown event  : https://stackoverflow.com/a/6132481

ahmadayubi:
Right that was only for `Enter`, for the popup issue, if Firefox is blocking it one work around could be to add a `<a href="link to goto">` to the DOM, click it using element.click() then removing that temporary `<a>` from the DOM.

A similar thing is done for copying to clipboard. Where a temporary element is added to the DOM, you interact with that elements API, then remove the temporary element.

ben-lu-uw:
>  if Firefox is blocking it one work around could be to add a <a href="link to goto"> to the DOM, click it using element.click() then removing that temporary <a> from the DOM.

Tried this and other work arounds but they all get blocked by Firefox when trying to open a new tab.  From what I have read, the user has to explicitly click themselves in order for the new tab to open or the user has to allow popups for the site.

ahmadayubi:
I guess they're really strict about popups. The other idea I had in mind was moving that function to the mapml-extension. Web extensions have a contextmenu API to add to the built in context menu, it would be a big undertaking though.

But like you mentioned the user always has the option of loosening their popup blocking settings in Firefox.

ahmadayubi (commented):
You can actually merge cases in JavaScript, something like this works as well:

```js
case 13:
case 32:
    if(....
    ...
break;

```

Malvoz (approved):

ahmadayubi (approved):

PR #523, opened 2021-09-29
Fix issue https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/469

ben-lu-uw:
Fix for issue https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/469 / https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/520#issuecomment-929346325
Just a quick work around and maybe not best practice? lmk.


Malvoz (approved):

prushforth:
My comment seems to have disappeared, so here it is, possibly for the second time:

This looks good to me!  My only comment is that the title of the PR should be descriptive, with the comment exactly as you have it.  I believe the issue should close automatically if the comment is correctly phrased. We'll  see!

PR #525, opened 2021-10-06
Get rid of old example. Refactor blog mapml into XHTML namespace.

prushforth:
Refactor MapML into map- custom elements. Unignore dist folder to update 
viewer with latest version.


prushforth:
@Malvoz  a bit of clean up. I deleted some stuff that might have been linked to from somewhere (the index page of this repo's gh-pages, for example).  Once I move this blog post somewhere, I believe we should delete the whole branch (use experiments for any/all demos).

PR #531, opened 2021-10-08
Refactoring HTML anatomy of items in layers

### LINKED ISSUE #263: HTML anatomy of items in the layers list needs refactoring

Anshpreet8:
Used the svg "More Vert" icon instead of the details element to avoid having to nest interactive elements inside the summary element, as the suggested layout in the [proposal for alternative layer item anatomy](https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/263#issuecomment-805854911)

Essentially the same PR as https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/530, which was closed due to renaming the branch to refactoringLayer

List of Issues that can be closed from this PR:
Closes #263

### DIFF
--- a/src/mapml/layers/MapLayer.js
+++ b/src/mapml/layers/MapLayer.js
@@ -482,17 +482,16 @@ export var MapMLLayer = L.Layer.extend({
     },
     getLayerUserControlsHTML: function () {
       var fieldset = L.DomUtil.create('fieldset', 'mapml-layer-item'),
-        input = document.createElement('input'),
+        input = L.DomUtil.create('input'),
         layerItemName = L.DomUtil.create('span', 'mapml-layer-item-name'),
-        buttonNameIcon = document.createElement('span'),
-        settingsButtonNameIcon = document.createElement('span'),
+        settingsButtonNameIcon = L.DomUtil.create('span'),
         layerItemProperty = L.DomUtil.create('div', 'mapml-layer-item-properties', fieldset),
         layerItemSettings = L.DomUtil.create('div', 'mapml-layer-item-settings', fieldset),
         itemToggleLabel = L.DomUtil.create('label', 'mapml-layer-item-toggle', layerItemProperty),
         layerItemControls = L.DomUtil.create('div', 'mapml-layer-item-controls', layerItemProperty),
         opacityControl = L.DomUtil.create('details', 'mapml-layer-item-opacity', layerItemSettings),
-        opacity = document.createElement('input'),
-        opacityControlSummary = document.createElement('summary'),
+        opacity = L.DomUtil.create('input'),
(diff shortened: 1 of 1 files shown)


ahmadayubi:
@Malvoz was there a specific reason for using SVG icons rather than HTML icons? Like the ones used for the zoom, reload and full screen buttons?

Malvoz:
@ahmadayubi

HTML characters can be problematic because they tend to render differently depending on the browser/OS.

They're also much more likely to be unintentionally affected by author-provided styles such as `font-size` and `letter-spacing`, which is why we [reset styles](https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/14375f26c641a03c74cb798856d63a9c232caa27/src/mapml-viewer.js#L118) (see how the zoom controls look with a 5px `letter-spacing` in https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/140).

ahmadayubi:
@Malvoz That makes sense but how differently could the `x` symbol, for example, render across browsers, it would seem like a pretty consistent symbol.

We used it in the past, and leaflet uses them too, I wonder if it would be better to stick to using them simply for consistency. Also having to generate SVG in JavaScript isn't the most developer friendly code to read either.

PR #532, opened 2021-10-14
Sets M.options based on defaults or the content of map-options element

### LINKED ISSUE #526: Add M.options for setting default options
The M.options parameter will set defaults that will be used to control user preferences, the state of these preferences can be maintained through the [mapml-viewer extension](https://github.com/Maps4HTML/mapml-extension).

ahmadayubi:
Sets M.options using the defaults, before defaulting it looks for a `<map-options>` element that is created by the mapml extension. 
This is set before the creation of any map on the webpage to allow the options to be used in the construction of maps.

Related to PR https://github.com/Maps4HTML/mapml-extension/pull/12 and https://github.com/Maps4HTML/mapml-extension/pull/16

Closes #526


prushforth:
Looks good, sorry it took me so long to get to it.

Issue #396, opened 2021-04-14
Meaningful Pan & Zoom (and Feature navigation)


Malvoz:
A developer reached out to me some weeks ago, and wrote <q>I found you through your contributions to Leaflet and Maps4HTML</q>, <q>I'm trying to decide on an open-source map library and one of my priorities is accessibility</q>, <q>Would you be willing to provide some advice regarding the state of accessibility with these mapping libraries?</q>.

While I have a pretty good understanding of the state of accessibility in some of the most popular web maps, it was saddening I couldn't recommend even a single accessible open-source web map. Not only do _all_ of the libraries fail to implement a basic UI such as buttons in an accessible manner (i.e. https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/247), but panning and zooming is for the most part meaningless.

Can pan and zoom (and map navigation as a whole) be made useful and meaningful, to everyone?

## The problems
1. **Panning & zooming is not meaningful to users of assistive technology**
    - There is no feedback provided to the user during/after pan and zoom, so these actions have no meaning to a blind user.
    - Panning and zooming has (mostly) no effect on the visibility/focusability of elements, which leads to the next problem:
2. **Navigating only features of interest**
    - It's not possible to only navigate features currently in view (whether "currently in view" is limited to a reticule/square in the center of the map or the visible content of the map viewer itself). Currently, if only 1 out of 100 focusable elements are in view there are still 100 elements in the tab order.

## Potential solution(s)
  1. **The direction in which the user is panning should be announced**
       - _Example:_ pressing <kbd>→</kbd> should announce <q>Moving East</q> (although cardinal directions are not relevant in [_drawings or schematics without geographic coordinates_](https://maps4html.org/HTML-Map-Element-UseCases-Requirements/#use-case-technical-drawings)).
      - _Why?_ This would allow blind users to learn in which direction a feature is located in relation to others.
      - _How?_  A (potentially [screen reader only](https://github.com/ampproject/amphtml/blob/3249e6b098c84066ce0859a651cc31b665a6ac02/css/ampshared.css#L217-L247)) ARIA [Live Region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) element (e.g. [`role="status"`](https://www.w3.org/TR/wai-aria-1.2/#status) with [`aria-atomic="true"`](https://www.w3.org/TR/wai-aria-1.2/#aria-atomic)):<br><br>
[...]

PR #533, opened 2021-10-14
Announce zoom and move and implementing map boundaries 

ben-lu-uw:
This implements meaningful pan and zoom from issue https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/396 and "bounce back" when max/min pan has been reached. 
As of the making of this draft this functionality has only been implemented for mapml-viewer + templatedtilelayers and only been tested on tms.html. Currently looking for feedback.
Known issues:
- [ ] When the map is dragged out of bounds the return behavior is inconsistent
- [ ] Max/min zoom announcement has only been implemented with +/- keys 
- [ ] Have not considered the behavior when multiple layers are present

### DIFF
--- a/src/mapml-viewer.js
+++ b/src/mapml-viewer.js
@@ -100,7 +100,6 @@ export class MapViewer extends HTMLElement {
   constructor() {
     // Always call super first in constructor
     super();
-    
     this._source = this.outerHTML;
     let tmpl = document.createElement('template');
     tmpl.innerHTML =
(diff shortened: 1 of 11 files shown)


prushforth:
@Malvoz why don't links announce where they go?  For map links, I think it would be good to know whether  the link will navigate the page or the map, for [example](https://maps4html.org/experiments/painting/).  Do web links not have that notion?  For visual users, we see the destination of the link in the bottom left of the screen/map, so there's some sense that we're not going to hell by clicking on the link.  Do non-visual users not get that same assurance somehow?

Malvoz:
@prushforth To my knowledge links are always just announced "link" along with their accessible name/desc, even when they're `target="_blank"` for example. So no, non-visual users do not get that same assurance, I believe. Which is bad for screen reader users, but good for us since we're using fake links and would otherwise have had to replicate and translate that language.

prushforth (approved):
I take it you have to reload the map for the extension setting to take effect on announcing/not announcing map location.  In general the experience is what I expect, more or less; a few extra things that I don't recognized are announced.   In this [example page](https://maps4html.org/experiments/change-projection/projection-negotiation/), when you tab to the second map, initially the location isn't announced.  rows and columns are projection specific, we may need to figure out improvements incrementally.  For example, as you remove layers from the map, the bounds should change. 

This is a big change @ben-lu-uw , thank you for your hard work on it. Merging.

PR #544, opened 2021-10-27
New layer control icon

Malvoz:
This is a proposal to change the icon of the layer control:

The current one screams "Leaflet!" - which isn't a bad thing - though a new icon may encourage developers to dig into the source code and realize the component is something a little bit different, if they weren't aware already.

I also don't think UA implementations of maps would use such themed/opinionated and dated looking (IMO) icons. And the proposed one looks similar in style to the other controls.

## Preview

(icon source: https://fonts.google.com/icons?selected=Material+Icons&icon.query=layers)

### DIFF
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -235,6 +235,11 @@
   border-bottom: 1px solid #e3e3e3;
 }
 
+.leaflet-control .leaflet-control-layers-toggle {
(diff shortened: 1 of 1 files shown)


Malvoz:
I've been playing around with both icos for a while and feel like the filled icon is way too distracting. So pushed https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/544/commits/4bc7608764fe81e2fca75f37f70d26eb7c6345e3 to use the outlined version instead. Let's give it a try, we can always change later...

Issue #553, opened 2021-10-31
Link colors are not reverted to UA styles correctly


Malvoz:
Introduced in https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/281:

![current-popup-link-color](https://user-images.githubusercontent.com/26493779/139593746-104fb57b-fc5a-49fe-9431-e6434ad2341c.png)
![current-legendlink-color](https://user-images.githubusercontent.com/26493779/139593751-d71721dc-b10d-4e4b-b4c6-d2159b14f20d.png)

PR #554, opened 2021-10-31
Fix link colors

Malvoz:
Fix https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/553.

## Preview

popup-link
legendlink-color

### DIFF
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -35,13 +35,6 @@
   font-size: 12px;
 }
 
-/* Remove opinionated color of links. */
-.leaflet-container a,
-.leaflet-container a.leaflet-popup-close-button,
-.leaflet-container a.leaflet-popup-close-button:hover {
-  color: initial;
-}
-
 /*
  * Controls.
  */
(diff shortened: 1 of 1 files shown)


src/mapml.css
```diff
@@ -63,6 +56,23 @@
...
+.mapml-reload-button,
+.leaflet-control-container a,
+.leaflet-container a.leaflet-popup-close-button,
+.leaflet-container a.leaflet-popup-close-button:hover {
+  color: #000;
+}
+/* Revert above `color: #000` for legendlinks. */
+.mapml-layer-item-name a {
+  color: revert;
+}
```
Malvoz, line 75: FYI: We wouldn't need most of these (and in other places where we "fix" Leaflet styles) if we could just edit the source leaflet.css file - in this case to simply remove the opinionated color they apply to links - but that's got its own set of problems I suppose...

Malvoz: Managing styles would also be simpler if Leaflet just used buttons for buttons, and not links for buttons.

PR #555, opened 2021-11-02
Move pre-styling in some examples so that comments make sense

Malvoz:


index-map-area.html
```diff
@@ -31,19 +31,19 @@
...
       
+      /* Pre-style to avoid FOUC of inline layer- and fallback content. */
+      map[is="web-map"]:not(:defined) + img[usemap],
+      map[is="web-map"]:not(:defined) > :not(area):not(.mapml-web-map) {
+        display: none;
+      }
+      
       /* Ensure inline layer content is hidden if custom/built-in elements isn't
       supported, or if javascript is disabled. This needs to be defined separately
       from the above, because the `:not(:defined)` selector invalidates the entire
```
Malvoz, line 49: The comment: "This needs to be defined separately from the above" now correctly refers to the other `display: none` declaration, which is now directly above. The order of styles should have been considered in https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/487.

PR #564, opened 2021-11-04
Remove fullscreen control dependency, make mapml-viewer / web-map the fullscreen element

### LINKED ISSUE #256: CSS `:fullscreen` is not applicable to the map element(s)
Currently, it is the `.leaflet-container` that goes into fullscreen mode when activating fullscreen, which means it isn't possible to change e.g. `background-color` of the map element using the [`:fullscreen`](https://developer.mozilla.org/en-US/docs/Web/CSS/:fullscreen) pseudo class on `web-map`/`mapml-viewer`.

I'd say this is low priority, but is unfortunate, authors would expect this to work on a native element.

prushforth:
Fixes #256.

### DIFF
--- a/src/mapml/control/FullscreenButton.js
+++ b/src/mapml/control/FullscreenButton.js
@@ -1,0 +1,160 @@
+export var FullscreenButton = L.Control.extend({
+        options: {
+            position: 'topleft',
+            title: {
+                'false': 'View fullscreen',
+                'true': 'Exit fullscreen'
+            }
+        },
+
+        onAdd: function (map) {
+            var container = L.DomUtil.create('div', 'leaflet-control-fullscreen leaflet-bar leaflet-control');
+
+            this.link = L.DomUtil.create('a', 'leaflet-control-fullscreen-button leaflet-bar-part', container);
+            this.link.href = '#';
+            this.link.setAttribute('role', 'button');
+
+            this._map = map;
+            this._map.on('fullscreenchange', this._toggleTitle, this);
+            this._toggleTitle();
+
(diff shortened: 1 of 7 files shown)


prushforth:
Probably wouldn't hurt to write a test or two.

src/mapml/control/FullscreenButton.js
```diff
@@ -0,0 +1,160 @@
+export var FullscreenButton = L.Control.extend({
```
ahmadayubi, line 1: Do you think crediting the original plugin developer is required in this case?

prushforth: I looked at the repo, it's a fork of a leaflet repo, so same license as that. So, I think we give that credit already?

Malvoz (approved):
Generally LGTM.

Some of the fullscreen styles do no longer apply because it is the shadow host element the `leaflet-fullscreen-on` class is set on, resulting in no icon switch when going into fullscreen. The selectors would have to be changed to something like `:host(.leaflet-fullscreen-on) .leaflet-control-fullscreen a {...}`. Feel free to merge and I can look into it later.

Edit: https://developer.mozilla.org/en-US/docs/Web/CSS/:host()

Malvoz:
@prushforth I've pushed https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/564/commits/53ed94a5c8c5435bdc634aa265a63800f08c6882 to fix the issue with the fullscreen control icon not switching properly, as described above in https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/564#pullrequestreview-798249107. Please take a look. 🙂

Issue #505, opened 2021-08-08
RTL issues


Malvoz:
https://maps4html.org/experiments/i18n/rtl/

- Context menu:
  - [ ] menu items not aligned right-to-left
  - [ ] keyboard shortcut annotations on wrong side
- Tooltips (https://github.com/Leaflet/Leaflet/issues/7201):
  - [ ] offset/incorrect position
  - [ ] arrow on tooltips is on the wrong side
- Layer control:
  - [ ] <q>Remove layer</q> button on wrong side

<hr>

Related issue: https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/495.

PR #565, opened 2021-11-07
Alignment improvements in the Layer control

Malvoz:
- [x] https://github.com/Maps4HTML/Web-Map-Custom-Element/commit/a056aa0a8f9bf903e89083ceec62cb047feb844f: re-use CSS to avoid duplicating `user-select` styles
- [x] https://github.com/Maps4HTML/Web-Map-Custom-Element/commit/ed8687db9e4438303ad0d1008bb9fe1fa06e895e: align the checkbox with other controls (also removes unnecessary `width` and the need for an `!important` statement)<br>
  **Before:**
input-not-aligned
  **After:**
input-aligned
- [x] https://github.com/Maps4HTML/Web-Map-Custom-Element/commit/b4f5afe1e74c8ec7981a68ca9418ebffdaa42fee: remove `text-align` on the layer name (because it 1: doesn't appear to be needed, and 2: doesn't look good if the map is set to RTL direction but the layer name is in an LTR script)<br>
  **Before:**
rtl-with-text-align
  **After:**
rtl-without-text-align
(i.e. now looks good independently of the direction of the map and/or the script used in the layer name, tested using [sample arabic text](https://r12a.github.io/scripts/tutorial/summaries/arabic))
- [x] https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/565/commits/c95d4e36fa17f1ef30890f557f07360b1454812f: removes the indentation/margin for better alignment and improves alignment between labels and radio inputs (also fixes an [RTL issue](https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/505))<br>
**Before:**
controls-alignment-before
**After:**
controls-aligntment-after
- [x] https://github.com/Maps4HTML/Web-Map-Custom-Element/commit/b6f59f0289f901a3549a73fcc89e54cc01e32c27: add padding to the layer name (see [screenshot](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/565#issuecomment-968199589) below)

### DIFF
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -213,7 +213,8 @@
   display: flex;
 }
 
-.mapml-layer-item-style input {
+.mapml-layer-item-style input,
+.mapml-layer-item-opacity input {
   margin-inline-start: 0;
 }
 
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -385,7 +385,8 @@
 .leaflet-control,
 .mapml-contextmenu,
 .mapml-debug,
-.mapml-focus-buttons {
+.mapml-focus-buttons,
+.mapml-layer-item-settings summary {
   -webkit-user-select: none;
   -moz-user-select: none;
   -ms-user-select: none;
@@ -691,11 +692,6 @@ button.mapml-button:disabled,
   padding-block-end: .25rem;
   padding-inline-start: 2rem;
   padding-inline-end: 1rem;
-}
-
-.mapml-layer-item-settings summary {
-  -webkit-user-select: none;
-          user-select: none;
 }
 
 .mapml-layer-item-opacity [type="range"] {
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -203,11 +203,18 @@
   display: inline;
 }
 
-.leaflet-control-layers-overlays > fieldset > .mapml-layer-item-settings > .mapml-control-layers > summary ~ * {
+.mapml-control-layers > :not(summary) {
   display: block;
-  margin-top: 5px;
-  margin-left: 15px;
-  width: calc(100% - 15px);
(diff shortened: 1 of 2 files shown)


ahmadayubi (approved):
I like the alignment changes as it makes it more consistent but I guess it's up to preference.

Malvoz:
> `c95d4e3`: Removes the indentation/margin for better alignment

The original indentation was intentional, but I don't remember if not indenting the opacity slider was intentional or not. I think the logical thing is to either indent all input controls or don't indent any. While I think indenting looks better when you expand all details elements, it kind of looks strange with only having the opacity slider visible:

nested-indentations

as opposed to:

no-indentation

It's not a big deal I guess, but if you prefer one thing over the other let me know. 🙂

prushforth:
I prefer the indented version. Thanks!

Malvoz:
@prushforth I've indented all the settings (including opacity slider) in https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/565/commits/15d3ed4e2fa828c84ddc7c611ac79950c34e2d4e:

indented-inputs

Just let me know if there's anything else. 👍

prushforth:
Looks good!

Malvoz:
While I'm at it, pushed https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/565/commits/b6f59f0289f901a3549a73fcc89e54cc01e32c27 to give the layer item name some padding in the block direction (`top`/`bottom` in this writing mode):

Before:

layer-name-padding-before

After:

layer-name-padding-after

(the effect of the padding is only visible if the layer name is long enough / layer name's container is small enough that the layer name wraps and exceeds the height of the container)

Issue #550, opened 2021-10-29
Clicking +/- keys still fires moveend event when zoom in/out is disabled


ben-lu-uw:
. . . unlike the on screen buttons and the scroll wheel

prushforth:
I am having difficulty reproducing this issue.  What branch should I be looking at to listen to the bug, perhaps [this one](https://github.com/ben-lu-uw/Web-Map-Custom-Element/tree/announce_zoom_and_move)?  So, if I listen to the [painting](https://maps4html.org/experiments/painting/) with the unfixed branch, should I experience this issue?

To solve the issue I think I should merge [this branch](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/567) into the head of the above?  Is that correct?

ben-lu-uw:
This happens on any branch. If you console log here you should be able to see that moveend is being called. 
https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/a086e23baaab557179b7a66ceb5ce2ae769f4f83/src/mapml-viewer.js#L443-L449

prushforth:
@ben-lu-uw can this be merged?

Malvoz:
Fixed in https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/567.

(it's the keyword syntax thing, mentioned in https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/469#issuecomment-930502685 😋)

PR #567, opened 2021-11-08
Fix moveend firing when zooming out/in at min/max

ben-lu-uw:
Closes issue https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/550

Redefines onKeyDown to have this line
https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/5c649039b6cace1ee0303f4540a6c9ae49eca2cb/src/mapml/keyboard.js#L38

### DIFF
--- a/src/mapml/keyboard.js
+++ b/src/mapml/keyboard.js
@@ -25,7 +25,7 @@ L.Map.Keyboard.include({
             if (!map._panAnim || !map._panAnim._inProgress) {
                 offset = this._panKeys[key];
                 if (e.shiftKey) {
-                    offset = toPoint(offset).multiplyBy(3);
+                    offset = L.point(offset).multiplyBy(3);
                 }
 
                 map.panBy(offset);
--- a/src/mapml/keyboard.js
+++ b/src/mapml/keyboard.js
@@ -1,0 +1,50 @@
+L.Map.Keyboard.include({
+    _onKeyDown: function (e) {
+
+        if (e.altKey || e.ctrlKey || e.metaKey) { return; }
+
+        let zoomIn = {
+            187: 187,
+            107: 107,
+            61: 61,
+            171: 171
+        };
+
+        let zoomOut = {
+            189: 189,
+            109: 109,
+            54: 54,
+            173: 173
+        };
+
+        var key = e.keyCode,
+            map = this._map,
+            offset;
+
+        if (key in this._panKeys) {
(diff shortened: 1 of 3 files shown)


src/mapml/keyboard.js
```diff
@@ -0,0 +1,50 @@
...
+        } else if (key === 27 && map._popup && map._popup.options.closeOnEscapeKey) {
+            map.closePopup();
+
+        } else {
+            return;
+        }
+
+        stop(e);
+    }
+});
```
prushforth, line 50: I have seen this a few times, always wondered what the significance was, so I [googled it.](https://stackoverflow.com/questions/5813311/whats-the-significance-of-the-no-newline-at-end-of-file-log)  I guess we should try to add newlines at end of file where appropriate, i.e. a blank line at the end of the file

prushforth (commented):
Question: this keyboard include looks interesting - did you find an example of this in Leaflet?  Could we use this technique to provide modifiers for the arrow keys when they are Shift-arrow or Ctrl-Shift-arrow for instance?  I have always thought it would be good to provide the ability to reduce or increase the pan amount by providing such modifier key mappings.  Anyway looks good, code is well organized to my eye.  I tried it and could not reproduce the problem, but let's review together and you can show me where I'm missing something.

ben-lu-uw:
We can change the modifier for the pan here except my redefine seems to have broken it since toPoint is not defined :
https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/5c649039b6cace1ee0303f4540a6c9ae49eca2cb/src/mapml/keyboard.js#L24-L31

prushforth:
Sorry, I meant can **this** be merged?

ahmadayubi:
> Question: this keyboard include looks interesting - did you find an example of this in Leaflet? Could we use this technique to provide modifiers for the arrow keys when they are Shift-arrow or Ctrl-Shift-arrow for instance?

Some leaflet documentation that may come in use for this functionality are:

Panning Delta - https://leafletjs.com/reference.html#map-keyboardpandelta

Zooming Delta - https://leafletjs.com/reference.html#map-zoomdelta

src/mapml/keyboard.js
prushforth: LGTM! Thanks @ben-lu-uw

prushforth (approved):

Issue #403, opened 2021-04-15
Replace popup implementation with popup event


prushforth:
In discussions with @shepazu, he convinced us that an interactive popup for maps is likely to be one of the most difficult things to negotiate with browser developers, since it is a complex topic with accessibility requirements that likely needs to extend beyond maps use cases.  Furthermore, there are discussions happening in the [HTML standard](https://github.com/whatwg/html/issues/6349) that suggest a popup might become an element supported by browsers independent of maps, so maps could work towards using the results of that work and not duplicating it.

So, in accordance with that thinking, this issue is for discussion of the idea of removing a map-based popup, and replacing it with a popup event.  The idea would be that the function that handles the event could create a UI appropriate to its needs, a sidebar, a toast, a popup, whatever was considered appropriate by the author.   In this project we could develop a small library to handle the popup event to fit the needs of a demonstration.

Further, it may provide us opportunities for enhancing the interactivity of features returned by query links, for example, which could be used not only for popups (as they currently are, exclusively) but also for links.  Currently a feature geometry can be a link, or it can have a popup, but not both; we are [working on](https://github.com/Maps4HTML/geoserver/issues/40) a way to return a feature with a link (by default there's currently always a popup).

prushforth:
When we do move to this mode, we will be in a better position to support [Feature Indexing](https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/406), in that feature captions that are loaded into the index can be either linked features OR popup-creators, either way the user will expose more information when activating.  Currently every feature returned by the reticle is a popup creator.

prushforth:
@erictheise  have you noticed this issue?  The idea here is to simplify the overall MapML proposal by removing the need to propose UI that is associated to popups and replacing that in the proposal with an appropriate map `popup` event(s) API, if possible. I was persuaded in my discussions with @shepazu (not through Doug's explicit efforts to persuade me, but by my own conclusion) that negotiating maps in HTML will be made too hard /too big by bundling more requirements into the one proposal than is strictly necessary.  Many others have asserted variations on this theme, by saying the overall proposal should be broken down into "primitves".  As such, I think the `popup` event API may well be such a primitive.

[...]

PR #569, opened 2021-11-12
Auto issue query on moveend, add test

### LINKED ISSUE #405: Make reticle automatically issue query on move end
Currently queries only occur when the user hits enter or spacebar.  In moving this feature towards a more accessible future, I believe it is appropriate to issue the query on move end when the reticle is deployed.

When we move to a square reticle with [feature index](https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/397), we'll perhaps be in a position to merge our treatment of tabbing through features already in the map layer and querying features from the server.

ahmadayubi:
https://user-images.githubusercontent.com/55214462/141503780-b049a4a2-348a-4a9c-97b5-0831d0c095b7.mp4

I'm not sure if this is a valuable feature but was an issue in the repo, try it out to see if the functionality is useful, and provide feedback.

Closes #405

### DIFF
--- a/src/mapml/layers/Crosshair.js
+++ b/src/mapml/layers/Crosshair.js
@@ -35,6 +35,7 @@ export var Crosshair = L.Layer.extend({
 
     map.on("layerchange layeradd layerremove overlayremove", this._toggleEvents, this);
     map.on("popupopen", this._isMapFocused, this);
+    map.on("moveend", this._moveEndQuery, this);
     L.DomEvent.on(map._container, "keydown keyup mousedown", this._isMapFocused, this);
 
     this._addOrRemoveCrosshair();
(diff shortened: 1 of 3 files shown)


Malvoz:
I don't think we should do this, see https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/405#issuecomment-893317493.

ahmadayubi:
> I don't think we should do this, see [#405 (comment)](https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/405#issuecomment-893317493).

I lean towards this as well, although once popup's are removed it may make more sense.

Malvoz:
Then perhaps it's best to revisit this once https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/403 is resolved, and close this (and maybe the corresponding issue) for now? @prushforth @ahmadayubi

PR #570, opened 2021-11-12
Prevent link activation on right click

### LINKED ISSUE #477: Right clicks shouldn't activate `map-a` links
Go to https://maps4html.org/experiments/custom-projections/Arctic-SDI/ for example, and right click (on any country I guess) to open the context menu. The link is activated and the context menu closes.

ahmadayubi:
Closes #477

This raises the question of what should show on right click of features, it could be a context menu with feature specific items, or just the general context menu.


Malvoz (approved):

Malvoz:
> This raises the question of what should show on right click of features, it could be a context menu with feature specific items, or just the general context menu.

I think showing the general context menu would be good, so that features don't block users from reaching it.

PR #573, opened 2021-11-13
Fix select in Firefox

### LINKED ISSUE #572: Time `<select>` not working in Firefox
Repro steps:

1. Open https://maps4html.org/experiments/datacube/ in Firefox
2. Open the time setting in the layer control
3. Click the `<select>` and mouse over the options
4. `<select>` closes

I thought I fixed this in https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/492/, apparently that only fixed the issue in Chrome, weird, I usually try changes in more than 1 browser. Anyhow, this is not really my area, perhaps @ahmadayubi can take a look?

ahmadayubi:
Allows users to use `select` elements in the layer controls in Firefox.

On a somewhat related note, I noticed the `select` element isn't actually visible when screen recording Firefox, but is visible in Chrome. 

Closes #572

### DIFF
--- a/src/mapml/layers/ControlLayer.js
+++ b/src/mapml/layers/ControlLayer.js
@@ -107,11 +107,16 @@ export var MapMLLayerControl = L.Control.Layers.extend({
 
     //overrides collapse and conditionally collapses the panel
     collapse: function(e){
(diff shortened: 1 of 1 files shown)


Malvoz (approved):
Works great, thanks!

PR #575, opened 2021-11-14
Templated features updates

prushforth:


prushforth:
These [two tests ](https://github.com/prushforth/Web-Map-Custom-Element/blob/f7f9aadf18f01349465c3b86ec5d0e90adcc0da8/test/e2e/core/keyboardInteraction.test.js#L189-L224)for tooltips return 2 when they are supposed to return 1.  Is that reasonably caused by a the new speedier testing protocol @ahmadayubi ?

ahmadayubi:
Potentially, you could try changing the timeout values to see if it makes a difference. For debugging the new tests you can also adjust:

https://github.com/prushforth/Web-Map-Custom-Element/blob/f7f9aadf18f01349465c3b86ec5d0e90adcc0da8/jest-playwright.config.js#L7-L10

Change it to non-headless mode and increase the sloMo to slow it down for viewing.

test/e2e/layers/mapMLTemplatedFeatures.test.js
```diff
@@ -61,4 +61,21 @@ describe("Playwright mapMLTemplatedFeatures Layer Tests", () => {
...
     });
   });
-});
\ No newline at end of file
+  describe("Simple query by select values without map extent filter tests", () => {
+    test("All features loaded at start", async () => {
+      const features = await page.$$("css= body > map:nth-child(2) > .mapml-web-map >> css= div > .mapml-templatedlayer-container > div > div > svg > g > g");
+      expect(features.length).toEqual(8);
+    });
+    test("User can select/filter features by category", async () => {
```
prushforth, line 69: @ahmadayubi I am having a time getting this test to run.  The first challenge is to expose the select list via a series of layer control actions, then to evaluate the number of features that result from one or two selections.

src/mapml/utils/Util.js
```diff
@@ -122,15 +125,25 @@ export var Util = {
...
+    if (numberOfAxes >= 2) {
+      locInputs = true;
+    }
     let zoomBoundsFormatted = {
       minZoom:+template.zoomBounds.min,
       maxZoom:+template.zoomBounds.max,
       minNativeZoom:nMinZoom,
       maxNativeZoom:nMaxZoom
     };
+    if(!locInputs && template.extentPCRSFallback && template.extentPCRSFallback.bounds) {
```
prushforth, line 137: I test for the existence of the template.extentPCRSFallback and its .bounds property, so why should I have to include the object at all in the test data?

PR #581, opened 2021-11-19
Remove fullscreen control dependency, make mapml-viewer / web-map the fullscreen element

prushforth:


prushforth:
@Malvoz thanks for previously reviewing this.  I changed the :host() selector to .mapml-fullscreen-on because earlier I hadn't realized what this class was for, just blindly copied it, but I don't think it's appropriate to have it showing up outside our shadow host.  I think it might not be necessary at all fwiw, but not really a big deal.  Will merge when you think its good!

src/mapml.css
```diff
@@ -168,13 +168,34 @@
...
+
+:host(:-webkit-full-screen) {
+  width: 100%!important;
+  height: 100%!important;
+}
+
+:host(.leaflet-fullscreen-on) {
+  width: 100%!important;
+  height: 100%!important;
+}
```
Malvoz, line 190:
```suggestion
:host(.mapml-fullscreen-on) {
  width: 100%!important;
  height: 100%!important;
}
```

Malvoz (changes requested):

Malvoz (approved):

PR #583, opened 2021-11-19
Use host selector for debug mode's fullscreen styles

Malvoz:
I forgot these styles existed when reviewing https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/581.

### DIFF
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -385,6 +385,11 @@
   display: inline;
 }
 
+:host(.mapml-fullscreen-on) .mapml-debug-grid {
+  color: #fff;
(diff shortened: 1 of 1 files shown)


src/mapml.css
```diff
@@ -385,6 +385,11 @@
   display: inline;
 }
 
+:host(.mapml-fullscreen-on) .mapml-debug-grid {
```
Malvoz, line 388: This may as well be `:host(:fullscreen) .mapml-debug-grid`, but changing for consistency.

Malvoz: (and https://caniuse.com/mdn-css_selectors_fullscreen)

Issue #153, opened 2020-08-24
Resources should be minified


Malvoz:
Google's [Lighthouse](https://chrome.google.com/webstore/detail/lighthouse/blipmdconlkpinefehnmjammfjpmpbjk?hl=en) reports a potential saving of 148.6 kb in transfer size (not including CSS):

<img width="600" src="https://user-images.githubusercontent.com/26493779/91040090-f229a400-e60d-11ea-858e-a527b42c54a9.png">

We should:

- [ ] Uglify all JS files
- [x] Minify and concatenate CSS files (fixed in https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/606)

Of course, license text should be preserved.

ahmadayubi:
A minified version of mapml is provided in the build branch and in releases, but having a minified version of the other javascript files would be good too.

Malvoz:
> A minified version of mapml is provided in the build branch and in releases

@ahmadayubi Is `mapml.min.js` used anywhere?

Would it be acceptable to simply uglify/minify everything in the `dist` folder without outputting `.min.(js|css)` file names?

ahmadayubi:
It's currently not used anywhere in the tests and experiments as people usually debug on those files so it's important they can read it (without having to pretty print in chrome debugger), you could definitely minify everything but it would add an extra step to debugging. 

It wouldn't be a big deal though, it's mostly a preference thing I believe. @prushforth has developed the minification process through GruntJS, maybe he could shed light on his decision to offer both .min.js and .js

Malvoz:
> you could definitely minify everything but it would add an extra step to debugging.

Unless we generate [source maps](https://developer.mozilla.org/en-US/docs/Tools/Debugger/How_to/Use_a_source_map) using grunt-contrib-uglify's [sourceMap](https://github.com/gruntjs/grunt-contrib-uglify/blob/main/docs/uglify-options.md#sourcemap) option?

ahmadayubi:
That would definitely work (after setting up and testing to make sure it works as expected), I hadn't considered using source maps in the past.

prushforth:
Addressed by https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/705

PR #586, opened 2021-11-21
Minify and concatenate CSS

Malvoz:
Small performance improvement. Fixes:

> - [ ] Minify and concatenate CSS files

of https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/153.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -11,6 +11,17 @@ module.exports = function(grunt) {
 //        dest: 'dist/c.js'
 //      }
 //    },
+    cssmin: {
+      options: {
+        mergeIntoShorthands: false,
+        roundingPrecision: -1
+      },
+      combine: {
+        files: {
+        'dist/mapml.css': ['node_modules/leaflet/dist/leaflet.css', 'src/mapml.css']
+        }
+      }
+    },
     uglify: {
       options: {
        banner: '/*! <%= pkg.name %> <%= grunt.template.today("dd-mm-yyyy") %> */\n'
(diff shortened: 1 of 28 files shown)


Malvoz:
(tests are consistently failing on my machine even on the main branch so I'm just winging it, hoping all goes well 🤞🏼)

ahmadayubi:
> (tests are consistently failing on my machine even on the main branch

If all the tests are failing you may need to run npm install, one of the packages related to testing was updated.

Malvoz:
@ahmadayubi npm install doesn't seem to have an effect on either branch (other than changing package-lock.json on the main branch, which was updated in this PR), the tests still fail most of the time, I'd say about 90% of the time. I tried increasing waitForTimeout in the failing keyboardinteraction and featurelinks tests and saw some improvements but still haven't figured it out.

prushforth:
I still get errors sometimes on main, fwiw.  I will try to fix those; that might improve the chances of this working.

prushforth:
I **hope** that if you pull from upstream main now that the tests will run. I commented out one suite that was causing problems.

prushforth:
So that didn't work.  I'll have to try again locally, although the problem this time appeared in trying to select stuff but timing out, which may indicate some possibly CSS related commit  issue.

Malvoz:
We can close this out, and try again after tests are more stable?

prushforth:
@Malvoz when I reversed the order of combining the css files, I get a pile of error messages that look like what's happening on the gh  action job.  Don't know if that means anything.

Malvoz:
The order of the combined CSS files is important, if it was reversed then leaflet.css would overwrite mapml.css styles. Anyhow, I saw pretty much the same issues on the main branch as I did with this, and even this branch passed the test runs on a few occasions locally so I don't think it's got to do with concatenating the files.

Again I think it's best to revisit this later. Thanks for looking into it though. 😋

Malvoz:
The test runs passed twice locally, thought I was good to go. 😝

PR #591, opened 2021-11-22
Layer control Accessibility Bug Fixes

### LINKED ISSUE #578: Layer context menu is not focused / announced when Shift-F10 is pressed with focus on layer item
Steps to reproduce

Load this page: https://maps4html.org/experiments/styling/alternate-styles/

Press Tab 6 times - focus should be on layer control
Press Enter - layer control open - focus should shift to first layer entry (imo), but does not
Press Tab - focus now on checkbox of first layer
Press Shift-F10 (on windows) - layer context menu appears, focus should be on first menu item but is not
Press Tab - focus should shift to second context menu item, but actually moves to next interactive element in layer control entry (X usually)  Context menu remains visible but inaccessible, except through shortcut keys (which are not announced).

Anshpreet8:
Fixed the following issues:

- shift the focus to first layer entry upon pressing enter to open the layer control
- shift focus to layer context menu when it's opened and shift focus back to the layer entry when it's closed

Closes #578

### DIFF
--- a/src/mapml/handlers/ContextMenu.js
+++ b/src/mapml/handlers/ContextMenu.js
@@ -426,6 +426,8 @@ export var ContextMenu = L.Handler.extend({
     if(e.originalEvent.button === 0 || e.originalEvent.button === -1){
       this._keyboardEvent = true;
       if(this._layerClicked){
+        let activeEl = document.activeElement;
+        this._elementInFocus = activeEl.shadowRoot.activeElement;
         this._layerMenuTabs = 1;
         this._layerMenu.firstChild.focus();
       } else {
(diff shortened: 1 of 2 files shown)


prushforth:
LGTM - nice work

Malvoz:
I noticed 2 things:

- If I <kbd>shift</kbd> + <kbd>tab</kbd> after the context menu has been opened focus always returns to the layer checkbox, it'd be good if focus instead returned to the button that was previously focused before opening the menu, for example the Layer Settings button.
- Pressing <kbd>esc</kbd> closes the context menu and the map is focused instead

Anshpreet8:
@prushforth using the shadowRoot allowed me to get the current element in focus, I have accordingly made the suggested fixes

prushforth (approved):
Thanks, looks good!

Issue #592, opened 2021-11-23
Keyboard panning scrolls document?


Malvoz:
I'm seeing [this](https://github.com/Malvoz/web-maps-wcag-evaluation/blob/master/images/on-input/google-maps-api.gif) in e.g. https://maps4html.org/experiments/painting/ (you may have to resize the window).

PR #595, opened 2021-11-23
Fix keyboard panning scrolling document issue

ben-lu-uw:
I did not properly call stop() when overriding the onKeyDown function (https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/567)
Closes https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/592

### DIFF
--- a/src/mapml/keyboard.js
+++ b/src/mapml/keyboard.js
@@ -45,6 +45,6 @@ L.Map.Keyboard.include({
             return;
         }
 
-        stop(e);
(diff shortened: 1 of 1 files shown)


Malvoz (approved):
Thanks!

prushforth:
@ben-lu-uw could we put a test on this, somehow?  Would be good to automatically detect regressions here, as it seems it might come from diverse code.  If it's too difficult, don't worry about it. Let's discuss.

PR #571, opened 2021-11-12
Focus features on tab

ahmadayubi:
https://user-images.githubusercontent.com/55214462/141527065-f3e91ac0-0805-482f-abaf-05d64b1b1597.mp4

- [ ] Add tests, needs #574 to be merged to allow new tests to follow new format
- [ ] Focus on popup navigation buttons

Closes #482

Issue #535, opened 2021-10-18
Replace `mapml-focus-buttons` with subclass role of composite widget role or proposed `focusgroup` attribute


Malvoz:
Because [map-based popups will be removed](https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/403) there shouldn't be UI components such as `mapml-focus-buttons` that expect popups (especially one that attempts to solve the general issue with overwhelming keyboard users with tab stops, by allowing users to escape the tab sequence of interactive features).

`mapml-focus-buttons` is problematic:
- There have to be features with popups for them to be available to users (i.e. not applicable to [feature links](https://maps4html.org/experiments/linking/features/))
- Users have to open a popup for a feature that they may not be interested in to be able to reach the `mapml-focus-buttons`

Instead there should be a separate navigation mechanism for users to navigate interactive features: <kbd>arrow</kbd> keys. It'll have to be conveyed to AT users that interactive features can be navigated using arrow keys, or else they may think there's only 1 interactive feature on a map because the next <kbd>tab</kbd> takes them to the attribution.

There are (soon) 2 ways to achieve that:

<details>
<summary>One of the non-abstract subclass roles of the <code>composite</code> widget role (e.g. <code>role="grid"</code> & CO)</summary><br>

https://www.w3.org/TR/wai-aria-practices-1.1/#keyboard:

> ## Developing a Keyboard Interface

> Unlike native HTML form elements, browsers do not provide keyboard support for graphical user interface (GUI) components that are made accessible with ARIA; authors have to provide the keyboard support in their code.

https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_generalnav:

> ## Fundamental Keyboard Navigation Conventions

> ARIA roles, states, and properties model accessibility behaviors and features shared among GUI components of popular desktop GUIs, including Microsoft Windows, macOS, and GNOME. Similarly, ARIA design patterns borrow user expectations and keyboard conventions from those platforms, consistently incorporating common conventions with the aim of facilitating easy learning and efficient operation of keyboard interfaces across the web.

> A primary keyboard navigation convention common across all platforms is that the tab and shift+tab keys move focus from one UI component to another while other keys, primarily the arrow keys, move focus inside of components that include multiple focusable elements.

[...]

PR #596, opened 2021-11-24
Tab order based on distance from center

### LINKED ISSUE #385: Assign tabindex=n in ascending order of distance of feature from map centre
It seems that having to tab through every feature in a map viewport could be a burden especially for blind users.  One of the benefits of maps is that sighted users are able to immediately see the relevance of spatial phenomena relative to locations, often the map centre point.  

There should be a way to improve the usability of maps with features for blind users, and I think this might help with one aspect of that, by reducing the amount of tabbing they have to do to get relevant info.  

When a server is involved, it could sort features naturally in that order before serializing the document, without having to set the tab order.  But it should also be a function that can be requested on the client, perhaps with a user preference.  Sorting these things can be an intensive process so it might not be something the viewer should do on every request, but it seems reasonable to allow a user to make it *their* default.

Unsure if there's a UCR issue for this, not at first glance anyway.

ahmadayubi:
https://user-images.githubusercontent.com/55214462/143302447-6f728be5-4dfd-4e09-bbeb-7368421bec97.mp4

- [x] Add tests / adjust tests for new behavior
- [x] Update popup navigation button functionality
- [x] Prevent features off screen from being tab focused

Closes #385

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@maps4html/web-map-custom-element",
-  "version": "0.8.3",
+  "version": "0.9.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/src/mapml-viewer.js
+++ b/src/mapml-viewer.js
@@ -152,6 +152,40 @@ export class MapViewer extends HTMLElement {
       });
     });
     this.controlsListObserver.observe(this, {attributes:true});
+    this._featureIndexOrder = [];
+    this._currFeatureIndex = 0;
+  }
+
+  _addToIndex(layer, path) {
+    let scale = this._map.options.crs.scale(this._map.getZoom());
+    let mc = this._map.options.crs.transformation.untransform(this._map.getPixelBounds().getCenter(),scale);
+    let lc = layer.getPCRSCenter();
+    let dist = Math.sqrt(Math.pow(lc.x - mc.x, 2) + Math.pow(lc.y - mc.y, 2));
+    let index = this._map.options.mapEl._featureIndexOrder;
+
+    let elem = {path: path, layer: layer, dist: dist};
+    path.setAttribute("tabindex", -1);
+    index.push(elem);
+    for (let i = index.length - 1; i > 0 && index[i].dist < index[i-1].dist; i--) {
+      let tmp = index[i];
+      index[i] = index[i-1];
+      index[i-1] = tmp;
+    }
+  }
+
+  _sortIndex() {
+    let scale = this._map.options.crs.scale(this._map.getZoom());
+    let mc = this._map.options.crs.transformation.untransform(this._map.getPixelBounds().getCenter(),scale);
+    let index = this._map.options.mapEl._featureIndexOrder;
+    index[0].path.setAttribute("tabindex", -1);
+    index.sort(function(a, b) {
+      let ac = a.layer.getPCRSCenter();
+      let bc = b.layer.getPCRSCenter();
+      a.dist = Math.sqrt(Math.pow(ac.x - mc.x, 2) + Math.pow(ac.y - mc.y, 2));
+      b.dist = Math.sqrt(Math.pow(bc.x - mc.x, 2) + Math.pow(bc.y - mc.y, 2));
+      return a.dist - b.dist;
+    });
+    index[0].path.setAttribute("tabindex", 0);
   }
   connectedCallback() {
     if (this.isConnected) {
@@ -375,6 +409,7 @@ export class MapViewer extends HTMLElement {
               {target: this}}));
       }
     });
+    this._map.on('mapfocused', () => this._sortIndex());
     this._map.on('load',
       function () {
         this.dispatchEvent(new CustomEvent('load', {detail: {target: this}}));
--- a/src/mapml-viewer.js
+++ b/src/mapml-viewer.js
@@ -152,49 +152,6 @@ export class MapViewer extends HTMLElement {
       });
     });
     this.controlsListObserver.observe(this, {attributes:true});
-    this._featureIndexOrder = [];
-    this._currFeatureIndex = 0;
-  }
-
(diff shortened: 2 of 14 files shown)


ahmadayubi:
@prushforth 
Assigning a tab index to features while the map has a `taxindex = 0` means the feature gets focused before the map. Is this the intended/accepted behavior?

prushforth:
> Is this the intended/accepted behavior?

My first impression is that the map should always come first.  It would be acceptable to skip the controls and go straight to the features, maybe.

ahmadayubi:
> > Is this the intended/accepted behavior?
> 
> My first impression is that the map should always come first. It would be acceptable to skip the controls and go straight to the features, maybe.

To get this behavior you either have to;
1. set the `tabindex` of the map to 1
2. remove the `tabindex` values when the map is not focused and re-add `tabindex` only when the map is focused

Malvoz:
I think 2 is the only option, we don't want to break the tab order of elements on people's websites.

Malvoz:
See also:

- https://adrianroselli.com/2014/11/dont-use-tabindex-greater-than-0.html
- https://dequeuniversity.com/rules/axe/4.0/tabindex
- etc.

Malvoz:
Would this be compatible with https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/571?

ahmadayubi:
> Would this be compatible with #571?

No I believe that's getting shelved for now

prushforth:
Could we establish a private focus order index value that we follow, instead of setting the tabindex > 0?  We could handle keystrokes while in a sequence of features and set focus to the next highest/ next lowest private focus order index, perhaps

prushforth:
I was looking for where @malvoz mentioned using arrow keys instead of tab keys within a group of features.  If we implemented that, we wouldn't be hacking the tabindex, I think.

Malvoz:
> I was looking for where @Malvoz mentioned using arrow keys instead of tab keys within a group of features.

That's https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/535.

prushforth:
So, I think we should be using tabindex=-1 and setting a private focus index value that we programmatically use to set the focus on a feature. The private value can be ordered ascending by distance??   Maybe the first feature in the DOM should be tabindex=0 and the rest -1?  WDYT?

ahmadayubi:
That would work, does handling the tab event ourselves have any benefit over adding tabindex when the map is focused and removing it once the map is out of focus?

prushforth:
Good question, but if we don't change the tabimdex in an antisocial way, it won't matter if our program crashes and can't change it back? Maybe

prushforth:
> Maybe the first feature in the DOM should be tabindex=0 and the rest -1? WDYT?

Maybe the first feature in each layer should be tabindex=0, with the rest =-1...

ahmadayubi:
https://user-images.githubusercontent.com/55214462/143720161-a8aa56af-b3cf-4455-ae87-88be98182af2.mp4

So I opted to use event handlers as this would also further improvements in the future brought up by @prushforth regarding arrow keys.

All the features in the map are considered rather than on a per layer basis. The highest `tabindex` value used is now 0, never exceeding this.

The final thing remaining is treating subparts (spans within coordinates) as independent when sorting, currently it treats subparts distance from the center as equal to the distance of the main component.

ahmadayubi:
The functionality should be reviewed before the tests are updated to reflect the new behavior. So if anyone has time please share any feedback or bugs.

prushforth:
Looking pretty nice!  A couple of things I noticed.

1) for https://maps4html.org/experiments/screenreader/restaurants/ page, if you build your branch and put the dist into a local clone of experiments, it sorts the restaurants very nicely, but if you zoom, you then have to tab twice to get past a marker.

2) in this page with the sorting build in dist https://maps4html.org/experiments/screenreader/closest-restaurants/ if you click the Restaurants button, you'll get the following error:

image

I'll continue to look for stuff....

ahmadayubi:
@prushforth The issues you mentioned should be fixed with the latest update, good finds!

prushforth:
@ahmadayubi the issues seem mostly fixed.  I was able to reproduce the first problem:

> 
> 1. for https://maps4html.org/experiments/screenreader/restaurants/ page, if you build your branch and put the dist into a local clone of experiments, it sorts the restaurants very nicely, but if you zoom, you then have to tab twice to get past a marker.  If you zoom twice, I think you have to tab 3 times to get past a marker, and the number of times you have to tab grows as you change zoom.

but not as bad, by first loading the restaurants, then tabbing through them, then zooming in and tabbing a few times, then right-click the restaurants layer entry and press "Zoom to layer".  The multiple-tabs-to-get-past-a-marker problem occurs.

ahmadayubi:
@prushforth Turns out the reason it needed multiple tab presses is because the features are duplicated in the DOM. The bug is related to `zoomToLayer` in combination with `TemplatedFeaturesLayer`, where the `_pullFeatureFeed` is called twice (through the moveend handler) instead of once, not sure why yet.

prushforth:
I have a feeling that might be due to a change I made during template features processing. So it's an unrelated bug to this.

ahmadayubi:
I've been meaning to fix the zoomToLayer anyways, it has a bad flicker due to how it functions but, I got a fix for it now.

Malvoz (approved):
LGTM!

This change on its own would make it difficult for screen reader users to discover that new features have become available after pan/zoom, but it's an important piece of the puzzle. And it all comes together with https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/636.

prushforth (approved):
Looks good to me.  When tabbing through many features, we do need a way to skip  to the end of the sequence so as to not trap the user.  This may come into play with feature indexing too.  I look forward to implementing arrow key feature navigation as a way of leaving tabs to inter-control/interactive element navigation.

prushforth:
Boom!

PR #598, opened 2021-11-24
Add role and aria-label to attribution control. 

prushforth:
Addresses one task in #247


Malvoz (approved):
Nice, tested in NVDA and it's definitely an improvement. Thanks!

PR #599, opened 2021-11-25
Add scroll test

ben-lu-uw:
Adds a test for issue https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/592 / https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/595


ben-lu-uw:
The keyboardinteraction test is currently commented out, so I just added this test as its own thing for now.

prushforth:
Thanks!

PR #600, opened 2021-11-26
Focus on Layer Menu Item or Link if Remove Button is clicked

### LINKED ISSUE #589: Nothing is focused when a layer is removed using the layer control
Tab to the remove layer button on the layer control and press enter.

Expected: the next element should be focused
Actual: Nothing is focused

Anshpreet8:
Once a layer is removed using the layer control, the focus shifts to the next layer menu item, if it exists, or the attribution link

Closes #589

### DIFF
--- a/src/mapml/layers/MapLayer.js
+++ b/src/mapml/layers/MapLayer.js
@@ -386,9 +386,6 @@ export var MapMLLayer = L.Layer.extend({
     getAttribution: function () {
         return this.options.attribution;
     },
-    getShadowRoot: function(root){
-
-    },
     getLayerUserControlsHTML: function () {
       var fieldset = L.DomUtil.create('fieldset', 'mapml-layer-item'),
         input = L.DomUtil.create('input'),
--- a/src/mapml/layers/MapLayer.js
+++ b/src/mapml/layers/MapLayer.js
@@ -386,6 +386,9 @@ export var MapMLLayer = L.Layer.extend({
     getAttribution: function () {
         return this.options.attribution;
     },
+    getShadowRoot: function(root){
+
+    },
     getLayerUserControlsHTML: function () {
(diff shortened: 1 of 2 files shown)


prushforth:
Works very well, tests too.  Good work!

prushforth (approved):
👍

src/mapml/layers/MapLayer.js
```diff
@@ -425,7 +425,21 @@ export var MapMLLayer = L.Layer.extend({
...
+              fieldset += 2; // find the next layer menu item
+              elem = elem.previousElementSibling;
+            }
+          } else {
+            // focus on the link
+            elem = "link";
+          }
           mapEl.removeChild(e.target.closest("fieldset").querySelector("span").layer._layerEl);
+          elem = elem ? root.querySelector(".leaflet-control-attribution").firstElementChild: elem = root.querySelectorAll('input')[fieldset];
+          setTimeout(() => elem.focus(), 800); // a timeout is set so "pressed remove layer" is announced first
```
Malvoz, line 442: This timeout seems to cause NVDA (using Chrome) to announce the document title + map then next layer/link, I'm not hearing "pressed remove layer". In that sense I think doing without the timeout provides a better experience?

Anshpreet8: Using a Mac and voiceover, I hear "pressed remove layer", but if that's the case with NVDA, I could remove the timeout for better consistency

Malvoz: > if that's the case with NVDA, I could remove the timeout for better consistency

There doesn't seem to be a whole lot of guidance on the desired behavior when removing/hiding the currently focused element. But I think removing the timeout provides a similar experience to when e.g. dialog elements are hidden as a result of closing it: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#result

Issue #388, opened 2021-04-11
Allow Ctrl-Arrow decrease the pan increment


prushforth:
Make it easier to target smaller features using keyboard + graticule

Malvoz:
2 observations:
- I've been looking into gestures and keyboard controls recently, and it seems none of the UCR reference tools use the <kbd>ctrl</kbd> modifier key for anything, for some reason. So if pursuing this it might be worth investigating if that may be an issue.
- AFAICT with [Feature Indexing](https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/397) this would potentially not be necessary.

ben-lu-uw:
In the _onKeyDown function in Leaflet they return if the control key is pressed, not really sure why though. 
![image](https://user-images.githubusercontent.com/69722289/143497306-7f79a487-dd3c-4f22-97c2-e83ec096b3f8.png)
Maybe cause Chrome has a [ctrl + arrow shortcut](https://support.google.com/chrome/answer/157179?hl=en&co=GENIE.Platform%3DDesktop#zippy=%2Ctab-and-window-shortcuts) set for moving tabs but it does not seem to work or do anything for me? 

I did try adding this feature here https://github.com/ben-lu-uw/Web-Map-Custom-Element/tree/ctrl_pan and it seems to be issue free for me

Malvoz:
Google Maps uses the <kbd>shift</kbd> modifier to decrease the pan increment, however Leaflet already uses that to increase it...

prushforth:
I tested powerpoint just now and ctrl-arrow doesn't appear to do anything different than just arrow by itself.
Shift-arrow increases the size of an object, and ctrl-shift-arrow also increases the size but reduces the size increment step.

Ctrl-scrollwheel or Ctrl-+/- changes the magnification of a domain in chrome.

Ctrl-arrow skips a word in most text editors, so increases the increment of cursor movement.

Let's dig a little more to see if there's anything arguing against using ctrl-arrow for decreased pan increments.

Malvoz:
[Ability to rotate map](https://github.com/Maps4HTML/Web-Map-Custom-Element/projects/2#card-43799698) will need a modifier too.

ben-lu-uw:
ctrl + alt + arrow keys is what I am used to for changing orientation usually.

ben-lu-uw:
[...]

PR #601, opened 2021-11-26
Ctrl arrow decreases pan offset

ben-lu-uw:
Closes https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/388
Leaflet does not want modifier + arrow keys being pressed -> https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/388#issuecomment-980096154 so leaving this as a draft for further discussion.

### DIFF
--- a/src/mapml/keyboard.js
+++ b/src/mapml/keyboard.js
@@ -28,7 +28,7 @@ L.Map.Keyboard.include({
                     offset = L.point(offset).multiplyBy(3);
                 }
                 if (e.ctrlKey) {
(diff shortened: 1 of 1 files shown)


prushforth:
I tried a L.point(offset).divideBy(**5**), I think that gives you a little finer control when [using the reticle](https://maps4html.org/experiments/datacube/)

Malvoz (approved):
My only concern is https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/388#issuecomment-979500927, but that's not relevant at this time.

PR #605, opened 2021-11-26
Focus on Layer Menu Item or Link if Remove Button is clicked

Anshpreet8:
Removed the timeout before focusing on the next element, as [suggested](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/600#discussion_r757655717)


src/mapml/layers/MapLayer.js
```diff
@@ -439,7 +439,7 @@ export var MapMLLayer = L.Layer.extend({
           }
           mapEl.removeChild(e.target.closest("fieldset").querySelector("span").layer._layerEl);
           elem = elem ? root.querySelector(".leaflet-control-attribution").firstElementChild: elem = root.querySelectorAll('input')[fieldset];
-          setTimeout(() => elem.focus(), 800); // a timeout is set so "pressed remove layer" is announced first
+          elem.focus(); // a timeout is set so "pressed remove layer" is announced first
```
Malvoz, line 442:
```suggestion
          elem.focus();
```

Malvoz (approved):
Thanks!

Issue #580, opened 2021-11-18
Feature links should delegate focus


Malvoz:
Repro steps for https://maps4html.org/experiments/linking/features/:

1. <kbd>Tab</kbd> three times to focus the Prince Edward Island feature
2. Press <kbd>Enter</kbd> to visit the feature link

On step 2:
 - "Interactive map, region" is announced 
    Expectation: "Prince Edward Island" is announced
 - The Prince Edward Island feature is not focused (tabbing starts from the map component -> Newfoundland and Labrador -> Prince Edward Island)
    Expectation: Prince Edward Island is focused

Can focus be sent to features after activating feature links?

prushforth:
> Can focus be sent to features after activating feature links?

I think so.   We'll try to fix this.

ben-lu-uw:
When you first tab to PEI it will announce "Prince Edward Island", and in step 2 you want it to announce it again? Isn't that a bit redundant. @Malvoz

Malvoz:
> in step 2 you want it to announce it again?

@ben-lu-uw I guess that's just the effect focus has, and is not unlike e.g:
```html
<!-- links in a table of contents -->
<a href="#cats">Cats</a> <!-- on focus: "Cats" announced -->

<!-- page content -->
<h2 id="cats">Cats</h2> <!-- on focus as a result of following the above link: "Cats" announced -->
```

Also, I'm not exactly sure how feature links work but I guess a feature can be named one thing but link to a feature with another name, in which case it makes even more sense that the feature name is announced as a result of focus.

PR #607, opened 2021-11-26
Focus on the feature when activating feature link

ben-lu-uw:
Closes https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/580


Malvoz:
Thanks @ben-lu-uw.

It seems I misunderstood how feature links work: https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/580#issuecomment-980448862, for some reason I thought they'd be rerendered and refocused which would cause the feature to be announced again once visited. Instead nothing is announced when a "same-map" link is visited.

Maybe @prushforth has some thoughts on this? Either way, this does improve the experience using a keyboard, maybe we can discuss the screen reader experience separately?

Malvoz (approved):

Malvoz:
> maybe we can discuss the screen reader experience separately?

Filed https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/615.

Issue #609, opened 2021-11-29
Repo sync does not contain the minified CSS


Malvoz:
After https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/606 I would've expected https://github.com/Maps4HTML/experiments/blob/main/dist/mapml.css to hold the contents of `node_modules/leaflet/dist/leaflet.css` but it does not, the result is broken maps in all synced repos.

Could it be as simple as updating:

https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/90540cbb67ba9ca3511087f62274a74581141831/.github/workflows/sync.yml#L21

per https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/606#issuecomment-980636641:

> CI action was running 'grunt clean copy rollup', not 'grunt default' so the minification task wasn't being run

?

Malvoz:
Side note: `leaflet.css` should be removed on the following line since we're only using `mapml.css`:

https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/90540cbb67ba9ca3511087f62274a74581141831/Gruntfile.js#L62

Malvoz:
Could it be as simple as updating:

https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/90540cbb67ba9ca3511087f62274a74581141831/.github/workflows/sync.yml#L21

per https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/606#issuecomment-980636641?

ahmadayubi:
It doesn't delete the files as I assumed the files in dist wouldn't be removed (changes and additions are considered), but that can be changed of course.

As for why some files date from long ago, that's because even if you replace a file with a new file, if the name and contents are identical git doesn't consider that a change, so the files that date from long ago are files that haven't been changed.

prushforth:
I think we should delete everything in dist and then copy everything from the source dist to the destination dist so that files that are deleted / no longer exist in the distribution are actually removed.

prushforth:
I could make that change, probably lol, if you are busy.

ahmadayubi:
> I could make that change, probably lol, if you are busy.

I actually was just submitting a PR for it haha. I tested the `rm -rf .../*` command locally, it should work in the github actions environment as well.

PR #610, opened 2021-11-29
Run `grunt default` in repo sync

Malvoz:
Fix https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/609.

I don't know that this is correct and that it fixes the issue. Does it look correct?


prushforth:
Thanks!

PR #614, opened 2021-11-29
Update package version

ben-lu-uw:


ben-lu-uw:
Not sure if merging this PR will also merge the v0.9.0 tag I added to ben-lu-uw:main

prushforth:
Let's see if it does

prushforth:
I don't see it

ben-lu-uw:
Added manually

PR #620, opened 2021-12-01
Support the `hidden` attribute on the map elements

Malvoz:
Quite an oversight. To support the global [`hidden`](https://html.spec.whatwg.org/multipage/interaction.html#the-hidden-attribute) attribute we have to handle the `display` ourselves, otherwise it is overridden by the [`display: inline-block` declaration](https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/7a0b4dc99e31c65f978816a8b60ff80b9c0869fe/src/mapml-viewer.js#L118) that is set for the default display.

I chose to use an `!important` declaration because if the authors changes the default from `display: inline-block` (which is used in a `:host` selector that has lower specificity than element selectors) to `mapml-viewer { display: block }` then uses `hidden` it'll still be `display: block`. And because: https://github.com/jensimmons/cssremedy/issues/71#issuecomment-771180150.

### DIFF
--- a/src/mapml-viewer.js
+++ b/src/mapml-viewer.js
@@ -124,6 +124,9 @@ export class MapViewer extends HTMLElement {
     `:host([frameborder="0"]) {` +
     `border-width: 0;` +
     `}` +
+    `:host([hidden]) {` +
+    `display: none!important;` +
+    `}` +
     `:host .mapml-contextmenu,` +
     `:host .leaflet-control-container {` +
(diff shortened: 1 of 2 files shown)


Malvoz:
(Thanks to the authors of `<i-slide>` for challenging my assumption that the presentational attribute would override the low specificity of the `:host` selector: https://github.com/w3c/i-slide/blob/f6747ab6e15e8de5347ba20ab0a8757e3a29bfdd/ISlide.js#L711. I should've tested the attribute a long time ago! 😅)

PR #621, opened 2021-12-03
Replace `style` attributes with presentation attributes in the context menu and crosshair

Malvoz:
Unlike `style=""`, the `hidden` attribute is not governed by CSP `style-src`. This change:
- removes the need for 3 `style-src` hashes in the currently ["required" CSP](https://maps4html.org/experiments/security/csp/)
- optimizes the byte size of the crosshair SVG (part of the process of moving from `style` attributes to SVG presentation attributes)
- removes the need to apply the "`visibility` hack" to the context menu
- removes some redundant CSS

### DIFF
--- a/src/mapml-viewer.js
+++ b/src/mapml-viewer.js
@@ -127,7 +127,6 @@ export class MapViewer extends HTMLElement {
     `:host([hidden]) {` +
     `display: none!important;` +
     `}` +
-    `:host .mapml-contextmenu,` +
     `:host .leaflet-control-container {` +
     `visibility: hidden!important;` + // Visibility hack to improve percieved performance (mitigate FOUC) – visibility is unset in mapml.css! (https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/154).
     `}`;
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -259,7 +259,6 @@
  */
 
 .mapml-contextmenu {
-  display: none;
   border-radius: 4px;
   padding: 4px 0;
   background-color: #fff;
@@ -270,7 +269,6 @@
 }
 
 .mapml-contextmenu button.mapml-contextmenu-item {
-  display: block;
   color: #222;
   font-size: 12px;
   line-height: 20px;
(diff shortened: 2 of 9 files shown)


ahmadayubi (approved):
Looks good! The related changes to CSS optimization sometimes go over my head as I'm not sure what alternative is the better but everything seems to work.

I wish there was a good way of moving the SVG content out of the JS files entirely.

Malvoz:
@ahmadayubi Moving from inline `style=""` to presentational attributes (or into a stylesheet) is just to minimize the amount of CSP hash sources needed (not needed if using nonces, but most developers who deploy a CSP don't use that).

Here's a minimal test case:

```html
<!doctype html>
<meta http-equiv="content-security-policy" content="default-src 'self';">
<div style="display: none"></div>
```

If you open that HTML in your browser, you'll find a console error message (at least in Chrome):

> Refused to apply inline style because it violates the following Content Security Policy directive: [...]

i.e. would require the following from the author:

```html
<!doctype html>
<meta http-equiv="content-security-policy" content="default-src 'self';
                                                    style-src 'self'
                                                              'unsafe-hashes' /* or just 'unsafe-inline' without hashes, though that's considered less safe */
                                                              'sha256-ZdHxw9eWtnxUb3mk6tBS+gIiVUPE3pGM470keHPDFlE=';">
<div style="display: none"></div>
```

whereas the following is not blocked on the CSP:

```html
<!doctype html>
<meta http-equiv="content-security-policy" content="default-src 'self';">
<div hidden></div>
```

So in general, if we can try to avoid inline `style=""` it'll help authors who want to deploy a non-nonce based CSP use the map component without (potentially) setting a bunch of hash sources.

Malvoz:
> I wish there was a good way of moving the SVG content out of the JS files entirely.

There is the option to convert the SVGs to data URIs and set them as background-images in mapml.css, but that results in a greater issue which is that colors aren't inherited from the document or shadow host (so does not work well with customization, dark mode, and high contrast mode). 🤔

Malvoz:
I'll go ahead and merge this, thanks!

PR #622, opened 2021-12-03
Move feature interaction styles to `mapml.css`

Malvoz:
ATM these styles are duplicated into each vector container. This change makes the feature styles more bandwidth friendly, and not add to the total DOM size.

### DIFF
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -735,3 +735,37 @@ label.mapml-layer-item-toggle {
   white-space: nowrap;
   width: 1px;
 }
+
+/*
+ * Feature styles.
+ */
+ 
+.mapml-vector-container svg :is(
+  [role="link"]:focus,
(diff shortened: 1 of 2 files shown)


prushforth:
I think the entire `_removeCSS` might be obsolete with this change.  I'll let @ahmadayubi decide.  In general, LGTM, I'm in favour of moving styles into CSS modules if possible.

ahmadayubi:
I believe `_removeCSS` is still needed as the container still contains the user created styles for the layer. Those user created styles still get added to the container on moveend so the old styles need to be removed.

prushforth:
> I believe `_removeCSS` is still needed

So, good to merge then?

ahmadayubi (approved):
Yeah everything else seems to work as expected for me,

PR #630, opened 2021-12-07
Remove applied `opacity` on disabled controls

### LINKED ISSUE #629: `opacity` issues on disabled controls
UAs use `opacity` on disabled controls, but it doesn't work well on the map controls for the following reasons:
1. Unlike UAs we use `aria-disabled` and not `disabled`, so the controls remain focusable, causing a barely visible `outline` on focus (because the outline too has the opacity applied).

2. Too low contrast on the Reload button (because it has both a lighter color + opacity as of https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/531).

3. Creates a [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context) (I believe) on disabled controls in the layer control, causing underlying layer setting buttons to show through the overlaying layer item on drag.

Malvoz:
Fix #629.

Side note: Opacity has the benefit that if authors were to customize and change colors of controls (no styling hooks currently provided), there'd still be an indicator that a control is disabled without them having to set a particular color for disabled controls, so may want to revisit opacity for controls in the future (e.g. after https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/19?).


Malvoz:
Updated the PR description with visuals.

prushforth (approved):
LGTM! Thank you.

PR #632, opened 2021-12-13
Allow Multiple extents

### LINKED ISSUE #439: Would it be useful to allow and enable more than one `<extent units="...">` per layer
One advantage _might_ be that you could put several different projections in a single layer / mapml file and the client could pick the one that suits the map at the time.  

Another possibility is that it could provide a scope for link templates and template variable names, which might help decrease error rates when editing them by hand / copying pasting+modifying (a standard coding method!).

Idea / question for discussion, pros and cons and so on.

Anshpreet8:
Each Extent added is presented in the layer control, under the corresponding layer. If the label attribute is missing from the extent, it's inherited from the layer (for now).

Note, there are some issues with opacity and query links when unchecking then checking an extent that I'm currently working on.

Other things I have not yet considered or taken a look at the behaviour of are:

- [ ] Zoom Bounds
- [x] Right Clicking on the extent in layer control results in an error
- [x] Reordering extents
- [ ] The behaviour when there is a feature query link and another query link (https://cwfis.cfs.nrcan.gc.ca/geoserver/public/wms?i={i}&j={j}&service=WMS&version=1.3.0&request=GetFeatureInfo&layers=public:fdr_current&QUERY_LAYERS=current:fdr_current&styles=&bbox={xmin},{ymin},{xmax},{ymax}&width={w}&height={h}&srs=EPSG:3978&INFO_FORMAT=text/html&m4h=t) within the same bounds

There might be other things or issues, feel free to list them.

Closes #439


Malvoz (commented):
Ideally each `mapml-layer-extent` would be exposed the same to ATs as the layers are, i.e.:
1. Change `div` to `fieldset`
2. Have the `fieldset` labelled (`aria-labelledby`) by the corresponding `mapml-layer-item-name`

Also, I think it'd be desired to have all `mapml-layer-extent` elements themselves be children of a group (`<fieldset>`) which is labelled (using `<legend>` or `aria-label`, probably the latter to keep it consistent) "Sublayers" (or "Extents"?).

Anshpreet8:
> 1. Change `div` to `fieldset`
Would it be okay to have nested fieldsets?

Malvoz:
@Anshpreet8 they state:

> Screen readers do not automatically indicate the end of the `<fieldset>` element

I don't know if that's the expectation, but even if that's true I don't see how that's a problem.

> so it is impossible for screen reader users to confidently know which fields belong within which fieldset.

That's an incorrect statement, if a fieldset is labelled - as suggested - the user will know which inputs belong to that fieldset.

I say go ahead 👍🏼

ahmadayubi:
The issue with zoom to layer is that the `<layer->` custom element's zoomTo function (the `focus()`) doesn't have an idea of multiple extents currently. It is a method of the `<layer->` custom element so it's simply using the `<layer->` context. 

The copy extent menu item also uses the parent `<layer->`'s context, not the various extent's context.

The `<layer->` API functions (`focus`) would need updating to consider the various extents seperately, or the functionality can be moved out of the `<layer->` API and become a event handler within mapml.js

The functions referenced are:

https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/816f7c19372bae1da571f74656b38a555d08bbab/src/layer.js#L317

https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/816f7c19372bae1da571f74656b38a555d08bbab/src/mapml/handlers/ContextMenu.js#L182

prushforth:
The issue with Zoom to layer was, as @Anshpreet8 suggested, an issue with extractInputBounds on the main branch, which has been fixed and merged into this branch.

src/mapml/layers/MapLayer.js
```diff
@@ -386,6 +497,172 @@ export var MapMLLayer = L.Layer.extend({
...
+                control.removeAttribute("aria-dropeffect");
+                control.style.pointerEvents = null;
+                control.style.transform = null;
+                let controlsElems = controls.children,
+                    zIndex = 0;
+                for(let c of controlsElems){
+                  let extentEl = c.querySelector("span").extent;
+                  
+                  extentEl.setAttribute("data-moving","");
+                  mapEl.insertAdjacentElement("beforeend", extentEl);
```
prushforth, line 649: When you drag an extent to re-order it, the \<map-extent> elements (all) get re-parented into the mapml-viewer element.

src/mapml/layers/TemplatedTileLayer.js
```diff
@@ -33,6 +33,9 @@ export var TemplatedTileLayer = L.TileLayer.extend({
       L.TileLayer.prototype.onAdd.call(this,this._map);
       this._handleMoveEnd();
     },
+    addTo: function(map){
```
prushforth, line 36: This function is unnecessary, because it overrides an ancestor function of the same name, and is probably not doing what it's supposed to: it is creating a whole new Leaflet TileLayer in place of the current TemplatedTileLayer, using the factory L.tileLayer method.  Although it's unnecessary, if we wanted to call the ancestor addTo function ourselves, as we do in the initialize constructor, we could use the L.TileLayer.prototype.addTo.call(this, map) technique described [here](https://leafletjs.com/examples/extending/extending-1-classes.html#methods-of-the-parent-class).

src/mapml/layers/MapLayer.js
```diff
@@ -72,8 +82,88 @@ export var MapMLLayer = L.Layer.extend({
...
+        } else {
+            L.DomEvent.stopPropagation(e);
+            extentEl.checked = false;
+            if(this._layerEl.checked) this._map.removeLayer(extentEl.templatedLayer);
+            this._getCombinedExtentsLayerBounds();
+        }
+    },
+
+    // if there are multiple extents, find the combined layerbounds
+    _getCombinedExtentsLayerBounds: function(){
```
prushforth, line 113: There should be only one way of setting `<layer->.extent`, regardless of how many `<map-extent>` elements there are.  This method would seem to be that way, need to factor out the other way.

src/mapml/layers/MapLayer.js
```diff
@@ -196,35 +301,29 @@ export var MapMLLayer = L.Layer.extend({
...
+                break;
+              } 
+            }
       }
+    }
       return !(noLayer || this.getProjection() !== map.options.projection.toUpperCase());
     },
 
     //sets the <layer-> elements .bounds property 
     _setLayerElExtent: function(){
```
prushforth, line 322: Probably need to get rid of this method entirely, keep (a possibly renamed)_getCombinedExtentsLayerBounds (above, L113).

src/mapml/layers/MapLayer.js
```diff
@@ -63,6 +63,16 @@ export var MapMLLayer = L.Layer.extend({
             this._container.style.zIndex = this.options.zIndex;
         }
     },
+    // remove all the extents before removing the layer from the map
+    _removeExtents: function(map){
```
prushforth, line 67: Looks like the MapMLLayer._extents._queries property should be purged here, too.

prushforth:
Hi @Anshpreet8 

This PR is very solid work, thank you!  Sorry to have taken so long to properly review, most changes are minor and to help the reader (me) understand the intent.

> Other things I have not yet considered or taken a look at the behaviour of are:

> [ ] Zoom Bounds
 > [x] Right Clicking on the extent in layer control results in an error
 > [x] Reordering extents
 > [ ] The behaviour when there is a feature query link and another query link (https://cwfis.cfs.nrcan.gc.ca/geoserver/public/wms?i={i}&j={j}&service=WMS&version=1.3.0&request=GetFeatureInfo&layers=public:fdr_current&QUERY_LAYERS=current:fdr_current&styles=&bbox={xmin},{ymin},{xmax},{ymax}&width={w}&height={h}&srs=EPSG:3978&INFO_FORMAT=text/html&m4h=t) within the same bounds

Can you clarify for me what's left to do for the unchecked items above?  I will try to address them if they still need attention.

The only thing that stands out to me was the need to cycle the debug mode off / on to reflect extent removal/addition.  I will take a look to see if that can be easily fixed.  I may just merge this PR and log an issue for later so that we can move on to other work.

Anshpreet8:
> Can you clarify for me what's left to do for the unchecked items above? I will try to address them if they still need attention.

Hi @prushforth , when there were was a feature query extent and a non-feature query extent, the behaviour was kind of odd, the non-feature link only pops up once and when I tried going forward and back to the previous queries, it was no longer there.

prushforth:
List of things to do related to multiple extents (feel free to add/suggest, remove items):

- [ ] prevent absent extent `label` attribute from making previous sibling extent elements hidden to the user
- [ ] enable `checked` attribute on `extent` so that extents can be author configured, slightly
- [ ] allow author to set extent opacity
- [ ] enable extent display/undisplay in debug mode when checked/unchecked
- [ ] label extents according to their `label` value in debug mode
- [ ] create / enable "Zoom To..." in layer control (extent) context menu
- [ ] create / enable "Copy Extent As..." in layer control (extent) context menu
- [ ] Enable mixed HTML/MapML query responses to be navigated via popup
- [ ] ...

Issue #612, opened 2021-11-29
Feature links layer control bug?


Malvoz:
https://user-images.githubusercontent.com/26493779/143908562-b782e746-52f7-4957-94da-4990d5d6a1c7.mp4

ahmadayubi:
Is the bug the flickering or the fact that the layer gets added multiple times when pressed multiple times?

Malvoz:
> More of a feature than a big, imho

What's the usefulness of re-adding layers in such a manner? I can only see how it can be confusing to users.

> Is the bug the flickering or the fact that the layer gets added multiple times when pressed multiple times?

@ahmadayubi Sorry I should've been more clear, I was in a rush. It is the latter.

Is the flickering perhaps related to:

https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/90540cbb67ba9ca3511087f62274a74581141831/src/mapml-viewer.js#L204-L209

?

ahmadayubi:
> Is the flickering perhaps related to:

I think the flickering is related to how layers are fit to the extent of the map. Since leaflets fitToBounds function only considers position and not zoom level a custom solution had to be used. This custom solution tries zoom levels until it finds the best one, although this works fine when it transitions to a different zoom level, it causes flickering when it's best fit is the current zoom. This could be improved by rather calculating fit (some what complicated) instead of trying it.

prushforth:
>>More of a feature than a big, imho

> What's the usefulness of re-adding layers in such a manner? I can only see how it can be confusing to users.

That file is a test file; the  `target="_blank"` means that the *author* has set the page up that way.  The default value of `target` is `_self`, which replaces the current layer.    Maybe you think that `_blank` for maps is not useful as implemented?  When the `type` attribute is `text/html`, `_blank` opens a new tab.  [When the mime type is `text/mapml`, `_blank` adds a new layer.](https://maps4html.org/web-map-doc/docs/other-elements/map-a/#target)   

You (the author) could use javascript to remove the link once it's traversed, or set the url to something else, or remove the originating feature etc.  Worth thinking about and experimenting with.  We don't currently have an [experiment for `_blank`](https://maps4html.org/experiments/)

> Also, because the layer is duplicated each time it could potentially have a huge performance impact? 

I suppose.  This is under the control of the author, so kind of their responsibility, I think.

PR #635, opened 2021-12-14
ZoomToLayer fix

### LINKED ISSUE #634: `zoomToLayer` on TemplatedFeatureLayer causes feature duplication
https://user-images.githubusercontent.com/55214462/146027653-918b8110-3064-4660-9277-81a03eb632ca.mp4

It also seems to only happen when the current zoom level is greater than the layers zoom level.

@prushforth Turns out the reason it needed multiple tab presses is because the features are duplicated in the DOM. The bug is related to `zoomToLayer` in combination with `TemplatedFeaturesLayer`, where the `_pullFeatureFeed` is called twice (through the moveend handler) instead of once, not sure why yet.

_Originally posted by @ahmadayubi in https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/596#issuecomment-993654052_

ahmadayubi:
Rather than zooming in/out and checking if a layer is within bounds repeatedly, calculate whether a layer would be in or out of bounds after a set number of zoom in/outs, ahead of time. 

Before: (flickering, instant movements)

https://user-images.githubusercontent.com/55214462/146052220-e0a7e713-5798-4aa4-948d-0c075418e5c1.mp4

After: (no flickering, removing repeated zoom in/out)

https://user-images.githubusercontent.com/55214462/146072850-70e84b9a-35f4-4a8e-855d-6987e96c7ec3.mp4

Closes issue with templatedFeatureLayer duplicates, although that issue may still remain in another form if the underlying bug wasn't the zoomToLayer feature but rather quick successive addition and removal of layers (which the old zoomToLayer feature would do).

Closes #634 

https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/612#issuecomment-982651989

### DIFF
--- a/src/layer.js
+++ b/src/layer.js
@@ -321,34 +321,58 @@ export class MapLayer extends HTMLElement {
       bR = this.extent.bottomRight.pcrs,
       layerBounds = L.bounds(L.point(tL.horizontal, tL.vertical), L.point(bR.horizontal, bR.vertical)),
       center = map.options.crs.unproject(layerBounds.getCenter(true)),
-      currentZoom = map.getZoom();
+      newZoom = map.getZoom();
 
-    map.setView(center, currentZoom, {animate:false});
-    let mapBounds = M.pixelToPCRSBounds(
-      map.getPixelBounds(),
-      map.getZoom(),
-      map.options.projection);
+    let maxZoom = this.extent.zoom.maxZoom, minZoom = this.extent.zoom.minZoom;
 
-    //fits the bounds to the map view
-    if(mapBounds.contains(layerBounds)){
-      while(mapBounds.contains(layerBounds) && (currentZoom + 1) <= this.extent.zoom.maxZoom){
-        currentZoom++;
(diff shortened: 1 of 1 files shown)


prushforth:
It does seem to fix the problem with duplicate markers.  The animation is not quite right though: it flashes dramatically.  To reproduce/ see what I mean, go to this [example](https://maps4html.org/experiments/linking/features/). Click on British Columbia.  Click on the marker.  Use Zoom To  Layer for the BC layer.

ahmadayubi:
The flashing in that certain case is due to the tiles loading in, if you zoom out manually you can see the same effect, the animation is handled by leaflets flyTo function, alternatively you can simply setView to avoid animations altogether.

prushforth:
Let's use the direct to location method then

ahmadayubi:
Maybe a fading effect would help, it would have to be a custom implementation since Leaflet doesn't offer layer fading in, I believe.

Malvoz (approved):

prushforth (approved):
LGTM -- thanks!

Issue #397, opened 2021-04-14
Feature Index


Malvoz:
A Feature Index exposes all features currently in view, in a list. It has the following benefits:
- User can tell what a feature is about without first interacting with it
- Allows interacting with stacked interactive features (otherwise, for example, if you have 2 clickable features, only 1 can be clicked using a mouse pointer or touch). So enables more easily interacting with the desired feature for all [types of users](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html#benefits) (including users who use speech recognition to control a cursor, which is not on that list)
- Allows querying any number of _layers_ (I believe, if they can somehow be exposed in the Feature Index)

For reference, here are screenshots of Feature Indexes as implemented by the [Google Maps site](https://www.google.com/maps) and [Esri's a11y-map](https://github.com/Esri/a11y-map):

<img width="400" src="https://user-images.githubusercontent.com/26493779/114731946-6bfbd280-9d42-11eb-9244-998647056750.png">
<img width="400" src="https://user-images.githubusercontent.com/26493779/114731952-6d2cff80-9d42-11eb-860a-dcec5aa2aade.png">

Malvoz:
I've updated the initial comment as it was a bit vague and left out important details.

Also, I'd like to note that it's not immediately clear to me at this moment if a Feature Index as described in this issue can make sense without https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/396. OTOH https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/396 does not necessarily depend on a Feature Index, see <q>[4. Features in view should be announced](https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/396#:~:text=Features%20in%20view%20should%20be%20announced)</q> options _a)_ and _b)_.

prushforth:
@Malvoz I think that the key buttons in the feature index should be actual buttons, so that mobile / touch device users can press them without access to a keyboard.  WDYT?

PR #636, opened 2021-12-15
Feature index

ben-lu-uw:
https://user-images.githubusercontent.com/69722289/146215389-16b5157b-7cb1-4915-b29b-c36cbff52a32.mov

I tried making the feature index table an aria-live region but it is an earful and the map location does not get announced. Left the table as tabbable for now.


Malvoz:
This looks pretty good. A few thoughts:

- Is `<table>` only used for layout purpose in this case? Note that [HTML semantics are typically not announced in Live Regions](https://www.scottohara.me/blog/2019/07/10/the-output-element.html#allowed-child-content).
  
  -  In quick testing using `role="status" aria-live="polite" aria-atomic="true"` on `mapml-feature-index` it seemed to work pretty well except "Feature index" was announced repeatedly - to solve that it'd have to be placed outside the live region - though I'm not sure we need a `caption` for the feature index, since the intent is to only announce features?

  - Is there a particular reason why `aria-label`s are used on `<td>` when the element's text content is the same? 

- Can `<table>` be changed to `<output role="status" aria-live="polite" aria-atomic="true">`, perhaps containing a list with list items for each feature instead?

> I guess they could share the output element

I think we'll get more control by separating them. For example, we may want to separate the zoom announcement from col and row such that zoom levels aren't announced when panning but only when the zoom level actually changes.

ben-lu-uw:
Yea the table element was only used for layout and so I could copy most of the CSS from debug mode.

Malvoz:
OTOH if we want to use valid HTML we can't use a list, only [phrasing content](https://html.spec.whatwg.org/multipage/dom.html#phrasing-content-2) (per the content model of [`output`](https://html.spec.whatwg.org/multipage/form-elements.html#the-output-element)), so maybe `<span>` for each feature, set to `display: block`?

ahmadayubi:
That looks really good! The naming is somewhat conflicting with https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/596, it maybe also be able to potentially merge together, instead of having to sort twice.

Checking distance could then be done simply by looping through the sorted inbounds array from the PR as that accounts for the subparts of features as well.

Malvoz (commented):
I'm not exactly sure what's in scope for this PR (in relation to https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/397), but here are a few more thoughts:

- [x] The list should probably not have more than 9 items at any given time, such that all features can be accessed using the <kbd>1</kbd>-<kbd>9</kbd> number keys (i.e. only 1 `mapml-feature-index-content` container is needed, where <kbd>9</kbd> would show more features, <kbd>8</kbd> would show the previously shown features):
  
  https://user-images.githubusercontent.com/26493779/146404123-86be9a84-9ac3-4935-b1b8-88f32bf3b500.mp4

  <ins>Fixed in https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/636/commits/19fbf5d9a7bada1b19f4a38970b5fa02fbed361e</ins>
	
- [x] What should happen when a number key is pressed? Probably send focus to the corresponding feature (Esri's a11y-map opens a popup, but we don't expect all features to have a popup)? This would allow the user to discover what type of feature it is (link, button, etc.(?)) before choosing to interact with it.

  <ins>Focus on feature when key is pressed fixed in https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/636/commits/9f33d64274b0c56d5f876955afb60438cd25a9e7</ins>
	  
- [ ] Should <kbd>Esc</kbd> - when focus is on a feature - return focus to the leaflet container? (so that the user doesn't have to tab all the way back to be able to pan the map. Esri does this too.)

Malvoz (commented):
<details><summary><i>Resolved by <a href="https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/636#issuecomment-997015091">#636 (comment)</a></i>.</summary><br>

Feature index bug?

duplicate-entries

</details>

Malvoz (commented):
<details><summary><i>Resolved in <code><a href="https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/636/commits/afe4f56aa3c3c33cbfe4a95d8eb7b0135af2fcc9">afe4f56</a></code></i>.</summary>

Small visual bug when feature index is empty:

empty-feature-index

</details>

ben-lu-uw:
> I'm not exactly sure what's in scope for this PR (in relation to #397)

Sorry forgot to mention, this draft PR is just to get feedback on points 1 and 2 in https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/397 atm.

src/mapml/layers/FeatureIndex.js
```diff
@@ -0,0 +1,147 @@
...
+                            xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+                            xmlns:svg="http://www.w3.org/2000/svg" 
+                            xmlns="http://www.w3.org/2000/svg" version="1.1"
+                            x="0px" y="0px" viewBox="0 0 99.999998 99.999998" xml:space="preserve">
+                                <g>
+                                    <rect  width="100" height="100"
+                                    style="color:#000000;clip-rule:nonzero;display:inline;overflow:visible;isolation:auto;mix-blend-mode:normal;color-interpolation:sRGB;color-interpolation-filters:linearRGB;solid-color:#000000;solid-opacity:1;fill:#000000;fill-opacity:0;fill-rule:nonzero;stroke:#000000;stroke-width:1;stroke-linecap:round;stroke-linejoin:round;stroke-miterlimit:4;stroke-dasharray:none;stroke-dashoffset:0;stroke-opacity:1;color-rendering:auto;image-rendering:auto;shape-rendering:auto;text-rendering:auto;enable-background:accumulate"
+                                    />
+                                </g>
+                            </svg>`;
```
Malvoz, line 15:
```suggestion
        let svgInnerHTML = `<svg xmlns="http://www.w3.org/2000/svg" xml:space="preserve" viewBox="0 0 100 100"><path fill="none" stroke="#000" stroke-linecap="round" stroke-linejoin="round" d="M0 0h100v100H0z" color="#000" overflow="visible"/></svg>`;
```

Optimized the SVG using [SVGOMG](https://jakearchibald.github.io/svgomg/), removes inline styles (see motivation in https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/621#issuecomment-996014756).

src/mapml/layers/FeatureIndex.js
```diff
@@ -0,0 +1,147 @@
...
+            } else if (e && e.type === "blur") {
+                obj._container.querySelector('rect').style.display = "none";
+                obj._output.style.display = "none";
+            } else if (obj._map.isFocused) {
+                obj._container.querySelector('rect').style.display = "inline";
+                obj._output.style.display = "block";
+            } else {
+                obj._container.querySelector('rect').style.display = "none";
+                obj._output.style.display = "none";
+            }
```
Malvoz, line 138:
```suggestion
            if (e && e.type === "focus") {
                obj._container.removeAttribute("hidden");
                obj._output.removeAttribute("hidden");
            } else if (e && e.type === "blur") {
                obj._container.setAttribute("hidden", "");
                obj._output.setAttribute("hidden", "");
            } else if (obj._map.isFocused) {
                obj._container.removeAttribute("hidden");
                obj._output.removeAttribute("hidden");
            } else {
                obj._container.setAttribute("hidden", "");
                obj._output.setAttribute("hidden", "");
            }
```

The `rect` selectors need to change per the [previous suggestion](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/636/files#r771174496). This change also removes inline styles (per motivation linked in that comment).

I changed to hide the entire `mapml-feature-index-box` and not only the SVG within it, though we can change to only hide the SVG if there's a reason not to hide the parent container.

src/mapml.css
```diff
@@ -776,3 +776,45 @@ label.mapml-layer-item-toggle {
...
+* Feature Index
+ */
+.mapml-feature-index-box {
+  margin: -36px 0 0 -36px;
+  width: 72px;
+  height: 72px;
+  left: 50%;
+  top: 50%;
+  content: '';
+  display: block;
```
Malvoz, line 791:
```suggestion
.mapml-feature-index-box:not([hidden]) {
  margin: -36px 0 0 -36px;
  width: 72px;
  height: 72px;
  left: 50%;
  top: 50%;
  content: '';
  display: block;
```

This is needed in conjunction to [a previous suggestion](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/636/files#r771205223) because the UA styles `display: none` for `hidden` is overriden by the `display: block`. 

We can remove this (`:not([hidden])`) once [`[hidden] { display: none!important }`](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/621/files#diff-56ad14054d43b64aa4b9c722e112e44b8769885fad4e9fc9311b63fea37cf643R495-R497) is applicable (after changes from that PR is pulled in or this PR is merged into master).

src/mapml.css
```diff
@@ -776,3 +776,45 @@ label.mapml-layer-item-toggle {
...
+  position: absolute;
+  z-index: 10000;
+}
```
Malvoz, line 794:
```suggestion
  z-index: 10000;
  outline: thin solid #fff;
}
```

Adding a thin outline to ensure the `mapml-feature-index-box` is visible independently of the background:

outline-feature-index-box

(makes me wonder whether we should just do a black border with white outline instead of an SVG? But let's not, for now... 😋)

src/mapml.css
```diff
@@ -776,3 +776,45 @@ label.mapml-layer-item-toggle {
   ) {
   outline: 0!important;
 }
+
+
+/**
+* Feature Index
+ */
+.mapml-feature-index-box {
```
Malvoz, line 784: We should add the `.mapml-feature-index-box` selector to this declaration:

https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/9f1d807e5cac3201512d94ac8c8bddee1c52afc4/src/mapml.css#L432-L439

because currently it's possible to click the feature index box with a mouse pointer (thus preventing underlying features from receiving the click).

ben-lu-uw:
> * [ ]  Should Esc - when focus is on a feature - return focus to the leaflet container? (so that the user doesn't have to tab all the way back to be able to pan the map. Esri does this too.)

Not sure where this is mentioned but I think someone wanted to use arrow keys to navigate features and tabbing would first put focus on the first feature and tabbing again would return focus to the map.

src/mapml.css
```diff
@@ -776,3 +776,45 @@ label.mapml-layer-item-toggle {
...
+  height: 72px;
+  left: 50%;
+  top: 50%;
+  content: '';
+  display: block;
+  position: absolute;
+  z-index: 10000;
+}
+
+.mapml-feature-index {
```
Malvoz, line 796:
```suggestion
.mapml-feature-index:not([hidden]) {
```

Oops I missed this one. Same as https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/636#pullrequestreview-834977702.

Malvoz:
> > Should Esc - when focus is on a feature - return focus to the leaflet container?
>
> Not sure where this is mentioned but I think someone wanted to use arrow keys to navigate features

Yes that's https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/535.

> and tabbing would first put focus on the first feature and tabbing again would return focus to the map.

Actually, tabbing past the features would focus controls, and tabbing past the controls would exit the map.

If features were navigable using arrow keys instead of tabs we'd have:

1. <kbd>Tab</kbd> once to reach container
2. <kbd>Tab</kbd> a second time to reach features
3. <kbd>Tab</kbd> a third time to reach controls

instead of:

1. <kbd>Tab</kbd> once to reach container
2. <kbd>Tab</kbd> a second time to reach features
3. <kbd>Tab</kbd> as many times as there are features to reach controls

In either of those cases the user would have to <kbd>Shift</kbd> + <kbd>Tab</kbd> (whether one time or a hundred) to go from features to the container (to pan). Of course, only having to <kbd>Shift</kbd> + <kbd>Tab</kbd> once is perhaps not much more difficult than to press <kbd>Esc</kbd>, but <kbd>Esc</kbd> may be more intuitive?

ben-lu-uw:
> Feature index bug?
> 
> duplicate-entries

Should be fixed when using https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/596

prushforth:
I noticed that NVDA only announces something when there is more than one feature in the reticle / index.

ben-lu-uw:
Hmm interesting, NVDA does not announce anything when going from hidden to visible for me. Still announces when there is one feature in the index as long as the feature index was previously visible.

Malvoz:
It should be more reliable to toggle the [`mapml-screen-reader-output`](https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/268c947517dcc651c7264cf660037da3994584eb/src/mapml.css#L736-L744) class rather than `hidden` on `mapml-feature-index`. That may also be useful for other reasons (see point 1 in https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/397).

src/mapml/layers/FeatureIndexOverlay.js
```diff
@@ -0,0 +1,167 @@
+export var FeatureIndexOverlay = L.Layer.extend({
+    onAdd: function (map) {
+        let svgInnerHTML = `<svg xmlns="http://www.w3.org/2000/svg" xml:space="preserve" viewBox="0 0 100 100"><path fill="none" stroke="#000" stroke-linecap="round" stroke-linejoin="round" stroke-width="3" d="M0 0h100v100H0z" color="#000" overflow="visible"/></svg>`;
+
+        this._container = L.DomUtil.create("div", "mapml-feature-index-box", map._container);
+        this._container.innerHTML = svgInnerHTML;
+
+        this._output = L.DomUtil.create("output", "mapml-feature-index", map._container);
```
Malvoz, line 8: The output element should have the following attributes:

- `role="status"`
- `aria-live="polite"`
- `aria-atomic="true"`

Motivated by https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/396:

> ```html
> 
> ```

And [MDN live regions](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions#roles_with_implicit_live_region_attributes):

> To maximize compatibility, add a redundant `aria-live="polite"` when using this role.

src/mapml/layers/FeatureIndexOverlay.js
```diff
@@ -0,0 +1,170 @@
...
+                }
+                index += 1;
+            }
+        });
+        this._addToggleKeys();
+    },
+
+    _updateOutput: function (label, index, key) {
+        let span = document.createElement("span");
+        span.setAttribute("index", index);
```
Malvoz, line 81: To be on the safe side, the `index` attribute should probably be [`data-`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/data-*) prefixed (related issue: https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/191).

src/mapml/layers/FeatureIndexOverlay.js
```diff
@@ -0,0 +1,170 @@
...
+            }
+        } else if(key === 56){
+            this._newContent(body, -1);
+        } else if(key === 57){
+            this._newContent(body, 1);
+        }
+    },
+
+    _newContent: function (body, direction) {
+        let index = body.firstChild.getAttribute("index");
```
Malvoz, line 120: The `index` attribute should probably be `data-` prefixed.

src/mapml.css
```diff
@@ -776,3 +777,46 @@ label.mapml-layer-item-toggle {
...
+  left: 5px;
+  padding: 5px 10px;
+  width: -webkit-fit-content;
+  width: -moz-fit-content;
+  width: fit-content;
+  font: inherit;
+}
+.mapml-feature-index-content > span{
+  display: block;
+  font-family: monospace;
```
Malvoz, line 820:
```suggestion
```

This was copied from the debug panel. We shouldn't use that font here IMO.

Malvoz:
There's an issue with features not being announced if they're in range of the feature index box initially.

I've tried many combinations of `aria-label`, `aria-labelledby`, `aria-describedby`, a CSS hack, and an `aria-relevant` hack. Chrome and Firefox seem to treat Live Regions a bit differently it seems, or perhaps they differ in the [Accessible Name and Description Computation](https://www.w3.org/TR/accname/#mapping_additional_nd_te).

This `aria-live` test also shows it is unreliable to play around with the `display` property: https://www.davidmacd.com/blog/test-aria-live-display-none.html.

If a particular solution worked perfectly in Firefox, there was double-speak in Chrome. If it worked perfectly in Chrome, nothing was announced in Firefox. Tested in NVDA and Narrator.

I think the correct thing for now is to avoid any sort of hacks change the following lines:

https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/2fcac6022f66003d9e5a27899904f8e26cd8fcf8/src/mapml-viewer.js#L227-L228

https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/2fcac6022f66003d9e5a27899904f8e26cd8fcf8/src/web-map.js#L266-L267

to either:

```js
this._container.setAttribute('role', 'application'); 
this._container.setAttribute('aria-roledescription', 'Interactive map'); 
this._container.setAttribute('aria-labelledby', 'mapml-feature-index');  // And set this ID on the <output>
```

```js
this._container.setAttribute('role', 'application'); 
this._container.setAttribute('aria-label', 'Interactive map'); 
this._container.setAttribute('aria-describedby', 'mapml-feature-index');  // And set this ID on the <output>
```

and remove these lines:

https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/2fcac6022f66003d9e5a27899904f8e26cd8fcf8/src/mapml-viewer.js#L224

https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/2fcac6022f66003d9e5a27899904f8e26cd8fcf8/src/web-map.js#L263

There will still be double-speak (but at least we're not doing anything too shady), just like on https://www.google.com/maps/ 😉

Malvoz:
> on https://www.google.com/maps/

(you'll have to <kbd>shift</kbd> + <kbd>tab</kbd> twice to get to their feature index to show, or <kbd>tab</kbd> 15~ times and sometimes have it show.)

prushforth:
Hey @Malvoz I don't know if you've got any time these days, but I think this PR could use a review by yourself.  Thanks!

Malvoz:
Looks good to me. It was decided to have the feature index disabled by default?

Things we can deal with after the merge:
- I think there is room for some small CSS improvements
- NVDA doesn't announce the features that are initially in view on the first <kbd>tab</kbd> press to focus the map.
  
  I tested a bunch of different solutions last year and found that it could be improved with some modifications to `aria-label` + `role="region"` on `leaflet-container`. Once that issue was fixed the result was (in Chrome, not Firefox) that the features were announced twice, so I filed: https://bugs.chromium.org/p/chromium/issues/detail?id=1285603

Malvoz (approved):

prushforth:
Yes, we put access to the feature index into the user preferences (extension).  We plan to implement feature navigation with arrow keys as the default, but user configurable.  We left the default at tabbing through features for now.  Is that appropriate?  I realize it doesn't showcase feature indexing.  I believe we could make an experiments page that is hard-coded to use feature index even without the extension installed.

Malvoz:
Anything that relies on the extension is by default inaccessible, so that's a consideration. Esri's a11y-map has a map control to toggle their feature index on/off. But we could indeed start with an experiments page.

I might've stumbled upon Goodmaps before, I usually leave quickly when I don't find a live demo of a map. 😋

prushforth (changes requested):
Thanks , looking much better with your popup fix.  I think it works better, visually at least.  I had a listen on NVDA, and I think if we could make it read out the entire feature index, that would be a better experience (see Robert's [comment](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/636#issuecomment-1132170353), maybe that is relevant).  The user could just listen and hit a key when something piqued their interest.  Another issue that I noticed was that I think the feature index should _replace_  _all_ the features in the tab order i.e. when you hit tab with the focus on the map/feature index, it should skip to the first control in tab order. Note that controls can be "pruned" individually (i.e. see the [controlslist](https://maps4html.org/web-map-doc/docs/elements/mapml-viewer/#controlslist) attribute), so the code will have to take that into account i.e. not hardcode which control gets the focus after the feature index.    Finally, is there a way to "hard code" a page to use the feature index, if the user doesn't have the extension installed and set to select that option?  We should make a copy of the restaurants page in our experiments repo and make a link to it from the index.html page in that repo.

prushforth:
@Malvoz 

> Anything that relies on the extension is by default inaccessible, so that's a consideration.

I understand.  The intention is to have an accessible map experience by default.  I think the feature index will  work well for some users, and tab/arrow keys for others.  Both experiences should be accessible, but I think the user should be able to set a preference for one or the other from the browser settings (in our case we are using our extension as a proxy for browser settings).  The a11ymap control is another possibility.  Would you want to allow a web site to turn it off, though?

ben-lu-uw:
> Finally, is there a way to "hard code" a page to use the feature index, if the user doesn't have the extension installed and set to select that option?

Yep you can just add this script into the head:
```
<script>
    let options = document.createElement("map-options");
    options.innerHTML = '{"featureIndexOverlayOption":true}';
    document.head.appendChild(options);
</script>
  ```

**Comment by Malvoz** [comment]
> - NVDA doesn't announce the features that are initially in view on the first <kbd>tab</kbd> press to focus the map.

Last year I tried a bunch of stuff, and concluded the following to be the most reliable to have the features announced on the initial <kbd>tab</kbd> to the `leaflet-container`:

Change these:

https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/2fcac6022f66003d9e5a27899904f8e26cd8fcf8/src/mapml-viewer.js#L227-L228

https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/2fcac6022f66003d9e5a27899904f8e26cd8fcf8/src/web-map.js#L266-L267

to either:

```js
this._container.setAttribute('role', 'application'); 
this._container.setAttribute('aria-roledescription', 'Interactive map'); 
this._container.setAttribute('aria-labelledby', 'mapml-feature-index');  // And set this ID on the <output>
```

or

```js
this._container.setAttribute('role', 'application'); 
this._container.setAttribute('aria-label', 'Interactive map'); 
this._container.setAttribute('aria-describedby', 'mapml-feature-index');  // And set this ID on the <output>
```

and remove these lines:

https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/2fcac6022f66003d9e5a27899904f8e26cd8fcf8/src/mapml-viewer.js#L224

https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/2fcac6022f66003d9e5a27899904f8e26cd8fcf8/src/web-map.js#L263

(These changes may affect https://github.com/Maps4HTML/Web-Map-Custom-Element/discussions/558)

Malvoz:
Re https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/636#pullrequestreview-979418583:

> Another issue that I noticed was that I think the feature index should replace all the features in the tab order i.e. when you hit tab with the focus on the map/feature index, it should skip to the first control in tab order.

If I understand correctly, what you're suggesting is "[Changing DOM order](https://github.com/Leaflet/Leaflet/issues/7479#changing-dom-order)" in https://github.com/Leaflet/Leaflet/issues/7479, whereas I think [arrow keys for feature navigation](https://github.com/Leaflet/Leaflet/issues/7479#arrow-key-navigation) (https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/535) may be a more desirable solution.

prushforth:
> If I understand correctly, what you're suggesting is "https://github.com/Leaflet/Leaflet/issues/7479#changing-dom-order" in https://github.com/Leaflet/Leaflet/issues/7479, whereas I think https://github.com/Leaflet/Leaflet/issues/7479#arrow-key-navigation (https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/535) may be a more desirable solution.

No, I think option C: Arrow keys is desirable, and I want us to implement it. But I think it should be EITHER feature index OR a tab stop + arrow keys for features (right now it is feature index then as many tabs as it takes to get through all features), although if we are able to implement arrow keys it would only be one tab stop, so perhaps not that big a deal.

So in the short term I think we should replace all the features' tab stops with the feature index, if that's doable.

ben-lu-uw:
> So in the short term I think we should replace all the features' tab stops with the feature index, if that's doable (edit only when the feature index is enabled).

So only investigate features using the number keys when using the feature index?
And then tab focuses the map and tabbing again would go to controls?

https://user-images.githubusercontent.com/69722289/169568822-959ca081-b1e3-4e11-a3fd-622752c4ec11.mp4

prushforth:
Yes I think so, hopefully Robert agrees

Malvoz:
Yes I think that should be fine. 🙂

prushforth:
So I think this is pretty good right now.  The only issue that Ben and I discussed was that when NVDA is running, and you select a feature by index number, you have to dismiss the popup in order to select another feature, whereas if NVDA is **not** running, you can select another feature from the index while the popup for the first feature is displayed.  We believe this has to do with NVDA switching to browse mode when the popup is read, re-mapping the index number keys to heading level shortcuts, something we feel we should not attempt to change (the popup could have headings and so on, who can say).  
Ben will push the changes shortly.

prushforth (approved):

Malvoz:
Sorry I have limited time to (re-)review, I'm about to go on vacation. I'm seeing the following after opening a dialog:

feature-index-box-overlapping

This could be fixed [similar to `mapml-outline`](https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/c38f32ec0de598819770af903237c6d5cee6e39e/src/mapml.css#L454-L459) for example:

```css
.leaflet-container:not(:focus) .mapml-feature-index-box {
  display: none;
}
```

However there'd still be the case where the `mapml-feature-index-box`'s visibility flashes on mouse click on the map.

--- 

`mapml-feature-index` is now both `mapml-screen-reader-output` (screen reader accessible but visibly hidden) and `aria-hidden="true"` if the map isn't focused. Should it just toggle the `hidden` attribute instead then?

OTOH, my intention was to always have the features accessible whether or not the feature index was visibly hidden or not. Though obviously some screen readers don't like to reread things even after a user focuses an element, thus we had the issue of features not being announced after the map was tabbed to. Toggling `aria-hidden` helps with that, but now features aren't announced on e.g. zoom from the +/- controls, etc.? 

---

> The only issue that Ben and I discussed was that when NVDA is running, and you select a feature by index number, you have to dismiss the popup in order to select another feature, whereas if NVDA is not running, you can select another feature from the index while the popup for the first feature is displayed. We believe this has to do with NVDA switching to browse mode when the popup is read

That's correct (https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/470):

https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/b108e33535b9aad61be269a0c97db2460d28e155/src/mapml/layers/MapLayer.js#L1246

A potentially valid workaround is to move the `document` role to a new wrapper element for the content in `mapml-popup-content`.

Unfortunately this will be the responsibility of all authors after https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/403. 😕

---

Longing for the day we'll be doing: 1) `<map(viewer)>` 2) Done. 😏

ben-lu-uw:
> Should it just toggle the hidden attribute instead then?

I chose to to toggle the aria-hidden attribute since toggling hidden would sometimes make it visually hidden in cases where it shouldn't be, like when opening popups since at that time the map is not focused.

prushforth:
> now features aren't announced on e.g. zoom from the +/- controls

Features are announced when the map is focused, and if you use the +/- _keys_ with the map focused, features _are_ announced.   Is that acceptable? I'm not sure if it would be a good idea to announce features when the control is focused anyway?

prushforth:
OK, looks good, big milestone 🍾 🥳 @ben-lu-uw congratulations.  We will deal with arrow key navigation soon.

Malvoz:
> > now features aren't announced on e.g. zoom from the +/- controls
> 
> Features are announced when the map is focused, and if you use the +/- keys with the map focused, features are announced. Is that acceptable? I'm not sure if it would be a good idea to announce features when the control is focused anyway?

We expect visual feedback when we click the zoom controls, likewise AT users would expect auditory feedback.

Users on touch devices (using e.g. TalkBack, VoiceOver) can't use the feature index in the [demo](https://maps4html.org/experiments/screenreader/feature-index/)  because it remains `aria-hidden="true"`.

prushforth:
> We expect visual feedback when we click the zoom controls, likewise AT users would expect auditory feedback.

Agreed.  I think we should make it so that zoom level changes are always announced, at initial load and whenever it changes, not just with the announce movement option enabled.

PR #640, opened 2021-12-21
Change how alternate styles are handled in the layer control

### LINKED ISSUE #639: Changing style in layer control is too immediate, doesn't allow user to listen to available options
This is a result of using a radio group for this control, and not implementing a custom control; the change / click event is used to trigger the replacement of the layer immediately, and thus the radio button labels don't get read.  

This is a separate issue from #462, but related.

prushforth:
Closes #639.

### DIFF
--- a/src/mapml/layers/MapLayer.js
+++ b/src/mapml/layers/MapLayer.js
@@ -861,31 +861,62 @@ export var MapMLLayer = L.Layer.extend({
                 M.parseStylesheetAsHTML(mapml, base, layer._container);
                 var styleLinks = mapml.querySelectorAll('map-link[rel=style],map-link[rel="self style"],map-link[rel="style self"]');
                 if (styleLinks.length > 1) {
+                  var changeStyle = function (e) {
+                    var href = getStyleURL(e);
+                    if (e.type === 'focusout' && !isStyleOption(e.relatedTarget) && href != layer._href) {
+                      layer.fire('changestyle', {src: href}, false);
+                    }
+                    function isStyleOption(item) {
+                      var form = e.target.form,  
+                          stylesRadioNodeList = form.elements["styles-"+L.stamp(layer)],
+                          isRB = false;
+                      for(var i=0;i<stylesRadioNodeList.length;i++) {
+                        if (item === stylesRadioNodeList[i]) {
+                          isRB = true;
+                          break;
+                        }
+                      }
+                      return isRB;
+                    }
+                    function getStyleURL(e) {
+                      var form = e.target.form,
+                      stylesRadioNodeList = form.elements["styles-"+L.stamp(layer)],
+                      href;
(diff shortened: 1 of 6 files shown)


src/mapml/layers/MapLayer.js
```diff
@@ -861,31 +861,62 @@ export var MapMLLayer = L.Layer.extend({
...
+                          break;
+                        }
+                      }
+                      return href;
+                    }
+                  };
                   var stylesControl = document.createElement('details'),
-                  stylesControlSummary = document.createElement('summary');
+                  stylesControlSummary = document.createElement('summary'),
+                  stylesForm = document.createElement('form');
```
Malvoz, line 896: Is an additional (form) element required?

src/mapml/layers/MapLayer.js
```diff
@@ -861,31 +861,62 @@ export var MapMLLayer = L.Layer.extend({
...
                   for (var j=0;j<styleLinks.length;j++) {
                     var styleOption = document.createElement('div'),
                     styleOptionInput = styleOption.appendChild(document.createElement('input'));
                     styleOptionInput.setAttribute("type", "radio");
                     styleOptionInput.setAttribute("id", "rad-"+L.stamp(styleOptionInput));
-                    styleOptionInput.setAttribute("name", "styles-"+layer._title);
+                    styleOptionInput.setAttribute("name", "styles-"+L.stamp(layer));
                     styleOptionInput.setAttribute("value", styleLinks[j].getAttribute('title'));
                     styleOptionInput.setAttribute("data-href", new URL(styleLinks[j].getAttribute('href'),base).href);
+                    styleOptionInput.setAttribute("tabindex", "0");
```
Malvoz, line 911: Is tabindex required? The input elements are focusable by default.

src/mapml/layers/MapLayer.js
```diff
@@ -861,31 +861,62 @@ export var MapMLLayer = L.Layer.extend({
                 M.parseStylesheetAsHTML(mapml, base, layer._container);
                 var styleLinks = mapml.querySelectorAll('map-link[rel=style],map-link[rel="self style"],map-link[rel="style self"]');
                 if (styleLinks.length > 1) {
+                  var changeStyle = function (e) {
```
Malvoz, line 864: Using a mouse pointer it feels a bit unresponsive to not have the style change on click. And using a keyboard it feels a bit disruptive to have the style change happen before intentionally exiting the layer control. But it's certainly an improvement, and fixes the bug. 👍🏼

src/mapml/layers/MapLayer.js
Malvoz: Would have to modify the following selector in this case, as it causes a small misalignment between the label and input:

https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/816f7c19372bae1da571f74656b38a555d08bbab/src/mapml.css#L226-L231

Issue #641, opened 2021-12-21
Attribution links are announced twice


Malvoz:
The non-default attribution links have a `title` same as the text content which results in some screen readers announcing the same thing twice.

AliyanH:
fixed by #815

PR #642, opened 2021-12-21
Attribution control

Malvoz:
- [x] https://github.com/Maps4HTML/Web-Map-Custom-Element/commit/7ed9d2742c2146de355e71267f6a5db0b862998d: Fixes https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/216.
- [x] https://github.com/Maps4HTML/Web-Map-Custom-Element/commit/241cf223a591742bacb5da9f500c5ed79f676dbc: Fixes https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/641.

## Preview

(colors seen below are outdated, see https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/642#discussion_r780915271)

Tooltip on hover:

closed-hover

Default attribution opened (we now have room to spell out "Maps4HTML" (CG) which I think seems a bit more enticing):

default-open

Keyboard toggling:

https://user-images.githubusercontent.com/26493779/146924233-2ced3aeb-2436-4f3a-a65f-098ef1809cc3.mp4

### DIFF
--- a/src/mapml.css
+++ b/src/mapml.css
@@ -140,19 +140,52 @@
 .leaflet-tooltip-pane .leaflet-tooltip,
 .leaflet-touch .leaflet-control-layers,
 .leaflet-touch .leaflet-bar,
-.leaflet-popup-tip {
-  box-shadow: 0 1px 3px rgba(0, 0, 0, 0.2), 0 1px 2px rgb(0, 0, 0, 0.3);
-}
-
-/* Increase contrast between the attribution container and the underlying content. */
+.leaflet-popup-tip,
 .leaflet-container .leaflet-control-attribution {
-  background-color: rgba(255, 255, 255, 0.95);
+  box-shadow: rgb(0 0 0 / 30%) 0px 1px 4px -1px;
 }
 
 /* Remove opinionated styles of attribution links. */
 .leaflet-control-attribution a {
   color: revert;
   text-decoration: revert;
+}
+
+
+/*
+ * Attribution.
+ */
+
+.leaflet-container .leaflet-control-attribution {
+  background-color: #fff;
+  border-radius: 1.5em;
+  color: currentColor;
+  margin: 5px;
+  min-height: 30px;
+  min-width: 30px;
+  padding: 0;
+}
+
+.leaflet-control-attribution summary {
+  display: block;
+  list-style: none;
+  border-radius: 100%;
+  position: absolute;
+  bottom: 0;
+  left: calc(100% - 30px);
+  line-height: 0;
+  width: 30px;
+  height: 30px;
+}
+
+.leaflet-control-attribution summary svg {
+  border-radius: 100%;
+  width: inherit;
+  height: inherit;
+}
+
+.mapml-attribution-container {
+  padding: 5px 35px 5px 10px;
 }
 
 /*
--- a/src/mapml/control/AttributionButton.js
+++ b/src/mapml/control/AttributionButton.js
@@ -1,0 +1,110 @@
+export var AttributionButton = L.Control.extend({
+	options: {
+		position: 'bottomright',
+		prefix: '<a href="https://www.w3.org/community/maps4html/" title="W3C Maps for HTML Community Group">Maps4HTML</a> <span aria-hidden="true">|</span> <a href="https://leafletjs.com" title="A JS library for interactive maps">Leaflet</a>'
+	},
+
+	initialize: function (options) {
+		L.Util.setOptions(this, options);
+		this._attributions = {};
+	},
+
+	onAdd: function (map) {
+		map.attributionControl = this;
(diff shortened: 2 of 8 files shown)


Malvoz:
I don't know if this is the most optimal way of adding/extending the [leaflet attribution control](https://github.com/Leaflet/Leaflet/blob/1d3e3a083361c52fb1d0b8104452afa0679eeee5/src/control/Control.Attribution.js), and I'm not comfortable writing eventual tests. The styling of the control (e.g. icon) can be changed of course, but I think it's looking pretty solid, also no RTL issues.

prushforth:
Looks good to me, thank you!

src/mapml/control/AttributionButton.js
```diff
@@ -0,0 +1,110 @@
...
+		var prefixAndAttribs = [];
+
+		if (this.options.prefix) {
+			prefixAndAttribs.push(this.options.prefix);
+		}
+		if (attribs.length) {
+			prefixAndAttribs.push(attribs.join(', '));
+		}
+
+		this._container.innerHTML = '<summary title="Map data attribution"><svg aria-hidden="true" xmlns="http://www.w3.org/2000/svg" height="30px" viewBox="0 0 24 24" width="30px" fill="currentColor"><path d="M0 0h24v24H0z" fill="none"/><path d="M12 2C6.48 2 2 6.48 2 12s4.48 10 10 10 10-4.48 10-10S17.52 2 12 2zm1 15h-2v-6h2v6zm0-8h-2V7h2v2z"/></svg></summary>' + '<div class="mapml-attribution-container">' + prefixAndAttribs.join(' <span aria-hidden="true">|</span> ') + '</div>';
```
prushforth, line 93: Seems that this is the only major difference between Leaflet's attribution control code and this version; maybe we could achieve the same effect by accessing the attributionControl._container property and performing this statement?

src/web-map.js
```diff
@@ -224,13 +224,12 @@ export class WebMap extends HTMLMapElement {
             // there is a better configuration than that, but at this moment
             // I'm not sure how to approach that issue.
             // See https://github.com/Maps4HTML/MapML-Leaflet-Client/issues/24
-            fadeAnimation: true
+            fadeAnimation: true,
+            attributionControl: false
           });
           this._addToHistory();
           // the attribution control is not optional
-          this._attributionControl =  this._map.attributionControl.setPrefix('<a href="https://www.w3.org/community/maps4html/" title="W3C Maps for HTML Community Group">Maps4HTML</a> | <a href="https://leafletjs.com" title="A JS library for interactive maps">Leaflet</a>');
```
prushforth, line 231: The leaflet object exposes the .getContainer() method to get a handle on their HTML.

src/mapml/control/AttributionButton.js
```diff
@@ -0,0 +1,110 @@
+export var AttributionButton = L.Control.extend({
```
ahmadayubi, line 1: It looks really clean imo.

One suggestion is if it extends Leaflet's Attribution, you can extend Attribution directly and simply overwrite the functions you need changed.

```js
export var AttributionButton = L.Attribution.extend({...
```

src/mapml.css
```diff
@@ -155,6 +151,43 @@
...
 
+
+/*
+ * Attribution.
+ */
+
+.leaflet-container .leaflet-control-attribution {
+  background-color: #fff;
+  border-radius: 1.5em;
+  color: currentColor;
```
ahmadayubi, line 162: This may just be a personal preference but the button may fit the overall theme if it's dark on a light background rather than light on a dark background.

Malvoz: I used dark on light initially, but felt better after trying light on dark. They're different variants of the SVG though. I can post screenshots of dark on light too, and we'll decide then?

ahmadayubi: I'll take your word for it, it probably has better contrast being light on dark anyways.

Malvoz: A compromise would be dark on light but with a dark (circular) stroke instead?

src/mapml/control/AttributionButton.js
Malvoz: I wasn't able to get that to work, feel free to make the suggested changes.

src/mapml.css
Malvoz: Changed to dark on light in https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/642/commits/c6420be80d66a2517a31bac9a9a79912dbe68309. I think it is a better default, and once we've enabled customization (using CSS Parts or Custom properties) the theme color can easily be changed by developers.

prushforth:
Working on merging these commits [here](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/815)

PR #657, opened 2022-04-05
Add Ukraine flag from Malvoz / Leaflet 1.8 to attribution control.

prushforth:


prushforth:
Thought it couldn't wait to add support for Ukraine, so I copied your flag and deployed it here.  🙇

Malvoz:
Actually, it is @mourner's flag! 😄

Issue #438, opened 2021-04-28
Implement the ` input step` attribute to enable scaling content across zoom intervals


prushforth:
MapML [specifies](https://maps4html.org/MapML/spec/#the-input-element-0) the `step` attribute, but doesn't really define the semantics.  

We've been discussing what might be a declarative way to allow the client to use the resources it requests for a the map at a particular zoom level to be scaled across a few zoom levels, until it reaches a level at which more resources may be obtained.  This would enable, for example, the ability to conserve bandwidth by scaling vectors or tiles or both, obtained at a lower zoom level through each higher level until either reaching a zoom limit or obtaining more / different representations at a higher zoom level.

You can observe this behaviour for azure maps samples [here](https://azuremapscodesamples.azurewebsites.net/Controls/Fullscreen%20control.html), by looking at the debug tools network trace and zooming in successively. Every few zoom levels, new vector tiles are obtained at an individual zoom level, and those tiles are scaled through two or three more zoom levels before being replaced with tiles at the new zoom level.

We were speculating that the `<input type="zoom" step="n">` attribute could be used to roughly control the zoom levels at which a map extent representation should request tiles, features or both.

For example

```html
<layer- label="Zoomy" checked>
  <meta name="zoom" content="min=0,max=7">
  <extent units="WGS84">
    <input name="zoomLevel" type="zoom" value="1" min="0" max="4" step="3"/>
    <input name="row" type="location" axis="row" units="tilematrix" min="0" max="2"/>
    <input name="col" type="location" axis="column" units="tilematrix" min="0" max="2"/>
    <link tms rel="tile" tref="https://maps4html.org/TiledArt-Rousseau/TheBanksOfTheBièvreNearBicêtre/{zoomLevel}/{col}/{row}.png" />
  </extent>
</layer->
```

From the above, I think the client could deduce that there are tiles available, at a minimum, at z=0, z=3 and z=4, so it could use the tiles it obtains at z=0 for zoom levels 0, 1, and 2 up to but not including 3, where it could obtain more tiles.  The behaviour at z=4 would require some consideration: should it by default obtain tiles at z=4 when zooming to zoom level 4, or should it scale those tiles it obtained at z=3 through 4, 5 up to but not including 6, where it would obtain the z=4 tiles and scale them if re-visiting 5 or 4.

The scaling would not only apply to tiles, but also to vector features obtained via a link template.

prushforth:
[...]

PR #662, opened 2022-05-16
Implement the step attribute

ben-lu-uw:
Closes https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/438


src/mapml/layers/TemplatedFeaturesLayer.js
```diff
@@ -76,14 +80,40 @@ export var TemplatedFeaturesLayer =  L.Layer.extend({
...
+    },
+
     _onMoveEnd: function() {
+      let history = this._map.options.mapEl._history;
+      let current = history[history.length - 1];
+      let previous = history[history.length - 2];
+      let step = this._template.step;
       let mapZoom = this._map.getZoom();
+      let steppedZoom = mapZoom;
+      if (((step !== "1") && ((mapZoom + 1) % step === 0) && current.zoom === previous.zoom - 1) ||
```
prushforth, line 98: When you first load the layer, previous can be undefined here, which happens when you press the "Reload" control button

src/mapml/layers/TemplatedFeaturesLayer.js
```diff
@@ -76,14 +80,41 @@ export var TemplatedFeaturesLayer =  L.Layer.extend({
...
+          (current.zoom === previous.zoom) ||
+          (Math.floor(mapZoom / step) * step !== Math.floor(previous.zoom / step) * step)) {
+          steppedZoom = Math.floor(mapZoom / step) * step;
+      }
+      else if(mapZoom % this._template.step !== 0) return;
+
+      let scaleBounds = this._map.getPixelBounds(this._map.getCenter(), steppedZoom);
+      let url = this._getfeaturesUrl(steppedZoom, scaleBounds);
+      if(url === this._url) return;
```
prushforth, line 108: I think it would be valuable to add a comment about why the early return

src/mapml/layers/TemplatedFeaturesLayer.js
```diff
@@ -76,14 +80,41 @@ export var TemplatedFeaturesLayer =  L.Layer.extend({
...
+
       let mapBounds = M.pixelToPCRSBounds(this._map.getPixelBounds(),mapZoom,this._map.options.projection);
       this.isVisible = mapZoom <= this.zoomBounds.maxZoom && mapZoom >= this.zoomBounds.minZoom && 
                         this.extentBounds.overlaps(mapBounds);
       
       this._features.clearLayers();
-      if(!(this.isVisible)){
+      this._removeCSS();
+      if(!(this.isVisible) && steppedZoom === mapZoom){
+        this._url = "";
```
prushforth, line 120: Early return comment

src/mapml/layers/TemplatedFeaturesLayer.js
```diff
@@ -76,14 +80,41 @@ export var TemplatedFeaturesLayer =  L.Layer.extend({
...
+            let parent = toDelete[i].parentNode;
+            parent.removeChild(toDelete[i]);
+        }
+    },
+
     _onMoveEnd: function() {
+      let history = this._map.options.mapEl._history;
+      let current = history[history.length - 1];
+      let previous = history[history.length - 2];
+      if(!previous) previous = current;
```
prushforth, line 95: Good opportunity for [nullish coalescing](https://twitter.com/thelittlestdev/status/1415213119731814401?s=20&t=eoaqarkZSLddrw1w62iG-g): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

src/mapml/layers/TemplatedFeaturesLayer.js
```diff
@@ -114,7 +145,9 @@ export var TemplatedFeaturesLayer =  L.Layer.extend({
             }
           }));
         };
-      _pullFeatureFeed(this._getfeaturesUrl(), MAX_PAGES)
+
+      this._url = url;
+      _pullFeatureFeed(url, MAX_PAGES)
```
prushforth, line 153: The then block could use a comment or two

src/mapml/layers/TemplatedFeaturesLayer.js
```diff
@@ -65,6 +65,10 @@ export var TemplatedFeaturesLayer =  L.Layer.extend({
               }
             }));
           };
+        if(map.getZoom() % this._template.step !== 0) {
```
prushforth, line 68: I'm curious if because we're calling the _onMoveEnd method here if we couldn't get rid of the call to _pullFeatuerFeed on line 72? Or would that unleash the hounds...

ben-lu-uw: The _pullFeatureFeed does not get called if the statement above happens

prushforth: _onMoveEnd calls _pullFeatureFeed eventually.  Just wondering if the conditions could be moved into _onMoveEnd, removing code / reducing complexity. perhaps not.

ben-lu-uw: I'm not sure why this was the case in the first place. Removing the _pullFeatureFeed from onAdd and leaving the map.fire('moveeend') to trigger _onMoveend as before does not seem to change anything.

prushforth: Can you put that code into another branch and tell me what the branch is? I'll look at it and if we agree it's good we can merge it into this branch to be included in the PR...

ben-lu-uw: https://github.com/ben-lu-uw/Web-Map-Custom-Element/pull/3

PR #214, opened 2020-12-01
Fixes disabled property and removes ability to render layers with mismatched projections

ahmadayubi:
Closes #211 and closes #213

PR #670, opened 2022-06-24
Refactor templated feature layers onAdd

ben-lu-uw:
Addresses https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/662#discussion_r903823184
Duplicate _pullFeatureFeeds in onAdd and onMoveEnd even though onAdd later calls onMoveEnd thus doing the same thing twice

### DIFF
--- a/src/mapml/layers/TemplatedFeaturesLayer.js
+++ b/src/mapml/layers/TemplatedFeaturesLayer.js
@@ -20,9 +20,7 @@ export var TemplatedFeaturesLayer =  L.Layer.extend({
     },
     onAdd: function () {
       this._map._addZoomLimit(this);
-      var mapml, headers = new Headers({'Accept': 'text/mapml'});
-          var parser = new DOMParser(),
-          opacity = this.options.opacity || 1,
+      var opacity = this.options.opacity || 1,
           container = this._container,
           map = this._map;
       if (!this._features) {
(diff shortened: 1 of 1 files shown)


src/mapml/layers/TemplatedFeaturesLayer.js
```diff
@@ -45,36 +43,8 @@ export var TemplatedFeaturesLayer =  L.Layer.extend({
...
-            return;
-        }
-      _pullFeatureFeed(this._getfeaturesUrl(), 10)
-        .then(function() { 
-              map.addLayer(features);
-              map.fire('moveend');  // TODO: replace with moveend handler for layer and not entire map
-        })
-        .catch(function (error) { console.log(error);});
+
+      map.fire('moveend');  // TODO: replace with moveend handler for layer and not entire map
```
prushforth, line 47: What would happen if we replaced this event firing with "this.redraw()"?

ben-lu-uw: Seemingly nothing changes, same with calling _onMoveEnd directly however not sure why ` map.fire('moveend');` was needed in the first place so needs a bit more digging

ben-lu-uw: Needed to fix some bugs: https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/214

prushforth: Seems that the map moveend firing may cover up some issues in other templated layers, so we can put that consideration aside for now.

PR #682, opened 2022-08-09
GeoJSON API

AliyanH:
* **geojson2mapml** - convert a GeoJSON feature or feature collection string or object to a MapML <layer-> element.
* **mapml2geojson** - serialize a MapML <layer-> or <map-feature> element as a GeoJSON feature collection object

Needs to be moved to an appropriate place. Needs more testing + Updates.


prushforth:
I would rename the geojson2mapml.js to geojson.js, and put it in its own subdirectory of src.  You could then make or modify a grunt task (copy) to copy the processed file into a dist/lib subdirectory (say).  Similar for testing, this category of script should be tested in its own directory of the test/e2e directory.  Not quite sure what to do with the sibling directories of e2e for now...

AliyanH:
@prushforth I have added additional optional parameters for the geojson2mapml and mapml2geojson functions based on the [doc ](https://maps4html.org/web-map-doc/docs/api/geojson-api). 

The 'transform' parameter for mapml2geojson has not yet been implemented, so the function currently should only be used for layers with gcrs. How should we go about transforming coordinates to gcrs?

AliyanH:
The GeoJSON API is no longer a 'module' and can now be included in html documents using script tags. The GeoJSON API only works when it accompanies the mapml-viewer.js script. i.e.

<script type="module" src="mapml-viewer.js"></script>
<script src="lib/geojson.js"></script>

More details will be documented in the web-map-doc

AliyanH:
> Would it be possible to convert our experiments that use the NRCan's geolocation and place names APIs to use this API?
> 
> Also, I think we should update [web-map-doc](https://maps4html.org/web-map-doc/docs/api/geojson-api) to include a code player with examples of this API in action.
> 
> We could merge both those separate PRs once this gets merged.

Yup good idea, updating the experiment applications with this would be good (will do that once this is finalized). I have been working on web-map-doc for this API, will add the draft PR soon.

prushforth:
Thanks, I figured

src/geojson/geojson.js
```diff
@@ -0,0 +1,453 @@
...
+    let geometryType = ["POINT", "LINESTRING", "POLYGON", "MULTIPOINT", "MULTILINESTRING", "MULTIPOLYGON", "GEOMETRYCOLLECTION"];
+    let jsonType = json.type.toUpperCase();
+    let layer = MapML;
+    let out = "";
+
+    // HTML parser
+    let parser = new DOMParser();
+
+    // initializing layer
+    if (layer == null) {
```
prushforth, line 56: Should be "==="

src/geojson/geojson.js
```diff
@@ -0,0 +1,453 @@
...
+            row.appendChild(th);
+            row.appendChild(td);
+        }
+    }
+    return table;
+}
+
+// Takes GeoJSON Objects and returns a <layer-> Element
+// geojson2mapml: geojson <layer-> -> <layer->
+function geojson2mapml(json, properties = null, geometryFunction = null, MapML = null) {
```
prushforth, line 42: Is MapML a wasted variable?  Maybe we should just use layer here...

src/geojson/geojson.js
```diff
@@ -0,0 +1,452 @@
...
+        let clone_feature = feature.cloneNode(true);
+        let curr_feature = clone_feature.querySelector('map-feature');
+
+        // Setting Properties
+        let p;
+        // if properties function is passed
+        if (typeof properties === "function") {
+            p = properties(json.properties);
+        } else if (typeof properties === "string") { // if properties string is passed
+            p = parser.parseFromString(properties, "text/xml").childNodes[0];
```
prushforth, line 114: I think we should be using the HTML parser here. I realize that you will not know what to insert because you end up getting a whole document back. But there's the insertAdjacentHTML method that parses string into a passed element, so we could create a temporary element and parse the string into it afterbegin, select the first child and return it. At least I think that might work. Also, check out the security considerations on that page. Is this considered user input??  Is sanitizing this our responsibility, or the responsibility of the developer using the function?

src/geojson/geojson.js
```diff
@@ -0,0 +1,452 @@
...
+        } else { // If no properties function or string is passed
```
prushforth, line 115: I don't know how valuable it will be, not too much beyond the string parameter, but we could include:

        } else if (properties instanceof HTMLElement) {
            p = properties;

before the last else, which would match the already-written [documentation](https://maps4html.org/web-map-doc/docs/api/geojson-api#methods) (<Function | String | **HTMLElement**> properties)

src/geojson/geojson.js
```diff
@@ -0,0 +1,452 @@
...
+        //console.log(p);
+        curr_feature.querySelector('map-properties').appendChild(p);
+
+        // Setting map-geometry
+        let g = geojson2mapml(json.geometry, properties, geometryFunction, layer);
+        if (typeof geometryFunction === "function") {
+            curr_feature.querySelector('map-geometry').appendChild(geometryFunction(g, json));
+        } else {
+            curr_feature.querySelector('map-geometry').appendChild(g);
+        }
```
prushforth, line 128: I'm confused why the "g" is a parameter of a passed geometryFunction -- shouldn't it just be the json.geometry that is passed, with the expectation that a \<mapml-geometry> is returned?

src/geojson/geojson.js
```diff
@@ -0,0 +1,452 @@
...
+            p = parser.parseFromString(properties, "text/xml").childNodes[0];
+        } else { // If no properties function or string is passed
+            p = properties2Table(json.properties);
+        }
+        
+        //console.log(p);
+        curr_feature.querySelector('map-properties').appendChild(p);
+
+        // Setting map-geometry
+        let g = geojson2mapml(json.geometry, properties, geometryFunction, layer);
```
prushforth, line 123: Not sure why the geometryFunction is passed here.  Shouldn't we be passing a defaultGeometryFunction and **then** the geometryFunction on line 125 if applicable?

src/geojson/geojson.js
AliyanH: ended up using [insertAdjacentHTML](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML) method to parse the string. Will document the need to sanitize on web-map-doc.

src/geojson/geojson.js
```diff
@@ -0,0 +1,494 @@
...
+        layer = parser.parseFromString(xmlStringLayer, "text/html");
+        //console.log(layer)
+        if (options.label !== null) {
+            layer.querySelector("layer-").setAttribute("label", options.label);
+        } else if (json.name) {
+            layer.querySelector("layer-").setAttribute("label", json.name);
+        } else if (json.title) {
+            layer.querySelector("layer-").setAttribute("label", json.title);
+        } else{
+            layer.querySelector("layer-").setAttribute("label", "Layer");
```
prushforth, line 83: This could be an internationalization opportunity / need, _perhaps_.  Of course the mapml-extension carries all these strings.

src/geojson/geojson.js
```diff
@@ -0,0 +1,494 @@
...
+    // initializing layer
+    if (layer === null) {
+        // creating an empty mapml layer
+        let xmlStringLayer = "<layer- label='' checked><map-meta name='projection' content='" + options.projection + "'></map-meta><map-meta name='cs' content='gcrs'></map-meta></layer->";
+        layer = parser.parseFromString(xmlStringLayer, "text/html");
+        //console.log(layer)
+        if (options.label !== null) {
+            layer.querySelector("layer-").setAttribute("label", options.label);
+        } else if (json.name) {
+            layer.querySelector("layer-").setAttribute("label", json.name);
```
prushforth, line 79: using json.name and json.title implies there are member variables that have these names, but I don't think that follows the GeoJSON spec (I haven't looked, but I guess nothing could _forbid_ it).   There are two cases: 1) "type": "Feature" and 2) "type": "FeatureCollection".  In case 1), there's often a "properties" member; in case 2) theres typically a "features" member (array).  For the purpose of this comment, we're looking to set the label attribute for the layer, and in case 1) (a single feature) it makes sense to use json.name/title OR json.properties.name/title, but for case 2) (a feature collection) it ONLY makes sense to use json.name/title

prushforth: Anyway, all that to say it's probably ok the way you've got it now.

AliyanH: json.title is mentioned in the geojson spec under "[foreign member](https://www.rfc-editor.org/rfc/rfc7946#section-6.1)", so I think the need for title is there, similarly json.name is used by many GIS applications when exported as geojson, such as QGIS. 

For case 1 json.properties.name/title can be a good resource to use if available also, perhaps we can think about adding the label from properties with something similar to what we do with the "caption" option, where users can insert a property name to have the caption set to the property value.

src/geojson/geojson.js
```diff
@@ -0,0 +1,494 @@
...
+        if (json.bbox) {
+            layer.querySelector("layer-").insertAdjacentHTML("afterbegin", "<map-meta name='extent' content='top-left-longitude=" + json.bbox[0] + ", top-left-latitude=" + json.bbox[1] + ", bottom-right-longitude=" + json.bbox[2] + ",bottom-right-latitude=" + json.bbox[3] + "'></map-meta>");
+        }
+
+        // Setting featurecaption
+        let featureCaption = layer.querySelector("layer-").getAttribute('label');
+        if (typeof options.caption === "function") {
+            featureCaption = options.caption(json);
+        } else if (typeof options.caption === "string") {
+            featureCaption = json.properties[options.caption];
```
prushforth, line 142: I think it might be preferable to actually use the string value as the caption in the case where `json.properties[options.caption]` is `undefined`.  Not sure, but 'undefined' as a caption is not useful.

AliyanH: Yup good point, better than undefined for sure. Will include this.

src/geojson/geojson.js
```diff
@@ -0,0 +1,494 @@
...
+        }
+        curr_feature.querySelector("map-featurecaption").innerHTML = featureCaption;
+
+        // Setting Properties
+        let p;
+        // if properties function is passed
```
prushforth, line 148: Thinking about the properties function parameter: maybe it would be even more useful to pass the feature (including its properties, of course), so that the function could operate on the geometry and any other property of the feature.  e.g. calculate area, length and so on.

AliyanH: Yup agreed, would also make it the same as the caption function, so less confusion!

src/geojson/geojson.js
```diff
@@ -0,0 +1,494 @@
...
+function geometry2geojson(el, source, dest, transform) {
+    let elem = el.nodeName;
+    let j = {};
+    let coord;
+
+    switch(elem.toUpperCase()) {
+        case "MAP-POINT":
+            j.type = "Point";
+            if (transform) {
+                let pointConv = proj4.transform(source, dest, ((el.querySelector('map-coordinates').innerHTML.split(/[<>\ ]/g)).map(Number)) );
```
prushforth, line 341: I am getting a "proj4 is not defined" exception on this line when called like so:

```JavaScript
    let geojson = mapml2geojson(mapml, function(propertiesElement){
      return {"properties": propertiesElement.innerText};
    }, true);
```

AliyanH: Good catch, looks like mapml-viewer.js does not fully get loaded in before that code is run. putting it inside a windows.onload event works for me. Will need to look for a solution...

`window.onload = (event) => {
  ...
};`

src/geojson/geojson.js
```diff
@@ -0,0 +1,494 @@
...
+            curr_feature.querySelector('map-properties').insertAdjacentHTML("beforeend", options.properties);
+            p = false;
+        } else if (options.properties instanceof HTMLElement) { // if an HTMLElement is passed - NOT TESTED
+            p = options.properties;
+        } else { // If no properties function, string or HTMLElement is passed
+            p = properties2Table(json.properties);
+        }
+        
+        if (p) {
+            curr_feature.querySelector('map-properties').appendChild(p);
```
prushforth, line 165: If my function returns a string I get an error here.  I think it might be better silently ignore non-node returns.  WDYT?  Maybe it's good feedback to the user, not sure.

AliyanH: what do we set the properties to when a non-node is returned, I presume we leave it blank. I think it would be good feedback to the user, perhaps I can change the error it throws to guide the user? or perhaps a silent warning in the console, while leaving the properties blank?

test/e2e/geojson/geojson2mapml.test.js
```diff
@@ -0,0 +1,139 @@
...
+  test("Point Geometry (string json)", async () => {
+    const out = await page.$eval(
+      "body > mapml-viewer#output > layer-:nth-child(1)", 
+      (node) => node.outerHTML
+    );
+    const exp = await page.$eval(
+      "body > mapml-viewer#expected1 > layer-", 
+      (node) => node.outerHTML
+    );
+    expect(out).toEqual(exp);
```
prushforth, line 26: I like the simplicity of these tests.  Also, it is surprising, but cool, that expect(out).toEqual(exp) will do a deep-equals on the retrieved nodes.

test/e2e/geojson/geojson2mapml.html
```diff
@@ -0,0 +1,715 @@
...
+  <script src="lib/geojson.js"></script>
+</head>
+
+<body>
+    <mapml-viewer id="output" projection="OSMTILE" zoom="10" lon="-75.7" lat="45.4" controls>
+        <!--geojson2mapml layer output goes here dynamically-->
+    </mapml-viewer>
+
+    <!-- The expected results for geojson2mapml go here-->
+    <mapml-viewer id="expected1" projection="OSMTILE" zoom="10" lon="-75.7" lat="45.4" controls>
```
prushforth, line 17: It may be the case that the many mapml-viewers here aren't necessary, since the actual layers never get rendered, they're just used for comparison with the output of geojson2mapml, perhaps it would be faster to just put the id on the `<layer->` elements and compare them directly.  You can always put a `hidden` attribute on the layer to prevent it from being displayed with a FOUC, I believe.  Anyway, not a big deal because the tests work as is.

test/e2e/geojson/geojson2mapml.html
```diff
@@ -0,0 +1,715 @@
...
+    </mapml-viewer>
+    <mapml-viewer id="expected10" projection="OSMTILE" zoom="10" lon="-75.7" lat="45.4" controls>
+      <layer- label="Point 1" checked><map-meta name="extent" content="top-left-longitude=-75.916809, top-left-latitude=45.886964, bottom-right-longitude=-75.516809,bottom-right-latitude=45.26964"></map-meta><map-meta name="projection" content="OSMTILE"></map-meta><map-meta name="cs" content="gcrs"></map-meta><map-feature><map-featurecaption>caption test</map-featurecaption><map-geometry><map-point><map-coordinates>-75.6916809 45.4186964</map-coordinates></map-point></map-geometry><map-properties><h1>This is a Point - with a featurecaption function</h1></map-properties></map-feature></layer->
+    </mapml-viewer>
+    
+
+
+    <!--All the goejson that will be converted to <layer->-->
+      <script>
```
prushforth, line 51: A technique [I just learned about today](https://github.com/Maps4HTML/experiments/blob/a1f2ab9d2888f7f43feb66dc1d3786418dab1b3a/api/geojson/canada.js) (TIL), is that you can use `<script src="..."></script>` to load a bunch of geojson-valued objects into the global scope; these can be used in your inline script code and they can be re-used by other tests.

test/e2e/geojson/geojson2mapml.html
```diff
@@ -0,0 +1,715 @@
...
+          layer = geojson2mapml(multipolygon);
+          mapml.appendChild(layer);
+
+          layer = geojson2mapml(geometrycollection);
+          mapml.appendChild(layer);
+
+          layer = geojson2mapml(featurecollection);
+          mapml.appendChild(layer)
+
+          layer = geojson2mapml(bbox_label_properties_and_caption_string, {label: "Point", caption: "prop0", properties: "<h1>String Properties test</h1>"})
```
prushforth, line 700: Probably we need to have a few more options tests, but let's wait until we brainstorem about what other options might be useful for devs who use these functions.

test/e2e/geojson/mapml2geojson.test.js
```diff
@@ -0,0 +1,681 @@
...
+           "type": "Point",
+           "coordinates": [-75.6916809, 45.4186964]
+       },
+       "properties": {
+           "prop0": "This is a Point"
+       }
+   }]
+  };
+
+let line = {
```
prushforth, line 98: If you put these variables into a .js file that was loaded by the mapml2geojson.html via `<script src=""></script>`, you could serialize them with stringify into a div that represents the "input" json, then compare the output with that input / expected value.  That way you wouldn't have these variables in here.  Just a suggestion, not sure it will work.

test/e2e/geojson/mapml2geojson.html
```diff
@@ -0,0 +1,288 @@
...
+<head>
+  <title>MapML 2 geoJSON</title>
+  <meta charset="UTF-8">
+  <script type="module" src="mapml-viewer.js"></script>
+  <script src="lib/geojson.js"></script>
+</head>
+
+<body>
+  <!--All the mapml layers- that will be converted to geojson go in the mapml-viewer-->
+    <mapml-viewer projection="OSMTILE" zoom="10" lon="-75.7" lat="45.4" controls>
```
prushforth, line 13: I'm guessing that the mapml-viewer isn't necessary and that the tests might run a bit faster without it.

test/e2e/geojson/mapml2geojson.html
```diff
@@ -0,0 +1,288 @@
...
+      <p id="json2"></p>
+      <p id="json3"></p>
+      <p id="json4"></p>
+      <p id="json5"></p>
+      <p id="json6"></p>
+      <p id="json7"></p>
+
+      <script>
+        window.onload = function() {
+            let json;
```
prushforth, line 250: We need to test the options object and their possible values.

prushforth (commented):
I would be willing to merge this.  We need a few more tests in the options area, but in general you were very thorough with the testing of geometries.  Thanks!  Good first contribution on this repo! (at least, I think it's your first, I may be forgetting). It's up to you how to proceed; I know you're working on > 1 thing rn.

src/geojson/geojson.js
AliyanH: If properties function now returns a non HTMLElement, it produces a silent error in the console, and leaves the properties empty.

test/e2e/geojson/geojson2mapml.html
AliyanH: Good point, removed mapml-viewers.

prushforth:
Awesome work, thank you!

Issue #267, opened 2021-01-28
Text instructions for keyboard shortcuts and the query reticule


prushforth:

Malvoz:
By "text instructions" do you mean instructions for how to operate it (keyboard shortcuts)?

prushforth:
Yes, maybe it needs them just as the a11y map and Google Maps do when the focus area is visible?

Malvoz:
I've renamed this issue to include text instructions for keyboard shortcuts in general as well.

Since last year when we reported accessibility issues in Google Maps at the time of the W3C/OGC Maps for the Web workshop they've made a ton of improvements (related [blog post](https://cloud.google.com/blog/products/maps-platform/improved-accessibility-maps-javascript-api)), including a new "keyboard shortcuts" button placed next to the attribution links:

<img width="350" src="https://user-images.githubusercontent.com/26493779/143273122-20243cc9-5304-4103-8c02-3fb8150d8bac.png">

which opens up a dialog with text instructions:

<img width="500" src="https://user-images.githubusercontent.com/26493779/143273164-8f73c200-f412-4425-a471-c5537b434b9d.png">

A dialog like this could hold instructions for keyboard shortcuts such as pan, zoom, context menu, and the query reticule.

prushforth:
It's great, your advocacy has had an impact on Google Maps too, by the looks of it.  It would be **greater** if they all came over here and helped us improve our proposal.  Not that you can do anything about that, mind you. I guess I'm happy to try to implement accessibility improvements to the polyfill as we can, including this one.

Issue #149, opened 2020-08-21
Localize the UI


prushforth:
When the custom element is loaded in a browser, messages should be available in the locale of the browser.  We need to design a way to store and load localized messages, such as those used by the context menu, but more generally for all information that needs localization.

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/i18n/getUILanguage

Malvoz:
@prushforth Should we have a test page with right-to-left text directionality? I worry that [`dir="rtl"`](https://html.spec.whatwg.org/multipage/dom.html#the-dir-attribute) (currently not a defined content attribute for mapml, probably should be) on either `<html>` or `<mapml>` may cause issues with `<coordinates>`/`<geometry>`, haven't done any testing though.

prushforth:
@Malvoz 

> Should we have a test page with right-to-left text directionality just to check that everything works well including inline mapml. I think it could be as simple as setting [`dir="rtl"`](https://html.spec.whatwg.org/multipage/dom.html#the-dir-attribute) (currently not a defined content attribute for mapml, probably should be) on `<html>` and/or `<mapml>`

Maybe?   Ahmad's on his last few days on the project so I want to help him get his latest PRs merged.  After that, I will be trying to get ready for the workshop, so this may have to be a bookmark for a little while...

ahmadayubi:
Some resources that may be of value in the future:
- [PolyglotJS - by AirBnB](https://airbnb.io/polyglot.js/)
- [i18nJS - ](https://i18njs.com/)

I think the usual design pattern is to store translations in a seperate json file that is loaded in using a library like PolyglotJS, which manages the loaded content. In our case there are only a handful of translations needed so I'm wondering if having the additional features these libraries provide is worthwhile or if a native approach would be more efficient. Not sure.

Malvoz:
It should be noted that keyboards suited for other languages than English can have totally different layouts, so keyboard shortcuts may have to be adapted as well, for a fully localized UI.

Malvoz:
It should be noted that [CSS Logical Properties](https://www.w3.org/TR/css-logical/) enables "layout through logical, rather than physical, direction and dimension mappings", which would go a long way in setting up the UI for internationalization (more info/demo/examples here: https://web.dev/logical-property-shorthands/). I'd be _very_ surprised if Leaflet actually decided to support it in the future however...

Malvoz:
[...]

PR #683, opened 2022-08-10
Add attribution for keyboard shortcuts

ben-lu-uw:
Closes https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/267
Add an attribution that opens a dialog element to show keyboard shortcuts. 

Dialog element is not supported in Internet Explorer, not sure if this is a concern or not given that it has been retired.


prushforth:
> Dialog element is not supported in Internet Explorer, not sure if this is a concern or not given that it has been retired.

Nope, because custom elements aren't supported there either, so this doesn't matter for that antecedent reason.

prushforth:
When I load the dist branch into experiments, I get this:

image

ben-lu-uw:
If you have the extension install and open a page with the map-options hard coded into the html for sake of demo that will happen.

prushforth:
Is the tab stop at the features removed inadvertently with this link? When tab from feature index (focus on map) it lands not on the features but on the attribution control, I think.

ben-lu-uw:
The tab order should be the same as it was originally. Map -> features -> controls -> attributions.

prushforth:
> If you have the extension installed without the corresponding branch and open a page with the map-options hard coded into the html for sake of demo that will happen.

So I do have the extension installed, and when the extension has the corresponding branch installed, I get the english dialog (did you localize the dialog content?  I find [deepl](https://www.deepl.com/translator) to be a great resource for the programmer who does it all).  The [restaurants page](https://maps4html.org/experiments/screenreader/restaurants/) doesn't have the map-options hard-coded, so it means that that's what a user who doesn't have the extension installed well get.

> With feature indexing on, it was originally decided that the features would not be part of the tab order until arrow key feature navigation got implemented. 

I forgot that it's not implemented yet, awaiting review 🤦‍♂️ ! 😄

ben-lu-uw:
I misspoke about the undefined part. That will happen if you have an old version of the extension. If the extension is not installed it uses the defaults set up in options.js.

prushforth:
I sent a PR with french and swedish localization (hopefully @Malvoz will vet my swedish).  One issue I notice is when I set chrome to display in french, the "Keyboard Shortcuts" attribution link is not localized:

image
image

src/mapml/control/AttributionControl.js
```diff
@@ -0,0 +1,17 @@
...
+    map._attributionControl = map._map.attributionControl.setPrefix('<button onclick="this.closest(\'.leaflet-container\').querySelector(\'.shortcuts-dialog\').showModal()" class="shortcuts-button mapml-button">Keyboard shortcuts</button> | <a href="https://www.w3.org/community/maps4html/" title="W3C Maps for HTML Community Group">Maps4HTML</a> | <img src="data:image/svg+xml;base64,PHN2ZyBhcmlhLWhpZGRlbj0idHJ1ZSIgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIiB3aWR0aD0iMTIiIGhlaWdodD0iOCI+PHBhdGggZmlsbD0iIzRDN0JFMSIgZD0iTTAgMGgxMnY0SDB6Ii8+PHBhdGggZmlsbD0iI0ZGRDUwMCIgZD0iTTAgNGgxMnYzSDB6Ii8+PHBhdGggZmlsbD0iI0UwQkMwMCIgZD0iTTAgN2gxMnYxSDB6Ii8+PC9zdmc+" style="padding-right: 0.3em;" alt="Slava Ukraini"> <a href="https://leafletjs.com" title="A JS library for interactive maps">Leaflet</a>');
+
+    let dialog = document.createElement("dialog");
+    dialog.setAttribute("class", "shortcuts-dialog");
+    dialog.setAttribute("autofocus", "");
+    dialog.onclick = function (e) {
+        e.stopPropagation();
+    };
+    dialog.innerHTML = `<b>${M.options.locale.kbdShortcuts} </b><button onclick='this.parentElement.close()'>X</button>` +
```
Malvoz, line 10:
```suggestion
    dialog.innerHTML = `<b>${M.options.locale.kbdShortcuts} </b><button aria-label="Close" onclick='this.parentElement.close()'>X</button>` +
```

The close button is missing an accessible name. See [MDN: Interactive elements must be labeled](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Text_labels_and_names#interactive_elements_must_be_labeled) for more info.

(I suppose "Close" should be translated as well at some point, per https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/149.)

Issue #684, opened 2022-08-15
Menu items in the context menu shouldn't wrap


Malvoz:
As seen in https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/683#issuecomment-1213453371:

<img src="https://user-images.githubusercontent.com/2437285/184430827-d0cebed3-e28e-4e04-a413-41dd0f97cff4.png">

Ideally the menu's width would adjust to the content within.

prushforth:
This perhaps relates to, or at least should inform how the copy paste menus are created  in #698

PR #685, opened 2022-08-15
Context menu wrapping

prushforth:


prushforth:
Fixes #684

AliyanH:
The context menu's width seems to be a lot bigger than the actual content for me:

image

The coord menu overlaps the context menu, shouldn't the placement of the coord menu be changed also? I believe this can be done through the __showCoordMenu_ method.

Examples:

coord menu overlapping
coord menu overlapping 2

PR #705, opened 2022-12-11
Uglify everthing

prushforth:


AliyanH (approved):
Looks Great! This uglification helps reduce the dist directory's size by 60%. Like @prushforth mentioned in our discussions, this can be reduced further in the future using other minification techniques such as "tree-shaking".

Good performance improvement, helps progress https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/153

PR #715, opened 2023-01-13
Upgrade to leaflet.1.9.3

### LINKED ISSUE #658: Test and Implement Leaflet 1.9.x*
Currently, there are a few issues I see when running web-map-custom-element under Leaflet 1.8.0.beta 3

- [ ] get tests running
- [ ] provide feedback to Leaflet committers
- [ ] other - please comment

I tested the experiments with 1.8.0 beta 3, and it seems to work pretty well.  I noticed a couple of things:

- layer control context menu no longer available /  replaced by browser context menu
- after following a map-a link, the mouse pointer is stuck in the dragging state until released by another event
- unable to distinguish if tests are broken because test infrastructure isn't great or if the upgraded API causes breakage.  Only a few tests (16) were broken, out of ~300.

AliyanH:
closes #658

### DIFF
--- a/src/mapml/layers/DebugLayer.js
+++ b/src/mapml/layers/DebugLayer.js
@@ -238,6 +238,13 @@ export var debugVectors = function (options) {
 
 
 var ProjectedExtent = L.Path.extend({
+  
+  getCenter: function (round) {
+    let crs = this._map.options.crs;
+    return crs.unproject(L.bounds(this._locations).getCenter());
+  
+  },
+
   options: {
     className: "mapml-debug-extent",
   },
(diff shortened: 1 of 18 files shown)


AliyanH:
Discovered another issue: when debug mode is on on the map, focusing on the map (aka clicking on the map) results in an error which is causing errors to occur on layers\multipleExtents.test.js
image

prushforth:
I guess we should fix #720 and add it to this PR before squash and merge

src/mapml/layers/DebugLayer.js
```diff
@@ -1,7 +1,12 @@
 export var DebugOverlay = L.Layer.extend({
 
+  getCenter: function () {
+    return L.latLng(49, -95);
+  },
+
   onAdd: function (map) {
 
+
```
AliyanH, line 9: Don't think this is needed @kevinkim31?

PR #721, opened 2023-01-24
Disables toggle controls context menu item when map started with no controls 

### LINKED ISSUE #696: When a map is created with no controls, the Toggle Controls context menu item should be disabled
Or not, but it currently doesn't do anything when the `controls` attribute is omitted, despite being enabled.

Jacky0299:
Closes #696


AliyanH (commented):
Looks great, Good work @Jacky0299! Just a few suggestions to make the code a bit more simple.

PR #722, opened 2023-01-24
prevent default scrolling event of spacebar

### LINKED ISSUE #694: Page jumps when keyboard query activated
https://geogratis.gc.ca/mapml/en/cbmtile/fdi/

When the \<mapml-viewer> is a block element in a scrollable page, and the query reticle is displayed (because the map has focus), hitting spacebar (activating the query) causes the page to quickly and annoyingly scroll down then up.  Hitting Enter also activates the query, but not the annoying scroll jump.

This doesn't happen when the map is the whole page, for example when fullscreen is active.

yhy0217:
Closes #694

### DIFF
--- a/src/mapml-viewer.js
+++ b/src/mapml-viewer.js
@@ -355,7 +355,10 @@ export class MapViewer extends HTMLElement {
             (layer) => {
               M._pasteLayer(document.activeElement, layer);
             });
-      }});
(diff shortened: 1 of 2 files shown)


AliyanH:
@yhy0217 Creating a new PR.

PR #732, opened 2023-01-27
fix contextmenu wrapping and submenu displaying bugs

### LINKED ISSUE #684: Menu items in the context menu shouldn't wrap

yhy0217:
Closes #684

- [x] fix the context menu wrapping bug
- [x] stop the context menu and submenu from going outside from the map when clicking at margin
- [x] fix the bug that submenu keeps displaying when moving mouse out from 'Copy Coordinates (C)' button in context menu 
- [x] fix the submenu cutting off issue when the whole menu is opened near the bottom of the page
- [x] create a test in mapContextMenu.test.js

### DIFF
--- a/src/mapml/handlers/ContextMenu.js
+++ b/src/mapml/handlers/ContextMenu.js
@@ -442,8 +442,6 @@ export var ContextMenu = L.Handler.extend({
 
   _show: function (e) {
     if(this._mapMenuVisible) this._hide();
-    // the 'hidden' attribute must be removed before any attempt to get the size of container
-    else this._container.removeAttribute('hidden');
     this._clickEvent = e;
     let elem = e.originalEvent.target;
     if(elem.closest("fieldset")){
@@ -451,10 +449,13 @@ export var ContextMenu = L.Handler.extend({
       elem = (elem.className === "mapml-layer-extent") ? elem.closest("fieldset").parentNode.parentNode.parentNode.querySelector("span") : elem.querySelector("span");
       if(!elem.layer.validProjection) return;
       this._layerClicked = elem;
+      this._layerMenu.removeAttribute('hidden');
       this._showAtPoint(e.containerPoint, e, this._layerMenu);
     } else if(elem.classList.contains("leaflet-container") || elem.classList.contains("mapml-debug-extent") ||
-      elem.tagName === "path") {
+    elem.tagName === "path") {
       this._layerClicked = undefined;
+      // the 'hidden' attribute must be removed before any attempt to get the size of container
+      this._container.removeAttribute('hidden');
       this._showAtPoint(e.containerPoint, e, this._container);
     }
     if(e.originalEvent.button === 0 || e.originalEvent.button === -1){
--- a/src/mapml/handlers/ContextMenu.js
+++ b/src/mapml/handlers/ContextMenu.js
@@ -442,6 +442,8 @@ export var ContextMenu = L.Handler.extend({
 
   _show: function (e) {
     if(this._mapMenuVisible) this._hide();
+    // the 'hidden' attribute must be removed before any attempt to get the size of container
+    else this._container.removeAttribute('hidden');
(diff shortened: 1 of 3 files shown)


AliyanH:
@yhy0217 found a small bug, when I open the layer context menu, the map contextmenu opens also:

image

Also, noticed this when opening the context menu for the layer near the margin, if its a simple fix, are you able to replicate the same function for the layer contextmenu as you did for the map contextmenu:

image

If that's not as simple, you can create a new issue and we can handle it later.

yhy0217:
> @yhy0217 found a small bug, when I open the layer context menu, the map contextmenu opens also:
> 
> image
> 
> Also, noticed this when opening the context menu for the layer near the margin, if its a simple fix, are you able to replicate the same function for the layer contextmenu as you did for the map contextmenu:
> 
> image
> 
> If that's not as simple, you can create a new issue and we can handle it in a separate PR.

Finished to fix the bug that two menus open at the same time; however the layer menu does not wrap up on my laptop, and I think it should not happen because the new codes changes the style of all buttons inside both two menus (as they all have same classname ".mapml-contextmenu-item"), but I may wrong. Could you please run the code again and see if there is any change about the styling?

AliyanH:
> Finished to fix the bug that two menus open at the same time

Nice!

> Could you please run the code again and see if there is any change about the styling?

Yup, the wrapping issue was solved, I may not have had the latest CSS.

Good work, this is looking great, the contextmenu is appearing as it should. Just found one other thing when the contextmenu is opened near the bottom of the map (the coordinates menu cuts out). Other than that, its perfect, really good work on this!
image

yhy0217:
> Just found one other thing when the contextmenu is opened near the bottom of the map (the coordinates menu cuts out).

Yes, it used to look like this, but I fixed this problem in my first commit, and it now works well on my laptop:

Picture1

So I think a re-run may solve this issue too! :)

AliyanH:
> So I think a re-run may solve this issue too!

I believe this is different, this happens when the right click is initiated from the red spot shown below:
image

yhy0217:
> I believe this is different, this happens when the right click is initiated from the red spot shown below:

Yes you are right, sorry for my misunderstanding. Have updated the code to fix it

src/mapml/handlers/ContextMenu.js
```diff
@@ -646,17 +649,21 @@ export var ContextMenu = L.Handler.extend({
...
-      menu.style.right = 160 + 'px';
+      menu.style.right = menuWidth + 'px';
     } else {
-      menu.style.left = 160 + 'px';
+      menu.style.left = menuWidth + 'px';
       menu.style.right = 'auto';
     }
 
-    if (click.containerPoint.y + 160 > mapSize.y) {
+    if (click.containerPoint.y + menuHeight + 73 > mapSize.y) {
```
prushforth, line 663: I'm only left wondering what the 73 is for.  You replaced other 'magic numbers' with variables, which makes the code much more understandable.  Could we assign 73 to a variable and use that?

prushforth (changes requested):
Looks good to me!  Changing the magic numbers to variables would make the code more meaningful, which you have mostly done.

src/mapml/handlers/ContextMenu.js
yhy0217: > I'm only left wondering what the 73 is for. 

it is the height difference between the main context menu and the submenu.

> Could we assign 73 to a variable and use that?

Yes, it will make more sense to change it to a variable! I have pushed a new commit and I also add a comment for it.

AliyanH (approved):
LGTM!

prushforth (approved):

prushforth:
LGTM, please merge / rebase and merge when ready.

PR #733, opened 2023-01-30
Setting "Opacity" as an attribute in <layer-> tag #issue667new

### LINKED ISSUE #667: <layer- style="opacity: 0.5"> does not set initial opacity
According to the spec, the \<layer-> element should have a settable opacity, but it doesn't currently work.

> [layer](https://maps4html.org/MapML/spec/#the-layer-element) transparency is controlled by the CSS opacity property. [[SVG2](https://maps4html.org/MapML/spec/#bib-svg2)] [[css-color-3](https://maps4html.org/MapML/spec/#bib-css-color-3)]

Jacky0299:
closes #667


prushforth (commented):
Although layerAttribute is a pretty good name, unless you are planning to test the other attributes of the layer element in this same file, I would focus the name the test layerOpacityAttribute.test.js, to be clear.

test/e2e/layers/layerAttribute.test.js
```diff
@@ -0,0 +1,45 @@
...
+    page = context.pages().find((page) => page.url() === 'about:blank') || await context.newPage();
+    await page.goto("layerAttribute.html");
+  });
+  test.afterAll(async function () {
+    await context.close();
+  });
+
+  test("Setting Opacity Attibute to Layer- Element", async () => {
+    let opacity_attribute_value = await page.$eval("body > mapml-viewer > layer-",(layer) => layer.getAttribute("opacity"));
+    if (opacity_attribute_value == null){
```
prushforth, line 17: I believe you want to use "===" mostly in JavaScript.  I believe that you can re-phrase the logic of this as:

```javascript
if (opacity_attribute_value) {
...
}

File: test/e2e/layers/layerAttribute.test.js [hunk 2]
```diff
@@ -0,0 +1,45 @@
...
+        let opacity = await page.$eval("body > mapml-viewer > layer-", (layer) => layer.opacity);
+        // console.log(opacity);
+        return;
+    }
```
prushforth, line 21: I prefer to (try to) maintain one style of code block

Generally I prefer

```javascript
if (something) {
...
} else if (something_else) {
...
} else {
...
}

prushforth (changes requested):
Functions very well, AFAICT! Please make the style changes suggested!  We can merge after that.

I think we could add an opacity attribute to the \<map-extent> element now and you will be better able to understand what the impacts are, although we don't have a \<map-extent> custom element yet (soon!).

AliyanH:
Works very nicely! Since a new attribute has been added to the layer, you should also update the web-map-doc to document the change (https://maps4html.org/web-map-doc/docs/layers/layer/#attributes).

prushforth (approved):

AliyanH:
Looks good @Jacky0299! Feel free to "Squash and merge" when ready. Good work! You can also merge the [documentation](https://github.com/Maps4HTML/web-map-doc/pull/100), once this is merged.

PR #740, opened 2023-02-03
width and height attribute added #issue152

### LINKED ISSUE #152: Support the `width` and `height` attributes
It's currently not possible to change the initial width and height, using HTML attributes:
`<mapml-viewer width="500" height="500">`

Jacky0299:
resolves #152


test/e2e/web-map/mapAttributes.test.js
```diff
@@ -0,0 +1,37 @@
...
+  test.beforeAll(async function() {
+    context = await chromium.launchPersistentContext('');
+    page = context.pages().find((page) => page.url() === 'about:blank') || await context.newPage();
+    await page.goto("mapAttributes.html");
+  });
+  test.afterAll(async function () {
+    await context.close();
+  });
+
+  test("Setting Opacity Attibute to mapml-viewer", async () => {
```
prushforth, line 15: I think this is a cut and paste error

test/e2e/mapml-viewer/mapml-viewerAttributes.test.js
```diff
@@ -0,0 +1,37 @@
...
+  test.beforeAll(async function() {
+    context = await chromium.launchPersistentContext('');
+    page = context.pages().find((page) => page.url() === 'about:blank') || await context.newPage();
+    await page.goto("mapml-viewerAttributes.html");
+  });
+  test.afterAll(async function () {
+    await context.close();
+  });
+
+  test("Setting Opacity Attibute to mapml-viewer", async () => {
```
prushforth, line 15: Copy / pasta - title should be something to do with width and height attributes

test/e2e/mapml-viewer/mapml-viewerAttributes.test.js
```diff
@@ -0,0 +1,37 @@
...
+    let height_attribute_value = await page.$eval("body > mapml-viewer",(val) => val.getAttribute("height"));
+    let width_attribute_value = await page.$eval("body > mapml-viewer",(val) => val.getAttribute("width"));
+    let height;
+    let width;
+    if (!height_attribute_value && !width_attribute_value){
```
prushforth, line 20: we don't return from tests without testing our expectations!  That makes us think the test passed but it didn't actually test anything.

test/e2e/web-map/mapAttributes.test.js
```diff
@@ -0,0 +1,37 @@
...
+    let height_attribute_value = await page.$eval("body > map",(val) => val.getAttribute("height"));
+    let width_attribute_value = await page.$eval("body > map",(val) => val.getAttribute("width"));
+    let height;
+    let width;
+    if (!height_attribute_value && !width_attribute_value){
+        return;
```
prushforth, line 21: replace this return with expects for the value of height_attribute_value and width_attribute_value.  If they fail, there was something wrong with the initial conditions, and you'd want to know about that.

test/e2e/web-map/mapAttributes.test.js
```diff
@@ -0,0 +1,37 @@
...
+    } else if (height_attribute_value && !width_attribute_value){
```
prushforth, line 22: Why do you need the logic here?  shouldn't you just expect a value for height

test/e2e/mapml-viewer/mapml-viewerAttributes.html
```diff
@@ -0,0 +1,79 @@
...
+       * { 
+         margin: 0; 
+         padding: 0; 
+       } 
+        
+       /* Specifying the `:defined` selector is recommended to style the map 
+       element, such that styles don't apply when fallback content is in use 
+       (e.g. when scripting is disabled or when custom/built-in elements isn't 
+       supported in the browser). */ 
+       mapml-viewer:defined {
```
prushforth, line 22: For testing width and height attributes, I might leave the CSS selectors and rules for width and height out of the test, so you are sure about what your starting conditions are.

test/e2e/mapml-viewer/mapml-viewerAttributes.html
```diff
@@ -0,0 +1,79 @@
...
+       /* Specifying the `:defined` selector is recommended to style the map 
+       element, such that styles don't apply when fallback content is in use 
+       (e.g. when scripting is disabled or when custom/built-in elements isn't 
+       supported in the browser). */ 
+       mapml-viewer:defined { 
+         /* Responsive map. */ 
+          max-width: 100%;  
+          
+         /* Full viewport. */ 
+          width: 100%;
```
prushforth, line 27: Maybe remove these, at least for beginning test

prushforth (changes requested):

src/mapml-viewer.js
```diff
@@ -31,6 +31,18 @@ export class MapViewer extends HTMLElement {
     if (this.controlslist.includes(lowerVal) || !options.includes(lowerVal))return;
     this.setAttribute("controlslist", this.controlslist+` ${lowerVal}`);
   }
+  get width() {
+    return this.hasAttribute('width') ? this.getAttribute("width") : window.getComputedStyle(this).width;
```
AliyanH, line 35: The `window.getComputedStyle(this).width` returns width in the following format '100px' while `this.getAttribute("width")` returns '100', maybe you can add something like `.replace('px','')` to keep things consistent. Same thing for `get height()`.

AliyanH:
I just started looking at this and it is looking good so far @Jacky0299! Since you are changing the attributes again, the [web-map-doc](https://maps4html.org/web-map-doc/docs/elements/mapml-viewer/) will also need to be updated. You can make your documentation similar to the MDN documentation for the width and height attributes, for ex [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-width).

prushforth:
I interactively rebased the commits related to width and height.  Let's see if the tests pass, as they did just now on my machine.

test/e2e/mapml-viewer/mapml-viewerAttributes.test.js
```diff
@@ -0,0 +1,27 @@
...
+  test.beforeAll(async function() {
+    context = await chromium.launchPersistentContext('');
+    page = context.pages().find((page) => page.url() === 'about:blank') || await context.newPage();
+    await page.goto("mapml-viewerAttributes.html");
+  });
+  test.afterAll(async function () {
+    await context.close();
+  });
+
+  test("Setting Width and Height Attibute to mapml-viewer", async () => {
```
prushforth, line 15: Much better title, thank you

test/e2e/mapml-viewer/mapml-viewerAttributes.test.js
```diff
@@ -0,0 +1,27 @@
...
+    let height_attribute_value = await page.$eval("body > mapml-viewer",(val) => val.getAttribute("height"));
```
prushforth, line 16: local variable name 'val' should be something meaningful, e.g. 'viewer'

test/e2e/mapml-viewer/mapml-viewerAttributes.test.js
```diff
@@ -0,0 +1,27 @@
...
+    let width_attribute_value = await page.$eval("body > mapml-viewer",(val) => val.getAttribute("width"));
```
prushforth, line 17: same comment re: local variable name

test/e2e/mapml-viewer/mapml-viewerAttributes.test.js
```diff
@@ -0,0 +1,27 @@
...
+    let height = await page.$eval("body > mapml-viewer", (val) => val._container.style.height);
+    let width = await page.$eval("body > mapml-viewer", (val) => val._container.style.width);
+    expect(height).toEqual(height_attribute_value+"px");
```
prushforth, line 20: This is good.  You could go so far as to name the "width" (and height) variable like "width_with_px" or "[used_width](https://developer.mozilla.org/en-US/docs/Web/CSS/used_value)", with a comment link to used values.  Tests are all about what you mean, so being a bit more verbose, but especially explicit, never hurts.

test/e2e/web-map/mapAttributes.html
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE html>
```
prushforth, line 1: The name of this file is a little too generic, unless you are testing all the attributes of the map.  I would probably call it "mapWidthAndHeightAttributes.html" fwiw.

test/e2e/web-map/mapAttributes.test.js
```diff
@@ -0,0 +1,27 @@
+import { test, expect, chromium } from '@playwright/test';
```
prushforth, line 1: Same comment as for the html file: name of mapAttributes is too generic, since you are only testing width and height, maybe make the file name mapWidthAndHeightAttributest.test.js

test/e2e/web-map/mapAttributes.test.js
```diff
@@ -0,0 +1,27 @@
...
+  });
+  test.afterAll(async function () {
+    await context.close();
+  });
+
+  test("Setting Width and Height Attibute to mapml-viewer", async () => {
+    let height_attribute_value = await page.$eval("body > map",(val) => val.getAttribute("height"));
+    let width_attribute_value = await page.$eval("body > map",(val) => val.getAttribute("width"));
+    let height = await page.$eval("body > map", (val) => val._container.style.height);
+    let width = await page.$eval("body > map", (val) => val._container.style.width);
```
prushforth, line 19: See previous comment, but this is better than before. Cheers

prushforth (approved):
You can change it if you are not sick of changing things!

prushforth:
If you want to update this PR, you'll have to fetch the branch from your origin, and then reset your working copy to that state, because I had to force push to your origin branch


git checkout issue152
git fetch origin
git reset --hard origin/issue152

should do it, **I think**.  Then your local issue152 branch will be in synch with your origin issue152 branch.

prushforth:
You should delete the old files (that are now duplicates with different names: mapml-viewerAttributes.html, mapml-viewerAttributes.test.js, mapAttributest.html, mapAttributest.test.js)

src/mapml-viewer.js
```diff
@@ -159,26 +173,25 @@ export class MapViewer extends HTMLElement {
...
+        h = this.hasAttribute("height") ? this.getAttribute("height") : parseInt(hpx.replace('px',''));
+      // if (!this.width || this.width !== w) {
+      //   this._container.style.width = wpx;
+      // } else {
+      //   this._container.style.width = this.width+"px";
+      // }
+   
+      this._changeWidth(w);
+       
+
```
AliyanH, line 186: You should remove all commented code so that it does not cause confusion and try to remove unneeded new lines.

AliyanH: Also this portion of the code and some other sections seems to be missing from web-map.js, make sure you are making the same changes to both

test/e2e/mapml-viewer/cssDomination.test.js
```diff
@@ -0,0 +1,31 @@
...
+
+  test("Css Values Dominate Attribute Values", async () => {
+    //height and width attribute value in the mapml-viewer tag
+    let height_attribute_value = await page.$eval("body > mapml-viewer",(viewer) => viewer.getAttribute("height"));
+    let width_attribute_value = await page.$eval("body > mapml-viewer",(viewer) => viewer.getAttribute("width"));
+    //actual height and width value of the map
+    let map_height = await page.$eval("body > mapml-viewer", (viewer) => viewer.height);
+    let map_width = await page.$eval("body > mapml-viewer", (viewer) => viewer.width);
+    expect(height_attribute_value).not.toEqual(map_height);
+    expect(width_attribute_value).not.toEqual(map_width);
```
AliyanH, line 23: good work testing this! You can take this even further and add more detail by checking to make sure the height and width attributes are equal to "600" just to test that the CSS does not change the values of width and height attributes.

test/e2e/mapml-viewer/mapml-viewerHeightAndWidthAttributes.test.js
```diff
@@ -0,0 +1,67 @@
...
+  test.beforeAll(async function() {
+    context = await chromium.launchPersistentContext('');
+    page = context.pages().find((page) => page.url() === 'about:blank') || await context.newPage();
+    await page.goto("mapml-viewerHeightAndWidthAttributes.html");
+  });
+  test.afterAll(async function () {
+    await context.close();
+  });
+
+  test("Setting Width and Height Attibute to mapml-viewer", async () => {
```
AliyanH, line 15: I think this test is not setting width and height attributes, maybe you can rename this and add another test that can test if setting width and height attribute to mapml-viewer works - same thing for web-map.js

AliyanH (commented):
Good work with the thorough testing, its testing all the important things! Your code and test for web-map and mapml-viewer is not in sync, you should make the same changes to both. Good work on this, seems to be working well for mapml-viewer!

src/web-map.js
```diff
@@ -555,15 +542,15 @@ export class WebMap extends HTMLMapElement {
   }
   
   _changeWidth(width) {
-    this.style.width = width+"px";
     this._container.style.width = width+"px";
+    document.querySelector('[is="web-map"]').style.width = width+"px";
```
prushforth, line 546: This only selects the 1st map is=web-map in the document, which isn't what you want here, I don't believe.

PR #741, opened 2023-02-03
Create <map-caption> custom element

### LINKED ISSUE #736: Web map title or caption
### Discussed in https://github.com/Maps4HTML/Web-Map-Custom-Element/discussions/558

<div type='discussions-op-text'>

<sup>Originally posted by **prushforth** November  2, 2021</sup>
When we listen to what a screen reader says about a map, I hear the page title and then "Application", "Interactive map", "Region", but I don't hear what the title of the map is.  This seems to me to be a shortcoming, I think each map should have a title (optionally) that is announced, and potentially visible somewhere, based on author  discretion.  

WDYT? </div>

kevinkim31:
Closes #736

### DIFF
--- a/src/map-caption.js
+++ b/src/map-caption.js
@@ -1,0 +1,44 @@
+import './leaflet.js'; 
+import './mapml.js'; 
+
+/* 
+implemented for both mapml-viewer and web-map; however web-map does not focus on map element in the browser resulting in NVDA 
+not being able to read out map-caption and stating that it's an interactive map region
+*/
+export class MapCaption extends HTMLElement {
+    constructor() {
+        super();
+        // calls MutationObserver; needed to observe changes to content between <map-caption> tags and update to aria-label
+        this.observer = new MutationObserver(() => {
+            this.parentElement.setAttribute('aria-label', this.textContent); 
+        });
+
+        this.map = this.parentElement;  
+    }
+    // function to retrieve caption content 
+    get ariaLabel() {
+        return this.textContent;
+    }
+    // function to allow changing of caption content programmatically (changes both aria-label and map-caption content)
+    set ariaLabel (val) {
+        this.innerHTML = val;
+        this.parentElement.setAttribute('aria-label', val);
+    }
+    // called when element is inserted into DOM (setup code)
+    connectedCallback() {
+        this.observer.observe(this, {
+            characterData: true,
+            subtree: true,
+            attributes: true,
+            childList: true
+        });
+        const ariaLabel = this.textContent;
+        this.parentElement.setAttribute('aria-label', ariaLabel);
+    }
+    disconnectedCallback() {
+        this.observer.disconnect();
+
+        // removes aria-label when map-caption is removed
+        this.map.removeAttribute('aria-label');
+    }
+}
--- a/src/map-caption.js
+++ b/src/map-caption.js
@@ -14,14 +14,14 @@ export class MapCaption extends HTMLElement {
     connectedCallback() {
 
         // calls MutationObserver; needed to observe changes to content between <map-caption> tags and update to aria-label
-        let mapcaption = document.querySelector('map-caption').innerHTML;
+        let mapcaption = this.parentElement.querySelector('map-caption').innerHTML;
 
         this.observer = new MutationObserver(() => {
-            let mapcaptionupdate = document.querySelector('map-caption').innerHTML;
+            let mapcaptionupdate = this.parentElement.querySelector('map-caption').innerHTML;
 
             if (mapcaptionupdate != mapcaption)
             {
-                this.parentElement.setAttribute('aria-label', document.querySelector('map-caption').textContent);
+                this.parentElement.setAttribute('aria-label', this.parentElement.querySelector('map-caption').textContent);
             }
         });
 
--- a/src/map-caption.js
+++ b/src/map-caption.js
@@ -8,13 +8,6 @@ not being able to read out map-caption and stating that it's an interactive map 
 export class MapCaption extends HTMLElement {
     constructor() {
         super();
-        // calls MutationObserver; needed to observe changes to content between <map-caption> tags and update to aria-label
-        this.observer = new MutationObserver(() => {
-            if (document.querySelector('map-caption').innerHTML == this.parentElement.getAttribute('aria-label'))
-            {
-                this.parentElement.setAttribute('aria-label', this.textContent); 
-            }
-        });
     }
     // function to retrieve caption content 
     get ariaLabel() {
@@ -23,6 +16,19 @@ export class MapCaption extends HTMLElement {
 
     // called when element is inserted into DOM (setup code)
     connectedCallback() {
+
+        // calls MutationObserver; needed to observe changes to content between <map-caption> tags and update to aria-label
+        let mapcaption = document.querySelector('map-caption').innerHTML;
+
+        this.observer = new MutationObserver(() => {
+            let mapcaptionupdate = document.querySelector('map-caption').innerHTML;
+
+            if (mapcaptionupdate != mapcaption)
+            {
+                this.parentElement.setAttribute('aria-label', document.querySelector('map-caption').textContent);
+            }
+        });
+
         this.observer.observe(this, {
             characterData: true,
             subtree: true,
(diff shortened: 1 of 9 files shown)


prushforth:
I think that the generated aria-label should be removed if the \<map-caption> is disconnected / deleted?  WDYT?

AliyanH:
@kevinkim31 since a new element is being added, we should also update the [web-map-doc](https://maps4html.org/web-map-doc/docs/elements/mapml-viewer/) to document the changes you are making. I am not sure if we should add a new page for map-caption, but maybe we can add it under a `child element` heading for the mapml-viewer page similar to whats being done [here](https://maps4html.org/web-map-doc/docs/elements/feature/#child-elements)? WDYT?

AliyanH:
This is working great, just had 2 comments:

- I think we should probably have a test for testing the removal of map-caption now that it is functioning properly.
- When you have multiple `<map-caption>`'s I think the most recently added map-caption is used, which seems logical, but once that most recent map-caption is removed and there are other map-captions that are present, the mapml-viewer does not incorporate them, which does not seem right. I would assume it should used the second most recent one?

kevinkim31:
> This is working great, just had 2 comments:
> 
> * I think we should probably have a test for testing the removal of map-caption now that it is functioning properly.
> * When you have multiple `<map-caption>`'s I think the most recently added map-caption is used, which seems logical, but once that most recent map-caption is removed and there are other map-captions that are present, the mapml-viewer does not incorporate them, which does not seem right. I would assume it should used the second most recent one?

src/map-caption.js
```diff
@@ -0,0 +1,43 @@
...
+        // calls MutationObserver; needed to observe changes to content between <map-caption> tags and update to aria-label
+        this.observer = new MutationObserver(() => {
+            if (document.querySelector('map-caption').innerHTML == this.parentElement.getAttribute('aria-label'))
+            {
+                this.parentElement.setAttribute('aria-label', this.textContent); 
+            }
+        });
+    }
+    // function to retrieve caption content 
+    get ariaLabel() {
```
AliyanH, line 20: I am not sure if we need a getter for map-caption as people can just use $map-caption.innerHTML, just like regular HTML. I am sure we will need getters and setters in the future though as map-caption gets more complex!

src/map-caption.js
```diff
@@ -0,0 +1,43 @@
...
+/* 
+implemented for both mapml-viewer and web-map; however web-map does not focus on map element in the browser resulting in NVDA 
+not being able to read out map-caption and stating that it's an interactive map region
+*/
+export class MapCaption extends HTMLElement {
+    constructor() {
+        super();
+        // calls MutationObserver; needed to observe changes to content between <map-caption> tags and update to aria-label
+        this.observer = new MutationObserver(() => {
+            if (document.querySelector('map-caption').innerHTML == this.parentElement.getAttribute('aria-label'))
```
AliyanH, line 13: Should use === instead of ==
```suggestion
            if (document.querySelector('map-caption').innerHTML === this.parentElement.getAttribute('aria-label'))
```

src/map-caption.js
AliyanH: Also noticed that changing the innerHTML for map-caption does not modify the aria-label? I believe this was working in your previous commits? Ex. 
`$map-caption.innerHTML = "test"` changes the map-caption's innerHTML but does not modify the aria-label on mapml-viewer.

src/mapml-viewer.js
```diff
@@ -234,6 +235,37 @@ export class MapViewer extends HTMLElement {
         this.dispatchEvent(new CustomEvent('createmap'));
       }
     }
+
+    /*
+      1. only deletes aria-label when the last (only remaining) map caption is removed
+      2. only deletes aria-label if the aria-label was defined by the map caption element itself
```
AliyanH, line 241: Will need to rethink how to handle multiple `map-captions` as per our discussion from this morning:
* Only use the first map-caption?
*  When multiple map-captions are added, append them to the aria-label - similar to what's done with fig-captions?

AliyanH (commented):
some comments

kevinkim31:
Fixes:
- Only allows innerHTML modifications to the first <map-caption>
- Doesn't delete aria-label when one was initially set up by user; even if separate <map-caption> elements are created
- Only the first <map-caption>can modify and remove the aria-label except for if aria-label is created manually (mentioned in previous point)

src/map-caption.js
```diff
@@ -0,0 +1,45 @@
+import './leaflet.js';
```
prushforth, line 1: Lines 1 & 2 aren't necessary, because you don't use any symbols from those imports.

src/map-caption.js
```diff
@@ -0,0 +1,45 @@
...
+        this.observer = new MutationObserver(() => {
+            let mapcaptionupdate = this.parentElement.querySelector('map-caption').innerHTML;
+
+            if (mapcaptionupdate != mapcaption)
+            {
+                this.parentElement.setAttribute('aria-label', this.parentElement.querySelector('map-caption').textContent);
+            }
+        });
+
+        this.observer.observe(this, {
```
prushforth, line 28: Maybe observing too much, perhaps only [subtree](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/observe#parameters) is necessary?

src/web-map.js
```diff
@@ -275,6 +276,36 @@ export class WebMap extends HTMLMapElement {
...
+
+    /*
+      1. only deletes aria-label when the last (only remaining) map caption is removed
+      2. only deletes aria-label if the aria-label was defined by the map caption element itself
+    */
+
+    let mapcaption = this.querySelector('map-caption');
+    
+    if (mapcaption != null){
+      setTimeout(() => {
```
prushforth, line 288: Why is it required to put this on the event queue?

src/web-map.js
```diff
@@ -275,6 +276,36 @@ export class WebMap extends HTMLMapElement {
...
+        let ariaupdate = this.getAttribute('aria-label');
+  
+        if (ariaupdate == mapcaption.innerHTML) {
```
prushforth, line 291: Should be [strict equality](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality)

src/web-map.js
```diff
@@ -275,6 +276,36 @@ export class WebMap extends HTMLMapElement {
...
+          this.mapCaptionObserver = new MutationObserver((m) => {
+            let mapcaptionupdate = this.querySelector('map-caption');
+            if (mapcaptionupdate != mapcaption)
```
prushforth, line 294: Should be [Strict inequality](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_inequality)

src/web-map.js
```diff
@@ -275,6 +276,36 @@ export class WebMap extends HTMLMapElement {
...
+            {
+              this.removeAttribute('aria-label');
+            }
+            
+          });
+          this.mapCaptionObserver.observe(this, {
+            characterData: true,
+            subtree: true,
+            attributes: true,
```
prushforth, line 303: Probably only need `attributes: true` here, no?  This is a big element with lots of subtree changes, could be a large performance burden due to all this observing going on.

prushforth (changes requested):
Tests look good, I suggest a few updates, if possible. Thanks!

src/map-caption.js
```diff
@@ -0,0 +1,45 @@
...
+export class MapCaption extends HTMLElement {
+    constructor() {
+        super();
+    }
+
+    // called when element is inserted into DOM (setup code)
+    connectedCallback() {
+
+        // calls MutationObserver; needed to observe changes to content between <map-caption> tags and update to aria-label
+        let mapcaption = this.parentElement.querySelector('map-caption').innerHTML;
```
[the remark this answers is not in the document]
kevinkim31: For mapcaption, I need this to return the first map-caption only so that I can store it to then compare it to the mapcaption that was removed/modified.

src/web-map.js
kevinkim31: This is so that the ariaupdate has time to load

prushforth (changes requested):
A couple of small things. A test for when you add \<map-caption> inside e.g. \<layer-> to show it doesn't add aria-label to that element would be good.

prushforth (approved):
LGTM

PR #742, opened 2023-02-05
Documentation refactoring

prushforth:
A lot of renaming to allow less cognitive load when reading code, hopefully.


prushforth:
This is a pretty wide-ranging change, hopefully will make the code a little more readable.  I doubt it fixes the flaky tests, actually, but seems to be trending in the right direction.  I think everyone is going to need to do a `git pull --rebase upstream main` on their branches after this, to be on the safe side i.e. detect conflicts, in particular, because of the number of files changed.  If the flaky tests aren't solved, we can deal with them as a second npm test command.

PR #750, opened 2023-02-09
Focus layer contextmenu when opened with Shift+F10

### LINKED ISSUE #747: layer contextmenu does not get focused when opened with Shift+F10
Layer contextmenu does not function as it used to. Tabbing to the layer and opening the layer contextmenu using shift+F10 does not move focus to the layer contextmenu. Might have been caused after the leaflet 1.9.3 upgrade. Need to fix and add tests.

After [Leaflet 1.9.3 upgrade](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/715), opening the layercontexmenu would open the browser's default contextmenu, this [commit](https://github.com/Maps4HTML/Web-Map-Custom-Element/commit/b1e260d655c8e30c451c8d0e54546876722e5d5a) tried to fix the changes but in doing so we lost the ability to focus on the layer contextmenu.

### LINKED ISSUE #752: Can not check/uncheck layers using keyboard navigation
For ex, tabbing to layer control [here](https://maps4html.org/experiments/screenreader/restaurants/) and checking/unchecking a layer with keyboard navigation (by pressing space bar or enter on the checkbox) does not function as intended. Bug appeared after [leaflet 1.9.3 upgrade](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/715).

Similar behavior can be seen on the [leaflet site](https://leafletjs.com/examples/layers-control/) where enter does not work when checkbox is focused, but spacebar works as intended.

AliyanH:
Fixes bug caused by leaflet1.9.3 upgrade
- [X] Opening layer contextmenu by pressing enter focuses the first element
- [X] Opening the contextmenu on a layer using Shift+F10 focuses on the layer contextmenu
- [x] Can not check/uncheck layers using keyboard navigation, resolves #752
- [x] Add Tests

resolves #747.


AliyanH:
Another bug found tabbing to layer contextmenu [here](https://maps4html.org/experiments/screenreader/restaurants/) and pressing enter on the checkbox does not uncheck it.

prushforth:
Just having a look at this now.  The _changeExtent event handler function is responsible for handling the change even on this input.  The event handler is / should be created in the MapMLLayer.getLayerExtentHTML(labelName, i) function call, which is run after the fetch / xhr has happened.  It may be that there is a timing issue going on, or it could be a Leaflet upgrade issue.  Not sure ATM, will investigate further.

src/mapml/control/LayerControl.js
```diff
@@ -25,7 +25,10 @@ export var LayerControl = L.Control.Layers.extend({
         this._initLayout();
         this._map.on('validate', this._validateInput, this);
         L.DomEvent.on(this.options.mapEl, "layerchange", this._validateInput, this);
-        L.DomEvent.on(this._container, 'keydown', this._focusFirstLayer, this._container);
+        // Adding event on layer control button
+        L.DomEvent.on(this._container.getElementsByTagName("a")[0], 'keydown', this._focusFirstLayer, this._container);
```
prushforth, line 29: This is in the onAdd(map) method, and I  think we had to change the element that has the keydown event handler to prevent fighting with leaflet over it?  I am wondering what happens when there is no \<a> element in the layer control _container (will this even come up??? We don't have a layer control when there are no layers, right?)  I find it weird that both the focusFirst AND the spacebar-to-toggle /trigger input change event both went missing. It's got to be something about keydown on the _container that's no longer working... investigating.

src/mapml/control/LayerControl.js
```diff
@@ -104,10 +108,28 @@ export var LayerControl = L.Control.Layers.extend({
 
     // focus the first layer in the layer control when enter is pressed
     _focusFirstLayer: function(e){
-      if(e.key === 'Enter' && this.className != 'leaflet-control-layers leaflet-control leaflet-control-layers-expanded'){
```
prushforth, line 107: The event was on _container before, but I can't understand why the != here didn't prevent _focusFirstLayer from having the desired effect.

src/mapml/control/LayerControl.js
```diff
@@ -104,10 +108,28 @@ export var LayerControl = L.Control.Layers.extend({
...
+      if(e.key === 'Enter' && this.className === 'leaflet-control-layers leaflet-control leaflet-control-layers-expanded'){
         var elem = this.children[1].children[2].children[0].children[0].children[0].children[0];
```
prushforth, line 112: This looks kinda fragile tbh. Best leave it for now, I guess.

test/e2e/mapml-viewer/viewerContextMenu.test.js
```diff
@@ -426,4 +426,54 @@ test.describe("Playwright mapml-viewer Context Menu (and api) Tests", () => {
...
+    const aHandle = await page.evaluateHandle(() => document.querySelector("mapml-viewer"));
+    const nextHandle = await page.evaluateHandle(doc => doc.shadowRoot, aHandle);
+    const resultHandle = await page.evaluateHandle(root => root.activeElement, nextHandle);
+    const nameHandle = await page.evaluateHandle(name => name.nodeName, resultHandle);
+    let name = await nameHandle.jsonValue();
+    await nameHandle.dispose();
+    expect(name).toEqual("INPUT");
+  });
+
+  test("Layer Context menu, Pressing enter on checkbox button toggles layer", async () => {
```
prushforth, line 468: spacebar

test/e2e/core/mapContextMenu.test.js
```diff
@@ -485,4 +485,53 @@ test.describe("Playwright Map Context Menu Tests", () => {
...
+    const aHandle = await page.evaluateHandle(() => document.querySelector(".mapml-web-map"));
+    const nextHandle = await page.evaluateHandle(doc => doc.shadowRoot, aHandle);
+    const resultHandle = await page.evaluateHandle(root => root.activeElement, nextHandle);
+    const nameHandle = await page.evaluateHandle(name => name.nodeName, resultHandle);
+    let name = await nameHandle.jsonValue();
+    await nameHandle.dispose();
+    expect(name).toEqual("INPUT");
+  });
+  
+  test("Layer Context menu, Pressing enter on checkbox button toggles layer", async () => {
```
prushforth, line 527: spacebar

prushforth:
[This is the commit](https://github.com/Maps4HTML/Web-Map-Custom-Element/commit/249b22b68fe3471308a5be2128260a6a671881f6) that caused the spacebar to stop working, but it may have something to do with the code block starting at line 361 of mapml-viewer.js in that commit (which handles "space" key 32). Same considerations likely apply to web-map.js. I'm pretty sure that was my fault; changes were made to remove parentElement without due consideration.  I'm not sure if it is why the checkbox no longer focused automatically when the layer control is opened.  I went back quite far in history and I found that the behaviour of Enter to open vs the layer control just opening when focused has changed a few times.

AliyanH:
> [This is the commit](https://github.com/Maps4HTML/Web-Map-Custom-Element/commit/249b22b68fe3471308a5be2128260a6a671881f6) that caused the spacebar to stop working, but it may have something to do with the code block starting at line 361 of mapml-viewer.js in that commit (which handles "space" key 32). Same considerations likely apply to web-map.js. I'm pretty sure that was my fault; changes were made to remove parentElement without due consideration. I'm not sure if it is why the checkbox no longer focused automatically when the layer control is opened. I went back quite far in history and I found that the behaviour of Enter to open vs the layer control just opening when focused has changed a few times.

Thanks @prushforth, good catch! I am not sure if the parentElement was the issue. That commit was handling the space key for all of mapml-viewer and preventing the default behavior for the space key. I have added an exception to input elements as of now.

PR #753, opened 2023-02-15
static attribute added to mapml-viewer

### LINKED ISSUE #749: Static maps
Relates to [use case](https://maps4html.org/HTML-Map-Element-UseCases-Requirements/#use-case-static-map): https://github.com/Maps4HTML/HTML-Map-Element-UseCases-Requirements/issues/17

https://github.com/Maps4HTML/HTML-Map-Element-UseCases-Requirements/issues/17#issuecomment-525039770

I think we should implement this as a boolean `<mapml-viewer static`> attribute, but your thoughts below would be appreciated.  Because it's static, a default behaviour would be to remove the pan/zoom/reload controls, and disable the forward/back/reload context items. If you remove the `static` attribute, the controls would appear and the map would become pannable/zoomable.  

A key goal of this use case is to maintain an accessible map image for users, but have it be georeferenced via the usual means (`<mapml-viewer lat=... lon=... zoom=...>`), so that it can be crawled and spatially indexed.

Jacky0299:
new attribute!

closes #749.

### DIFF
--- a/src/mapml-viewer.js
+++ b/src/mapml-viewer.js
@@ -244,7 +244,7 @@ export class MapViewer extends HTMLElement {
         this.dispatchEvent(new CustomEvent('createmap'));
       }
 
-      if(this._map&&this.hasAttribute('static')){
+      if (this._map && this.hasAttribute('static')) {
         this._toggleStatic();
       }
 
--- a/src/mapml-viewer.js
+++ b/src/mapml-viewer.js
@@ -365,34 +365,31 @@ export class MapViewer extends HTMLElement {
       }
       break;
       case 'static':
-      const isStatic = this.hasAttribute('static');
-      if (this._map) {
-        if (isStatic) {
-          this._toggleStatic();
-        } else {
-          this._toggleNonStatic();
-        }
-      }
+      this._toggleStatic();
       break;   
     }
   }
   _toggleStatic(){
-    this._map.dragging.disable();
-    this._map.touchZoom.disable();
-    this._map.doubleClickZoom.disable();
-    this._map.scrollWheelZoom.disable();
-    this._map.boxZoom.disable();
-    this._map.keyboard.disable();
-    this._zoomControl.disable();
-  }
-  _toggleNonStatic(){
-    this._map.dragging.enable();
-    this._map.touchZoom.enable();
-    this._map.doubleClickZoom.enable();
-    this._map.scrollWheelZoom.enable();
-    this._map.boxZoom.enable();
-    this._map.keyboard.enable();
-    this._zoomControl.enable();
+    const isStatic = this.hasAttribute('static');
+    if (this._map) {
+      if (isStatic) {
+        this._map.dragging.disable();
+        this._map.touchZoom.disable();
+        this._map.doubleClickZoom.disable();
+        this._map.scrollWheelZoom.disable();
+        this._map.boxZoom.disable();
+        this._map.keyboard.disable();
+        this._zoomControl.disable();
+      } else {
+        this._map.dragging.enable();
+        this._map.touchZoom.enable();
+        this._map.doubleClickZoom.enable();
+        this._map.scrollWheelZoom.enable();
+        this._map.boxZoom.enable();
+        this._map.keyboard.enable();
+        this._zoomControl.enable();
+      }
+    }
   }
   _dropHandler(event) {
     event.preventDefault();
--- a/src/mapml-viewer.js
+++ b/src/mapml-viewer.js
@@ -5,7 +5,7 @@ import { MapCaption } from './map-caption.js';
 
 export class MapViewer extends HTMLElement {
   static get observedAttributes() {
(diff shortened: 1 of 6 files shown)


AliyanH:
Looks pretty good so far @Jacky0299! Seems to function nicely for all devices. I think disabling the zoom buttons was the right approach as users can always use the [controlslist](https://maps4html.org/web-map-doc/docs/elements/mapml-viewer/#controlslist) attribute to remove them if they wanted.
The map is still not fully 'static', for example clicking on a feature at the corner of the map, still pans the map to make the popup visible, but I don't thing this is a big issue:
Animation1

src/mapml-viewer.js
```diff
@@ -342,26 +342,32 @@ export class MapViewer extends HTMLElement {
...
+  }
+  _toggleNonStatic(){
+    this._map.dragging.enable();
+    this._map.touchZoom.enable();
+    this._map.doubleClickZoom.enable();
+    this._map.scrollWheelZoom.enable();
+    this._map.boxZoom.enable();
+    this._map.keyboard.enable();
+    this._zoomControl.enable();
+  }
```
AliyanH, line 370: Can probably merge these two methods into one using a parameter, ex. _toggleStatic(true) or _toggleStatic(false)

src/mapml-viewer.js
```diff
@@ -233,7 +243,10 @@ export class MapViewer extends HTMLElement {
       if(!custom){	
         this.dispatchEvent(new CustomEvent('createmap'));
       }
-
+      if(this._map&&this.hasAttribute('static')){
```
AliyanH, line 246:
```suggestion

      if(this._map && this.hasAttribute('static')){
```

src/mapml-viewer.js
```diff
@@ -350,7 +363,32 @@ export class MapViewer extends HTMLElement {
       if (oldValue !== newValue) {
         this._changeWidth(newValue);
       }
-      break;  
+      break;
+      case 'static':
+      this._toggleStatic();
+      break;
```
AliyanH, line 369:
```suggestion
      case 'static':
        this._toggleStatic();
      break;   
```

test/e2e/mapml-viewer/staticAttribute.html
```diff
@@ -0,0 +1,78 @@
...
+         mapml-viewer:not(:defined) {
+           display: initial;
+           contain: initial;
+           contain-intrinsic-size: initial;
+         }
+       </style> 
+     </noscript> 
+  </head>
+  <body>
+    <mapml-viewer projection="CBMTILE" zoom="2" lat="45" lon="-90" controls static>
```
AliyanH, line 74: In your tests you are testing to add static attribute to the map, so you should not have static attribute from the start.
```suggestion
    <mapml-viewer projection="CBMTILE" zoom="2" lat="45" lon="-90" controls>
```

test/e2e/mapml-viewer/staticAttribute.test.js
```diff
@@ -0,0 +1,59 @@
...
+    page = context.pages().find((page) => page.url() === 'about:blank') || await context.newPage();
+    await page.goto("staticAttribute.html");
+  });
+  test.afterAll(async function () {
+    await context.close();
+  });
+
+  test("Setting New Static Attribute to mapml-viewer", async () => {
+    //setting static attribute in the mapml-viewer tag
+    await page.$eval("body > mapml-viewer", (viewer) => viewer.setAttribute("static",""));
```
AliyanH, line 17: instead of setAttribute you can try using the setter here to test your static API also. When you use the API to set the static attribute you can also add an `expect(_____.hasAttribute("static")).toEqual(true)` or something of that nature to test to see if the attribute gets set.

test/e2e/mapml-viewer/staticAttribute.test.js
```diff
@@ -0,0 +1,59 @@
...
+    expect(scrollWheelZoom).toEqual(false);
+    expect(boxZoom).toEqual(false);
+    expect(keyboard).toEqual(false);
+    expect(zoomControl).toEqual(true);
+
+  });
+  
+  test("Removing Static Attribute", async () => {
+    //removing static attribute in the mapml-viewer tag
+    await page.$eval("body > mapml-viewer", (viewer) => viewer.removeAttribute("static"));
```
AliyanH, line 38: Same thing for this test.

AliyanH (commented):
Just some minor styling and test changes. You will need to make changes to the [documentation](https://maps4html.org/web-map-doc/docs/elements/mapml-viewer/#attributes) also, you can link the web-map-doc PR to this PR when done. Good work on this!

AliyanH (approved):
LGTM!

prushforth:
Really good work, @Jacky0299! Thank you.  Squash and merge when ready!

prushforth:
Don't forget to do a 

"git pull upstream main" on your main local branch once you have merged upstream.

prushforth (approved):

PR #758, opened 2023-02-16
mapml-viewer controls attribute/property does not work as an API

### LINKED ISSUE #672: Appending a layer to layer-less map does not add layer controls
... while if a layer- element was pre written in the html it would appear as normal.
image

### LINKED ISSUE #673: `document.createElement("mapml-viewer")` fails
This is a bug; the constructor for a custom element should not create any DOM children, that work should be done in the connectedCallback.  

_Originally posted by @prushforth in https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/672#issuecomment-1182669630_

### LINKED ISSUE #675: mapml-viewer controls attribute/property does not work as an API
Toggling the value of either attribute or property does not toggle actual map controls

kevinkim31:
- [X] Fix mapml-viewer & web-map so they can be created via domAPI. closes #673 
- [X] Fix issue where layer control fails to appear when added programmatically (first layer only). closes #672 
- [X] Add domAPI testing
- [x] Get Testing working for web-map, using:
   - `viewer = document.createElement("map", {is:"web-map"})`
   - `viewer.setAttribute("is","web-map")`
- [X] Fix issue where controls fails to hide/disappear when removed programmatically
- [X] Add 'controlslist' as an attribute and make it work through using the API
   -  [X] implement 'controlslist' as a DOMTokenList
   -  [x] implement DOMTokenList as a class to help support it's [methods](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList)
- [X] `_changeWidth` and `_changeHeight` are causing errors in the console now (on web-map, same thing with mapml-viewer, when map created with width and height attributes it throws an error), as the _container does not get created in the constructor anymore, and is generated afterwards in the connectedcallback
 
closes #675


test/e2e/api/domApi.test.js
```diff
@@ -0,0 +1,92 @@
...
+    let layerControlHidden = await page.$eval(
+      "css=body > mapml-viewer >> css=div > div.leaflet-control-container > div.leaflet-top.leaflet-right > div",
+      (elem) => elem.hasAttribute("hidden")
+    );
+    expect(layerControlHidden).toEqual(false);
+
+    // set the layer's hidden attribute, the layer should be removed from the layer
+    // control (but not the map), which leaves 0 layers in the layer control, which means the
+    // layer control should disappear
+    await page.evaluateHandle((layer) => layer.setAttribute("hidden", "hidden"), layerHandle);
```
AliyanH, line 57: Good job on the testing, I think another test should be added similar to this, which checks for the layer control to disappear when you remove the layer element from the map.

AliyanH: Same tests for web-map should also be added.

AliyanH:
Added a relatively big commit [9717c44](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/758/commits/9717c44b574567af1084d82dbf744006cdaab885) to reimplement setControls(), which now allows the controls and controlslist API to work properly, Changes made:

- Added support for controlslist attribute, which is now represented by a [DomTokenList](https://developer.mozilla.org/en-US/docs/Web/API/DOMTokenList)
- Deconstructed the setControls function to 2 separate functions: 
   - setupControls() - creates the controls 
   - setControls() - which sets the controls visibility using the hidden attribute based on the map's control and controllist attribute
- Modified old test to change from looking for deleted controls to now looking for the hidden controls
- Added Dom API tests (more tests possibly needed)
   -  need to add tests for web-map which currently does not work when the map is created using `document.createElement`
- Removed duplicate test files (domApi.test.js + domApi.html)
- TODO - create a DomTokenList class for controlslist?

AliyanH:
- Added DomTokenList for the map's `controlsList`, removed getter and setter for the `controlslist` attribute, though controlslist attribute can still be set using `setAttribute` to emulate the behavior from the [video element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video).

Few notes/questions:
- jshint is throwing errors for build (when making private properties in the new class), possibly due to https://github.com/jshint/jshint/issues/3139
- currently the controlsListDomTokenList.js is added directly to the dist dir, is this appropriate?


- [ ] Need to add DomTokenList Specific tests

src/controlsListDomTokenList.js
```diff
@@ -0,0 +1,101 @@
+export default class DOMTokenList {
```
prushforth, line 1: If we think this class is general, we could just name the file DOMTokenList.js, since otherwise it associates to the controlsList property pretty strongly.

prushforth: Maybe the class isn't that general, please ignore

AliyanH: > Maybe the class isn't that general, please ignore

Yes unfortunately not.

prushforth (commented):
Just a few comments

src/mapml-viewer.js
```diff
@@ -579,7 +650,11 @@ export class MapViewer extends HTMLElement {
   }
   _toggleControls() {
```
prushforth, line 651: I am wondering if _toggleControls is needed any longer?  Can't we simply do document.querySelector('mapml-viewer').controls = false?  It seems to be used by ContextMenu.

src/mapml-viewer.js
```diff
@@ -17,20 +18,18 @@ export class MapViewer extends HTMLElement {
...
-    return this.hasAttribute('controlslist') ? this.getAttribute("controlslist") : "";
+  get controlsList() {
+    return this._controlsList;
   }
-  set controlslist(val) {
-    let options = ["nofullscreen", "nozoom", "nolayer", "noreload"],
-        lowerVal = val.toLowerCase();
-    if (this.controlslist.includes(lowerVal) || !options.includes(lowerVal))return;
-    this.setAttribute("controlslist", this.controlslist+` ${lowerVal}`);
+  set controlsList(value) {
```
prushforth, line 30: I noticed that if you have got a map state set with, for example conrolslist="nofullscreen" or anything else., if you do a map.removeAttribute("controlslist") it doesn't change the state of the map. You have to do a map.controlsList = "" to achieve that effect.  Not sure it's a big deal; probably an artifact of having the getter / setter property name be different than the attribute name.

src/mapml-viewer.js
prushforth: I removed it.

src/mapml-viewer.js
prushforth: This is recently fixed, good work!

prushforth:
LGTM, merge when you're happy with it.

PR #761, opened 2023-02-17
Place templated image layer at at (0,0) of CS when add it to map

### LINKED ISSUE #759: Templated image layer doesn't show at right place when turning it off then on
Have one templated image layer (_Toporma_) and a templated tile layer (_static MapML with tiles_) 
image
The two layers are expected to be "overlapped" when both of them are visible.

If move it up, and uncheck the templated image layer and check it again:

**Expect:**
the two layers are overlapped;
image
**Received:**
the templated image layer is slightly off.
image

However, the two layers will become overlapped again after moving it by dragging or panning.

yhy0217:
Closes #759

- [x] fix bug that the templated image layer is misplaced when it is re-added to map
- [x] fix bug that the templated image layer is duplicated when zooming in/out
- [x] add a test

### DIFF
--- a/src/mapml/layers/TemplatedImageLayer.js
+++ b/src/mapml/layers/TemplatedImageLayer.js
@@ -14,7 +14,8 @@ export var TemplatedImageLayer =  L.Layer.extend({
     },
     getEvents: function () {
         var events = {
-            moveend: this._onMoveEnd
+            moveend: this._onMoveEnd,
+            zoomstart: this._clearLayer,
         };
         return events;
     },
@@ -22,12 +23,6 @@ export var TemplatedImageLayer =  L.Layer.extend({
         this._map._addZoomLimit(this);  //used to set the zoom limit of the map
         this.setZIndex(this.options.zIndex);
         this._onAdd();
-        // make sure templatedImage Layer is placed at (0,0) of the cs
-        // when it is added to the map (solve issue #759)
-        if (this._layer.isUnChecked) {
-            this.redraw();
-            this._layer.isUnChecked = false;
-        }
     },
     redraw: function() {
         this._onMoveEnd();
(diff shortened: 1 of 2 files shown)


prushforth:
@yhy0217 I don't know if it's the same bug, but [this page](https://maps4html.org/experiments/custom-projections/BNG/) really demonstrates templated image misplacement

src/mapml/layers/TemplatedImageLayer.js
```diff
@@ -132,6 +139,7 @@ export var TemplatedImageLayer =  L.Layer.extend({
     onRemove: function (map) {
       this._clearLayer();
       map._removeZoomLimit(this);
+      this._layer.isUnChecked = true;
```
AliyanH, line 142: why is `this._layer` added and what's is use? can't see where this is being used

src/mapml/layers/TemplatedImageLayer.js
```diff
@@ -45,6 +47,7 @@ export var TemplatedImageLayer =  L.Layer.extend({
             this._imageOverlay._overlayToRemove = overlayToRemove._url;
             this._imageOverlay.on('load error', function () {map.removeLayer(overlayToRemove);});
         }
+        this.location = loc;
```
AliyanH, line 50: What's the use for `this.location`? I can see it is only used in the tests. I don't think it should be added if it's only use is for testing?

src/mapml/layers/TemplatedImageLayer.js
yhy0217: Yes it is from the previous commit I made, I simply forgot to delete this. What I was trying to do in my previous commit was to ensure the redraw() function should only be called inside the onAdd function when the layer is **re**-added to the map. I added it as an attribute of this._layer is simply because I found the templated image layer itself as well as its container will be cleared once I turn the layer off, so if I make it as an attribute of this, it can no longer be used when I add the layer back. Overall, thank you for bringing this up, I will remove it from the codes!

src/mapml/layers/TemplatedImageLayer.js
yhy0217: Yes it is only for testing. if it is not appropriate to add it, I can try to find another way to test it. I just tried some ways to test (count the  number of layer etc.), and then found they are much harder than testing in this way

src/mapml/layers/TemplatedImageLayer.js
```diff
@@ -1,6 +1,7 @@
 export var TemplatedImageLayer =  L.Layer.extend({
     initialize: function(template, options) {
         this._template = template;
+        this._layer = options.pane.parentElement;
```
prushforth, line 4: Don't need this anymore, I believe, with your latest change.

prushforth:
LGTM

Issue #201, opened 2020-11-04
Keyboard shortcut & context menu item to toggle fullscreen mode


Malvoz:
As suggested in https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/18#issuecomment-588576345, it'd be great to have a context menu item to toggle fullscreen mode.

IMO it is a highly valuable feature. It should be available even if the author chooses to disable the control in the UI using `controlslist="nofullscreen"` (disabling a control does not disable its corresponding feature, per the spec).

I believe the conventional keyboard shortcut for fullscreen is <kbd>F</kbd>, which would clash with the "Forward (<kbd>F</kbd>)" menu item. 🤔 In the gaming world it is not uncommon to navigate (e.g. move a character) using the [WASD keys](https://en.wikipedia.org/wiki/Arrow_keys#WASD_keys), so maybe that's something the Back/Forward/Reload items could use instead? 🤷‍♂️

prushforth:
Resolved by https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/765

PR #765, opened 2023-02-21
full screen context menu item added and fixed some context menu related tests

Jacky0299:
New "Exit Fullscreen" locale added to the mapml-extension here: https://github.com/Maps4HTML/mapml-extension/commit/c446c5410438235437d2b51ffcd58ebc694d5568

solves issue #201


test/e2e/core/mapContextMenu.test.js
```diff
@@ -534,4 +538,31 @@ test.describe("Playwright Map Context Menu Tests", () => {
...
+    expect(fullScreen).toEqual('View fullscreen (E)');
+    let copy = await page.$eval("div > div.mapml-contextmenu > button:nth-child(6)",(btn) => btn.textContent);
+    expect(copy).toEqual('Copy (C)');
+    let paste = await page.$eval("div > div.mapml-contextmenu > button:nth-child(8)",(btn) => btn.textContent);
+    expect(paste).toEqual('Paste (P)');
+    let controls = await page.$eval("div > div.mapml-contextmenu > button:nth-child(10)",(btn) => btn.textContent);
+    expect(controls).toEqual('Toggle Controls (T)');
+    let debug = await page.$eval("div > div.mapml-contextmenu > button:nth-child(11)",(btn) => btn.textContent);
+    expect(debug).toEqual('Toggle Debug Mode (D)');
+    let source = await page.$eval("div > div.mapml-contextmenu > button:nth-child(12)",(btn) => btn.textContent);
```
AliyanH, line 559: Good work testing this, this test is very important! I wonder if you should also add test to check `button:nth-child(13)` to make sure that does not exist, will help with testing if we decide to add more items to the contextmenu.

AliyanH:
When I enter fullscreen by clicking on the button, The contextmenu still says "View Fullscreen" instead of "Exit Fullscreen". Also since you are adding a new locale option (btnExitFullScreen), you should also create translations for it on the extension, similar to what's been done [here](https://github.com/Maps4HTML/mapml-extension/pull/47).

prushforth:
The context menu works when you enter and exit fullscreen via the context menu, but if you enter fullscreen via the context menu and exit via "Escape", the context menu still continues to read "Exit fullscreen"

prushforth (changes requested):

prushforth:
A good place to start reading about the browsers' Fullscreen API is [here](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API/Guide); there are likely things in there that would be useful in the code for this PR.

AliyanH:
Found a weird behavior when navigating the history using alt+Arrow keys, when I go back in history many times using the alt+back arrow key, the forward history seems to get deleted, as seen below:

https://user-images.githubusercontent.com/55751566/221648685-a4ebb1cb-e45f-439b-b0d8-d6de7287a2ea.mp4

src/mapml/handlers/ContextMenu.js
```diff
@@ -629,19 +638,22 @@ export var ContextMenu = L.Handler.extend({
       case 77: //M KEY
         this._copyMapML(e);
         break;
-      case 70: //F KEY
-        this._goForward(e);
-        break;
+      // case 70: //F KEY
```
prushforth, line 641: We shouldn't leave commented code here; I think you have replace F with Alt+Right-Arrow, so maybe this is redundant now?

src/mapml/handlers/ContextMenu.js
```diff
@@ -629,19 +638,22 @@ export var ContextMenu = L.Handler.extend({
...
+      case 70: //F KEY
+        this._toggleFullScreen(e);
+        break;
       case 80: //P KEY
         this._paste(e);
         break;
-      case 82: //R KEY
-        this._reload(e);
-        break;
+      // case 82: //R KEY
```
prushforth, line 654: Let's get rid of commented code.

src/web-map.js
```diff
@@ -590,6 +590,32 @@ export class WebMap extends HTMLMapElement {
...
+      if (document.fullscreenElement === null) {
+        // full-screen mode has been exited
+        this._map.contextMenu.setViewFullScreenInnerHTML('view');
+      }else{
+        this._map.contextMenu.setViewFullScreenInnerHTML('exit');
+      }
+    });
+    this.addEventListener('keydown', function(event) {
+      // Check if Ctrl+R is pressed and map is focused
+      if (event.ctrlKey && event.keyCode === 82 && document.activeElement.nodeName === "DIV") {
```
prushforth, line 603: This looks a bit funny.  The event listener is checking to see that the active element is a div, which is very non-specific, so it will fire more often than you wish.

prushforth (commented):
A couple of small clarifying changes

prushforth:
> Found a weird behavior when navigating the history using alt+Arrow keys, when I go back in history many times using the alt+back arrow key, the forward history seems to get deleted, as seen below:
> 
> Probably because the alt+arrowkeys are still doing something when at the end of history, when they are at the end of history it should just prevent default and not be going forward/backwards - that might be the cause of this.

Is this on web-map or mapml-viewer?  Not sure if it makes a difference, but I noticed that alt+left-arrow is enabled on any div, not just the map div.

prushforth:
@Jacky0299 does this require an additional pull request on the extension for the translated strings?

Jacky0299:
> Found a weird behavior when navigating the history using alt+Arrow keys, when I go back in history many times using the alt+back arrow key, the forward history seems to get deleted, as seen below:
> 
> Probably because the alt+arrowkeys are still doing something when at the end of history, when they are at the end of history it should just prevent default and not be going forward/backwards - that might be the cause of this.

I think it is because pressing it too fast, if you press it once by once, it's working fine, I can try to put a timeout between each time back() and forward() triggers.

prushforth:
I noticed that if you use "Alt+Left-Arrow" until there is no more history, the "Reload" button does not become disabled, until you actually reload / Ctrl-R.

Jacky0299:
> 


> I noticed that if you use "Alt+Left-Arrow" until there is no more history, the "Reload" button does not become disabled, until you actually reload / Ctrl-R. That is, when you pop the last history item, it should disable the reload button.

I don't really see this problem at my side, when there's no more history, the reload is disabled. It should be the same as before since I did not really touch the functions.

src/mapml-viewer.js
```diff
@@ -550,6 +550,32 @@ export class MapViewer extends HTMLElement {
...
+      if (event.altKey && event.keyCode === 39 && document.activeElement.nodeName === "MAPML-VIEWER") {
+        // Prevent default browser behavior for Alt+Right-Arrow, if any
+        event.preventDefault();
+        this.forward();
+      }
+      if (event.altKey && event.keyCode === 37 && document.activeElement.nodeName === "MAPML-VIEWER") {
+        // Prevent default browser behavior for Alt+Left-Arrow (Back in browser history)
+        event.preventDefault();
+        this.back();
+      }
```
AliyanH, line 577: it would be good to put all of this code inside the following if statement, so the condition isnt being checked 3 times:
```
if (document.activeElement.nodeName === "MAPML-VIEWER") {
      // code here
}
```
Similarly, instead of having 3 separate if  statements, you can use if/else if to save extra computation.

AliyanH: Same thing for web-map

AliyanH (commented):
code structure change

AliyanH:
I think we can revert the settimeout change and look at async await if possible.

With the settimeout, we are only able to go back with intervals of 0.5 seconds. So If I were to go back twice within 0.5 secs, It would only go back once:

https://user-images.githubusercontent.com/55751566/221908052-b6f9bac8-e0ab-449b-b44f-e39603c7f3e2.mp4

src/mapml-viewer.js
```diff
@@ -639,13 +676,14 @@ export class MapViewer extends HTMLElement {
...
-      this._map.contextMenu._items[1].el.el.disabled = false; // forward contextmenu item
+      this._map.contextMenu.toggleContextMenuItems("Forward", "enabled");// forward contextmenu item
       this._historyIndex--;
       let prev = history[this._historyIndex];
       // Disable back, reload contextmenu item when at the end of history
       if (this._historyIndex === 0) {
-        this._map.contextMenu._items[0].el.el.disabled = true; // back contextmenu item
-        this._map.contextMenu._items[2].el.el.disabled = true; // reload contextmenu item
+        this._map.contextMenu.toggleContextMenuItems("Back", "disabled"); // back contextmenu item
+        this._map.contextMenu.toggleContextMenuItems("Reload", "disabled"); // reload contextmenu item
```
prushforth, line 685: Here is also where you should disable the reload button on the map

src/mapml/handlers/ContextMenu.js
```diff
@@ -697,5 +700,38 @@ export var ContextMenu = L.Handler.extend({
   _onItemMouseOut: function (e) {
     L.DomUtil.removeClass(e.target || e.srcElement, 'over');
     this._hideCoordMenu(e);
+  },
+
+  toggleContextMenuItems: function (options,state) {
+    if (state === "disabled"){
+      if (options === "Toggle Controls"){
```
prushforth, line 707: See previous comments about using a case-insensitive single token, not a full string.

src/mapml/handlers/ContextMenu.js
```diff
@@ -697,5 +700,38 @@ export var ContextMenu = L.Handler.extend({
   _onItemMouseOut: function (e) {
     L.DomUtil.removeClass(e.target || e.srcElement, 'over');
     this._hideCoordMenu(e);
+  },
+
+  toggleContextMenuItems: function (options,state) {
```
prushforth, line 705: Change name to toggleContextMenuItem

src/mapml/handlers/ContextMenu.js
```diff
@@ -697,5 +700,38 @@ export var ContextMenu = L.Handler.extend({
...
+      if (options === "Toggle Controls"){
+        this._items[8].el.el.disabled = true;
+      }else if (options === "Back"){
+        this._items[0].el.el.disabled = true;
+      }else if (options === "Forward"){
+        this._items[1].el.el.disabled = true;
+      }else if (options === "Reload"){
+        this._items[2].el.el.disabled = true;
+      }
+    }else if(state === "enabled"){
```
prushforth, line 716: Be consistent about spaces around javascript keywords.  See previous comment about format.

prushforth (changes requested):
A few comments.  Almost there!

src/mapml-viewer.js
```diff
@@ -529,6 +529,7 @@ export class MapViewer extends HTMLElement {
       function () {
         this._updateMapCenter();
         this._addToHistory();
+        this._reloadButtonState();
```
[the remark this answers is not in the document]
Jacky0299: _addToHistory is called before the reloadbutton is added to the map in the constructor, so it would not work if I put this function in _addToHistory

src/mapml-viewer.js
```diff
@@ -621,16 +648,26 @@ export class MapViewer extends HTMLElement {
...
-      this._map.contextMenu._items[0].el.el.disabled = false; // back contextmenu item
-      this._map.contextMenu._items[1].el.el.disabled = true; // forward contextmenu item
-      this._map.contextMenu._items[2].el.el.disabled = false; // reload contextmenu item
+      this._map.contextMenu.toggleContextMenuItems("Back", "enabled"); // back contextmenu item
+      this._map.contextMenu.toggleContextMenuItems("Forward", "disabled");// forward contextmenu item
+      this._map.contextMenu.toggleContextMenuItems("Reload", "enabled"); // reload contextmenu item
+    }
+  }
+  _reloadButtonState(){
+    if(this.hasAttribute("controls")){
```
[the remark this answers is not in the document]
Jacky0299: If the map does not have controls buttons, the _reloadbutton is gone

prushforth:
@Jacky0299 I pushed some changes (pull them and have a look!) that maybe weren't clear; I left the comments above so you can compare what I did with what the comment asked for.

src/mapml-viewer.js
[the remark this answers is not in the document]
prushforth: See the code I pushed; if the reload button doesn't exist, it won't enable/disable.  I removed the _reloadButtonState() method and replaced it with this._reloadButton?.enable() / .disable() in the place where the other enabling / disabling is done.   Hopefully because the code is all co-located it will be more "maintainable", at least that's the theory.

prushforth (approved):

src/mapml/handlers/ContextMenu.js
```diff
@@ -17,17 +17,21 @@ export var ContextMenu = L.Handler.extend({
     //setting the items in the context menu and their callback functions
     this._items = [
       {
-        text: M.options.locale.cmBack + " (<kbd>B</kbd>)",
+        text: M.options.locale.cmBack + " (<kbd>Alt+Left Arrow</kbd>)",
         callback:this._goBack
       },
       {
-        text: M.options.locale.cmForward + " (<kbd>F</kbd>)",
+        text: M.options.locale.cmForward + " (<kbd>Alt+Right Arrow</kbd>)",
```
AliyanH, line 24: Looks good, just dont forget to add localized versions of this in the extension, similar to what you did for "Exit Fullscreen". Create a PR for the extension.

Nice Work on this!

AliyanH (changes requested):

PR #768, opened 2023-02-23
Remove support for sharding

AliyanH:
Addresses https://github.com/Maps4HTML/MapML/issues/233.

### DIFF
--- a/src/mapml/layers/MapMLLayer.js
+++ b/src/mapml/layers/MapMLLayer.js
@@ -926,21 +926,7 @@ export var MapMLLayer = L.Layer.extend({
 
                 inputs.push(inp);
                 includesZoom = includesZoom || inp.hasAttribute("type") && inp.getAttribute("type").toLowerCase() === "zoom";
-                if (inp.hasAttribute('shard')) {
-                  var id = inp.getAttribute('list');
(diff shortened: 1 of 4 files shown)


AliyanH:
I am not sure if this code is linked to sharding or not:
https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/0e006d230b3726b8f2c40c7c749f444b1068d0b1/src/mapml/layers/TemplatedTileLayer.js#L170-L174

prushforth:
> I am not sure if this code is linked to sharding or not:
> 
> https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/0e006d230b3726b8f2c40c7c749f444b1068d0b1/src/mapml/layers/TemplatedTileLayer.js#L170-L174

No.  It tests to see if there are any variables in the URL template that aren't part the parameters required to obtain a tile i.e. the row/col/zoom or the tile's extent coordinates (left/right/top/bottom).  If such a variable exists, it's name is made a member of the variables who's value will be generated via controls in the layer control e.g. time select or something like that.

prushforth:
LGTM

prushforth (approved):

PR #783, opened 2023-03-09
opacity attribute for map extent solves issue #735

Jacky0299:
solves issue #735

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -123,11 +123,24 @@ module.exports = function(grunt) {
             dest: 'dist/images/'
           }
         ]
+      },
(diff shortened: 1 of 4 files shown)


AliyanH (commented):
Looking great so far Jacky, just a few comments. 

Also, was looking at the onAdd function, I know its not your code, but I think [this ](https://github.com/AliyanH/Web-Map-Custom-Element/blob/main/src/mapml/layers/MapMLLayer.js#L122)is a bug right?

It should be `this._extent**.**_mapExtents` instead of what it is currently on main:
`this._extent_mapExtents`

PR #790, opened 2023-03-13
Format Code - prettier

AliyanH:
created a new grunt task (to replace the 'test' grunt task) to format the code using prettier.
One of the code formats ([Conditional ternary operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator)) performed by prettier seems to cause errors with jshint.

Seems to format code nicely, though we can finetune it as needed also - https://prettier.io/docs/en/options.html

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,9 @@
+{
+  "singleQuote": true,
+  "overrides": [
+		{
+			"files": ".firebaserc",
+			"options": { "parser": "json" }
+		}
+	]
+}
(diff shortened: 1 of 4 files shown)


prushforth (approved):
Looks good to me.  Once we squash and merge, we can add a .git-blame-ignore-revs file to the root folder, containing the commit hash of the squashed commit.  We also have to remember to document that new devs should use the following command to ignore this particular commit, permanently:

```bash
 git config --global blame.ignoreRevsFile .git-blame-ignore-revs
```

PR #794, opened 2023-03-13
Add a dynamic scale bar and announce a scale ratio for each map view for screen readers

### LINKED ISSUE #777: Add a dynamic scale bar and announce a scale ratio for each map view for screen readers
The scale bar might be a candidate for a new control, and should be dynamically rendered on move end.  Similarly, the scale ratio at the centre of the map should be announced.

Removal of scale bar control could be achieved via controlslist="noscale", which would be a new value.

kevinkim31:
closes #777


prushforth:
Haven't listened to it yet, but looks nice.  I think you'll have to adjust the space between the scalebar and the bottom of the map, up a bit, because on [maps that display links](https://maps4html.org/experiments/custom-projections/Arctic-SDI/), the links draw over top of the scalebar.

kevinkim31:
> initial announce template

Not ready for review just yet. Still need to work on the announcing but should be done soon.

src/mapml-viewer.js
```diff
@@ -194,6 +194,29 @@ export class MapViewer extends HTMLElement {
         }
       }, 0);
     }
+
+    let scaleinitial = this.shadowRoot.querySelector(".leaflet-container");
```
AliyanH, line 198: make sure all the new variables you create are lower camelCase to stay consistent with the code base.

src/mapml/control/AnnounceScale.js
```diff
@@ -0,0 +1,67 @@
...
+        let invisibleScaleLine = invisibleScale._container.getElementsByClassName('leaflet-control-scale-line')[0];
+
+        container.setAttribute('aria-label',  this._setText(invisibleScaleLine));
+        container.style.display = 'none';
+      });
+
+      return container;
+    },
+
+    _pixelsToDistance: function (px, units) {
```
prushforth, line 32: It looks to me like the announced scale ratio calculates the length of the CSS line to be about 1/2 what it is physically. i.e. when it says "1 centimeter", it looks like about 2 centimeters to me.

prushforth: "96 dpi" is in inches so maybe that accounts for the discrepancy?

kevinkim31: hmm. The distances look pretty accurate on my screen. Do you know what zoom level it says 1cm?

src/mapml/control/AnnounceScale.js
```diff
@@ -0,0 +1,67 @@
+export var AnnounceScale = L.Control.extend({
+    options: {
+      position: 'bottomleft',
+      maxWidth: 100,
+      updateWhenIdle: true,
+    },
+  
+    onAdd: function (map) {
+      var container = L.DomUtil.create('div', 'accessible-scalebar');
```
prushforth, line 9: I think you might want to maintain a "private" `_container` property, that you could return with a custom AnnounceScale.getContainer() method.

src/mapml/control/ScaleBar.js
```diff
@@ -0,0 +1,48 @@
...
+      position: 'bottomleft'
+    },
+
+    onAdd: function (map) {
+      let container = L.DomUtil.create('div','leaflet-control-scale');
+
+      let scaleControl = L.control.scale(this.options);
+      scaleControl.addTo(map);
+      container.appendChild(scaleControl.getContainer());
+      container.title = "Scale bar";
```
prushforth, line 16: I think this announces "Scale bar" twice, because below you add `aria-label` and put "Scale bar: blah blah blah" in there.  Maybe ditch the title unless it gets added automatically?

src/mapml/control/ScaleBar.js
```diff
@@ -0,0 +1,48 @@
...
+      container.setAttribute('tabindex',0);
+
+      this._container = container;
```
prushforth, line 19: Unless the leaflet `L.Control.getContainer()` returns this container, I think you should create an overriding `M.ScaleBar.getContainer()`, so you can work with the container for this control.

src/mapml/control/ScaleBar.js
```diff
@@ -0,0 +1,48 @@
+export var ScaleBar = L.Control.extend({
+    options: {
+      metric: true,
```
prushforth, line 3: I think you should use the value specified by `M.options.whatever` to restrict the scale to metric or imperial, but not both.

src/mapml/control/ScaleBar.js
```diff
@@ -0,0 +1,48 @@
...
+
+      let scaleLineMetric = this._scaleControl._container.getElementsByClassName('leaflet-control-scale-line')[0].textContent.trim();
+      scaleLineMetric = scaleLineMetric.replace(/(\d+)\s*m\b/g, "$1 meters");
+      scaleLineMetric = scaleLineMetric.replace(/ km/g, " kilometers");
+
+      let scaleLineImperial = this._scaleControl._container.getElementsByClassName('leaflet-control-scale-line')[1].textContent.trim();
+      scaleLineImperial = scaleLineImperial.replace(/ft/g, "feet");
+      scaleLineImperial = scaleLineImperial.replace(/mi/g, "miles");
+
+      this._container.setAttribute('aria-label', `Scale bar: ${scaleLineMetric} or ${scaleLineImperial}`);
```
prushforth, line 43: Pick just one of metric or imperial, based on the `M.options.whatever` value.  I think you should be putting the actual scale here, just as you do in the `<output>` element, if that's possible?

src/mapml/control/ScaleBar.js
```diff
@@ -0,0 +1,48 @@
...
+    },
+  });
+export var scaleBar = function (options) {
+	return new ScaleBar(options);
+};
```
prushforth, line 48: GitHub leaves you a "No Entry" sign image when you omit the required blank line at the end of a  file.

src/web-map.js
```diff
@@ -211,6 +234,9 @@ export class WebMap extends HTMLMapElement {
     let output = "<output role='status' aria-live='polite' aria-atomic='true' class='mapml-screen-reader-output'></output>";
     this._container.insertAdjacentHTML("beforeend", output);
 
+    let outputScale = "<output role='status' aria-live='polite' aria-atomic='true' class='mapml-screen-reader-output-scale'></output>";
```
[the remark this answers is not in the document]
kevinkim31: Yes, I believe it would be more organized and also easier to debug (future purposes) for scale bar to have its own output element. As well, adding text onto another element with already existing text might cause timing issues related to NVDA.

src/mapml/control/ScaleBar.js
```diff
@@ -0,0 +1,95 @@
+export var ScaleBar = L.Control.Scale.extend({
+    options: {
+      maxWidth: 100,
+      updateWhenIdle: true,
+      position: 'bottomleft'
+    },
+
+    _container: null,
```
prushforth, line 8: A way to initialize stuff is in a constructor, which in leaflet is always called `initialize`.  You can also do a setOptions in there to merge the options you set above with the options that may get passed to `initialize`: https://leafletjs.com/examples/extending/extending-1-classes.html#lclassinitialize

src/mapml-viewer.js
```diff
@@ -341,6 +342,14 @@ export class MapViewer extends HTMLElement {
 
     this._layerControl = M.layerControl(null,{"collapsed": true, mapEl: this}).addTo(this._map);
 
+    let scaleValue = M.options.announceScale;
+
+    if (typeof scaleValue === "string") {
+      scaleValue = JSON.parse(scaleValue);
+    }
+
+    this._scaleBar = M.scaleBar(scaleValue).addTo(this._map);
```
prushforth, line 351: I think this should be conditional on it not existing yet, as I believe this is how the controls work now.  i.e. if (!this._scaleBar) this._scaleBar = ...

prushforth: Maybe? Also perhaps there's a method on L.ScaleBar to allow you to switch measurement system dynamically?

src/mapml/control/ScaleBar.js
prushforth: I tried removing your _container: null member and replacing it with the equivalent inside a constructor, but the scalebar started showing both imperial and metric at the same time again, so there's something else that I'm missing here.

prushforth (approved):
LGTM! Great work Kevin!

PR #796, opened 2023-03-13
make map move when setters used for lat/lon/zoom

### LINKED ISSUE #588: Setters for zoom, lat, lon should change / move the map location
There is the zoomTo method to change all three at once, yes. But from an API perspective, if these attributes and properties aren't read-only, setting them should have a visible effect.  I don't recall if there's an actual reason that they don't, but if it's a problem we should document that, or fix it.

AliyanH:
- [X] Pan/zoom the map when lat/lon/zoom are set.
- [x] Add tests
- [ ] update [documentation](https://github.com/Maps4HTML/web-map-doc/issues/56)
- [x] Update [Spec](https://github.com/Maps4HTML/MapML/pull/239)

resolves #588.

### DIFF
--- a/src/mapml-viewer.js
+++ b/src/mapml-viewer.js
@@ -51,6 +51,9 @@ export class MapViewer extends HTMLElement {
   set lat(val) {
     if (val) {
       this.setAttribute("lat", val);
+      if (this._mapZoomLocationAPI) {
+        this.zoomTo(val,this.lon,this.zoom);
+      }
     }
   }
   get lon() {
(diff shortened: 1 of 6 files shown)


src/mapml-viewer.js
prushforth, line 1: Since there are no attributeChangedHandler switches lat,lon,zoom maybe it would be a good idea to remove them from the observedAttributes array?  Although this was the case before, we didn't have such switches then either, I imagine that it's triggering something and it occurred to me that they could be removed especially if we are going to ignore changes to the attributes?

prushforth:
This looks good.  Let's wait ${some_time} to see if we get an answer about [whether we **should** be reflecting lat,lon,zoom to the content attributes](https://github.com/w3ctag/design-principles/issues/424).  We will have to update the spec to codify the behaviour, one way or another.

AliyanH:
> This looks good. Let's wait ${some_time} to see if we get an answer about [whether we **should** be reflecting lat,lon,zoom to the content attributes](https://github.com/w3ctag/design-principles/issues/424). We will have to update the spec to codify the behaviour, one way or another.

Yup, sounds good!

PR #797, opened 2023-03-13
Rename the layer-.focus() method to layer-.zoomTo()

### LINKED ISSUE #587: <layer->.focus() is misnamed, should be split up and renamed
There is a standard [HTMLElement.focus() method](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus), which doesn't do what this method does, which is to zoom to the whole layer at the maximum possible zoom.  We will need, at a minimum, to rename this method.

Since programmatic focus has become important, the HTMLLayerElement.focus() implementation should either focus the layer in the layer control, or focus the first or next element in the layer, without moving the map.  We should figure out what the correct behaviour is, but I'm pretty sure the correct behaviour is not what it currently does.  We probably need a separate method for each type of focus, actually.

prushforth:
Closes #587

### DIFF
--- a/src/layer.js
+++ b/src/layer.js
@@ -324,7 +324,7 @@ export class MapLayer extends HTMLElement {
               {target: this}}));
         },this);
   }
-  focus(){
+  zoomTo() {
     if(!this.extent) return;
     let map = this._layer._map,
       tL = this.extent.topLeft.pcrs,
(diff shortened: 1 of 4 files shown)


AliyanH (approved):
looks good! Was pretty confusing distinguishing between the two different focus() being used, this should help clear that up!

PR #798, opened 2023-03-14
Develop <map-feature> custom element

### LINKED ISSUE #528: Implement `<map-feature>` custom element
We should try to implement `<map-feature>` as a custom element.  Some behaviour that this should encompass, tentative list:

- ability to add/remove `<map-feature>` as a child (or perhaps descendant) of `<layer->` and render / remove the rendered resulting markup (SVG).
- perhaps: modify the coordinates and have those modifications rendered dynamically
- provide internal support for returning a GeoJSON representation of `<map-feature>` geometry. The objective is to be able to use turf.js to sort and analyze features for accessibility support e.g. sort ascending on distance from map centre
- per discussion with @AliyanH, a custom element could directly provide interactivity, such as a .click() method, so that applications could work with features on the map without having to delve into our shadow DOM
- method to calculate and return the extent of the feature as a JavaScript object

yhy0217:
Closes #528 
- [x] make mapFeature element interactive
-   when mapFeature is removed from DOM, the associated g and path elements are removed
-   when mapFeature (and its children elements) is re-added / updated in the DOM, the associated g and path elements are updated dynamically

- [x] provide methods and APIs as suggestion in #528
-   .click(): simulate a click event by default; customizable to users
-   .focus(): simulate a mouse-hover event by default; customizable
-   .geometryTogeojson(): return geojson representation of geometry
-   .extent: return feature extent

- [x] create and attach a shadow root for the layer- elements with src attribute
- [x] create tests for mapFeature custom element


src/mapml/layers/MapMLLayer.js
```diff
@@ -1136,8 +1162,37 @@ export var MapMLLayer = L.Layer.extend({
...
+              layer._layerEl.parentElement._toggleControls();
+            }
             layer._layerEl.dispatchEvent(new CustomEvent('extentload', {detail: layer,}));
-        }
+          }
+
+          function _attachToLayer() {
+            let mapml = xhr.responseXML;
+            if (mapml) {
+              let shadowRoot = this._layerEl.shadowRoot || this._layerEl.attachShadow({mode: 'open'});
```
prushforth, line 1176: I think the attachShadow should be done by the `<layer->` element, either during the constructor (preferable) or the connectedCallback (less preferred, I think, because you then have to check if it exists, whereas if you do it during construction, it will exist from then on).

src/map-feature.js
```diff
@@ -0,0 +1,266 @@
...
+          document.dispatchEvent(new MouseEvent('click', {
+            bubbles: false
+          }));
+        }
+        // AFTER the mousedown and mouseup events:
+        // case 1: the layer el is not re-attached to the map, the <map-feature> el is still CONNECTED
+        // case 2: the layer el is re-attached to the map; the disconnectedCallback() is invoked;
+        //         the <map-feature> el (THIS) is now DISCONNECTED, this._featureLayer is removed and a new featureLayer is created
+        if (properties && this.isConnected) {
+          this._featureLayer.isPopupOpen() ? this._featureLayer.closePopup() : this._featureLayer.openPopup();
```
prushforth, line 228: This is giving a JSHint warning, should be replaced with:

```javascript
if (this._featureLayer.isPopupOpen()) {
  this._featureLayer.closePopup();
} else {
  this._featureLayer.openPopup();
}
```

src/mapml/layers/MapMLLayer.js
```diff
@@ -1136,8 +1162,37 @@ export var MapMLLayer = L.Layer.extend({
...
+              if (elements) {
+                let baseURL = mapml.children[0].children[0].querySelector('map-base')?.getAttribute('href');
+                for (let el of elements) {
+                  // resolve relative url
+                  if (el.nodeName === 'map-link') {
+                    let url = el.getAttribute('href') || el.getAttribute('tref');
+                    // if relative
+                    if (url && (url.indexOf('http://') === 0 || url.indexOf('https://') === 0)) {
+                      let resolvedURL = baseURL + url;
+                      el.hasAttribute('href') ? el.setAttribute('href', resolvedURL) : el.setAttribute('tref', resolvedURL);
```
prushforth, line 1187: Replace with

```javascript
if (el.hasAttribute('href') {
  el.setAttribute('href', resolvedURL);
} else {
  el.setAttribute('tref', resolvedURL);
}
```

merci
```

File: src/map-feature.js [hunk 4]
```diff
@@ -0,0 +1,266 @@
...
+      super();
+    }
+
+    connectedCallback() {
+      if (this.parentNode.nodeType !== document.DOCUMENT_FRAGMENT_NODE && 
+          this.parentNode.tagName.toLowerCase() !== 'layer-') {
+        return;
+      }
+
+      this._layerParent = this.parentNode._layer ? this.parentNode._layer : this.parentNode.host._layer;
```
prushforth, line 46: I think I understand but a comment may be warranted: the feature can be a child of `<layer->`, directly, or may have been created by migration from a `<mapml->` document loaded via `<layer- src=...>`, in which case it is re-parented to exist in the shadowRoot of the `<layer- src>.shadowRoot` (a document fragment).  Is that right?

src/map-feature.js
```diff
@@ -0,0 +1,266 @@
...
+      this._map = this._layerParent._map;
+      if(this._layerParent._layerEl.hasAttribute("data-moving")) return;
+      
+      this._observer = new MutationObserver((mutationList) => {
+        // map-properties, map-geometry, map-featurecaption, map-a
+        for (let i = 0; i < mutationList.length; ++i) {
+          this._remove();
```
prushforth, line 53: Seems odd to do these functions in a loop, but you can explain it to us!

src/mapml/layers/MapMLLayer.js
yhy0217: Is it better to make it done in connectedCallback()? Because we need to check attribute "src" to see whether a shadow root is necessary, and attributes of elements are only meaningful AFTER connecting to the DOM tree I believe?

prushforth: Yup if attributes are in play, connectefCallback is the right choice, I believe

PR #800, opened 2023-03-15
geolocation button closes issue#780

### LINKED ISSUE #780: Implement geolocation control
i.e. a map control, possibly lower right corner by default which allows the user to request that the map centre on their location as reported by the geolocation API. Maybe implement with a plain ol radio button:

image

Jacky0299:
closes #780


src/mapml-viewer.js
```diff
@@ -324,6 +324,16 @@ export class MapViewer extends HTMLElement {
         totalSize += 49;
         this._fullScreenControl = M.fullscreenButton().addTo(this._map);
       }
+      if (this.controlslist.toLowerCase().includes("geolocation") && !this._geolocationButton && (totalSize + 49) <= mapSize){
+        totalSize += 49;
+        this._geolocationButton = M.geolocationButton({
+          showPopup: false,
```
AliyanH, line 330: I was listening to the page using NVDA, maybe we should change the strings [option](https://github.com/domoritz/leaflet-locatecontrol#possible-options), so it says "locate button" or something of that nature.

AliyanH (commented):
Was just taking a look at the `git log` for your PR, seems to be pretty outdated, make sure to pull from upstream as there will be some conflicts!

Also don't forget web-map!

src/mapml-viewer.js
```diff
@@ -354,6 +354,19 @@ export class MapViewer extends HTMLElement {
       totalSize += 49;
       this._fullScreenControl = M.fullscreenButton().addTo(this._map);
     }
+    if (this.getAttribute("controlslist") === "geolocation" && !this._geolocationButton && (totalSize + 49) <= mapSize){
```
prushforth, line 357: Since the geolocation control is no longer at the top left of the map, I don't believe it needs to be taken into consideration in the totalSize calculation. If I recall correctly the purpose of this calculation, it's to not add a control if it will not be visible on a very small map; since we moved this control to the bottom right, it shouldn't be considered for that purpos.

AliyanH:
Right now geolocation only works when a map is created with a controlslist attribute with **only** geolocation:
```html
<mapml-viewer projection="CBMTILE" controlslist="geolocation" zoom="2" lat="45" lon="-90" controls>
   <map-caption>A pleasing map of Canada</map-caption>
   <layer- label="CBMT" src="https://geogratis.gc.ca/mapml/en/cbmtile/cbmt/" checked></layer->
</mapml-viewer>
```

when adding anything to the controlslist attribute alongside geolocation, it does not function, for ex. when given the following, geolocation control does not display:

`controlslist="geolocation nolayer"`

setting the geolocation using [the API](https://maps4html.org/web-map-doc/docs/api/mapml-viewer-api#controlslist) also does not work, make sure to update the geolocation code to match the implementation of the other controls.

Gruntfile.js
```diff
@@ -9,7 +9,7 @@ module.exports = function(grunt) {
       },
       combine: {
         files: {
-        'dist/mapml.css': ['node_modules/leaflet/dist/leaflet.css', 'src/mapml.css']
+        'dist/mapml.css': ['node_modules/leaflet/dist/leaflet.css', 'node_modules/leaflet.locatecontrol/dist/L.Control.Locate.css', 'node_modules/leaflet.locatecontrol/dist/L.Control.Locate.mapbox.css', 'src/mapml.css']
```
prushforth, line 12: Is the mapbox.css actually necessary here?

src/mapml-viewer.js
```diff
@@ -354,6 +354,18 @@ export class MapViewer extends HTMLElement {
...
+      this._geolocationButton = M.geolocationButton({
+        showPopup: false,
+        strings: {
+          title: "Locate"
+        },
+        position: "bottomright",
+        locateOptions: {
+          maxZoom: 16
+        },
+      },this._map);
```
prushforth, line 367: I believe this should be:

```javascript
  this._geolocationButton = M.geolocationButton({...}).addTo(this._map)
```
but it seems to work??

src/mapml-viewer.js
```diff
@@ -354,6 +354,18 @@ export class MapViewer extends HTMLElement {
       totalSize += 49;
       this._fullScreenControl = M.fullscreenButton().addTo(this._map);
     }
+    if (!this._geolocationButton){
+      this._geolocationButton = M.geolocationButton({
+        showPopup: false,
+        strings: {
+          title: "Locate"
```
prushforth, line 361: Don't forget to add localized strings to the mapml-extension.  "Locate" should be: "Show my location"

src/mapml-viewer.js
```diff
@@ -354,6 +354,18 @@ export class MapViewer extends HTMLElement {
...
+        },
+        position: "bottomright",
+        locateOptions: {
+          maxZoom: 16
```
prushforth, line 365: maxZoom should be 24 or so, I think, depending on the projection.

src/web-map.js
```diff
@@ -397,6 +397,18 @@ export class WebMap extends HTMLMapElement {
       totalSize += 49;
       this._fullScreenControl = M.fullscreenButton().addTo(this._map);
     }
+    if (!this._geolocationButton){
```
prushforth, line 400: See comments for same block in mapml-viewer

prushforth (changes requested):
We need to add the `locate` API method and associated events, to `<mapml-viewer>` and `<map is=web-map>`

src/mapml-viewer.js
Jacky0299: I made geolocationButton a function in the geolocationButton.js file

Gruntfile.js
Jacky0299: It somehow changes the how the icon looks

test/e2e/api/locateApi.test.js
```diff
@@ -0,0 +1,57 @@
...
+  });
+  test.afterAll(async function () {
+    await context.close();
+  });
+
+  test("Using locate API and locate button to find myself", async () => {
+    await context.grantPermissions(['geolocation']);
+    await page.$eval("body > mapml-viewer",(viewer) => viewer.locate());
+
+    let locateAPI_lat = await page.$eval("body > mapml-viewer", (viewer) => viewer.lat);
```
prushforth, line 25: locateAPI_* are never tested

test/e2e/api/locateApi.test.js
prushforth, line 1: I think you should separate the API test from the UI / control test, into different folders.  You should also test for `maplocationfound` and `maplocationerror` if possible, and test that the properties meet expectations.

prushforth:
The control has 3 states: off (black) tracking (orange) and last known location (blue).  These visual states need to be available to screen readers, so that when the focus lands on the control, it reads out "Show my location control - location tracking mode" (for example). "Show my location - last known location mode" and "Show my location - location off".

prushforth:
@Jacky0299  you're working on this, is that correct?

> The control has 3 states: off (black) tracking (orange) and last known location (blue). These visual states need to be available to screen readers, so that when the focus lands on the control, it reads out "Show my location control - location tracking mode" (for example). "Show my location - last known location mode" and "Show my location - location off".

Jacky0299:
> @Jacky0299 you're working on this, is that correct?
> 
> > The control has 3 states: off (black) tracking (orange) and last known location (blue). These visual states need to be available to screen readers, so that when the focus lands on the control, it reads out "Show my location control - location tracking mode" (for example). "Show my location - last known location mode" and "Show my location - location off".

Jacky0299:
> @Jacky0299 you're working on this, is that correct?
> 
> > The control has 3 states: off (black) tracking (orange) and last known location (blue). These visual states need to be available to screen readers, so that when the focus lands on the control, it reads out "Show my location control - location tracking mode" (for example). "Show my location - last known location mode" and "Show my location - location off".

yes

prushforth:
@Jacky0299 L.Control.Locate is ending up in dist folder. It should be bundled via rollup with in leaflet.js. Checkout how the other controls manage this, use them as example of how-to.

src/mapml-viewer.js
```diff
@@ -656,6 +711,20 @@ export class MapViewer extends HTMLElement {
       }
     });
   }
+  locate(options){
```
AliyanH, line 714: Seems like the locate method does not work, if the user has interacted with the geolocation control in anyway:


https://user-images.githubusercontent.com/55751566/228655623-61120617-cc6d-470d-a220-d68c1ed5dd03.mp4

AliyanH: What is the `options` parameter? It would be good to add some comments in the code showing/linking to what the options parameter might be.

src/mapml-viewer.js
```diff
@@ -525,6 +547,39 @@ export class MapViewer extends HTMLElement {
...
+        locateControl.firstChild.title = M.options.locale.btnLocTrackOff;
+      }
+    });
+    // Configure the observer to watch for changes to the class name
+    var observerConfig = { attributes: true, attributeFilter: ['class'] };
+    observer.observe(locateControl, observerConfig);
+
+    this._map.on('locationfound',
+      function (e) {
+        this.dispatchEvent(new CustomEvent('maplocationfound', {detail:
```
AliyanH, line 567: The events are working pretty nicely!
I dont think this:

```js
this.addEventListener('maplocationfound', function(e) {
  //'Location found:', e.detail.latlng, 'Radius:', e.detail.radius
});
```
and this:
```js
this.addEventListener('locationerror', function(e) {
  //error:e.detail.error
});
```
is needed, as this is something the users will add.

src/mapml-viewer.js
```diff
@@ -525,6 +547,39 @@ export class MapViewer extends HTMLElement {
...
+          {latlng: e.latlng,     radius: e.accuracy}
```
AliyanH, line 568: any specific reason for naming this property `radius`? I think `accuracy` might be more useful! Will need to somehow document this later that this number is in meters.

prushforth (approved):
Good work Jacky!  Looks good to me LGTM!

AliyanH (approved):
Looks great @Jacky0299, great work!

Issue #507, opened 2021-08-08
Popup keyboard navigation bug


Malvoz:
https://maps4html.org/experiments/screenreader/restaurants/

Repro steps:

1. Tab to the <q>The Prescott</q> marker and press <kbd>enter</kbd> or <kbd>space</kbd> to open the popup
2. Tab twice to focus the <q>Focus Map</q> control
3. Tab backwards once to focus the <q>The Prescott</q> link inside the table
4. Popup closes

**Expectation**

On step 4. the popup should remain open and the <q>The Prescott</q> link should have focus.

Malvoz:
I'm inclined to close this out as <q>wontfix</q>, because https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/535 suggests the removal of the popup navigation controls which I believe is the source of this bug.

PR #801, opened 2023-03-15
Develop <map-feature> custom element

### LINKED ISSUE #528: Implement `<map-feature>` custom element

yhy0217:
Closes Maps4HTML/Web-Map-Custom-Element#528 
- [x] make mapFeature element interactive
-   when mapFeature is removed from DOM, the associated g and path elements are removed
-   when mapFeature (and its children elements) is re-added / updated in the DOM, the associated g and path elements are updated dynamically

- [x] provide methods and APIs as suggested in Maps4HTML/Web-Map-Custom-Element#528
-   .click(): simulate a click event by default; customizable to users
-   .focus(): simulate a :focus state; show the focus rectangle and change the current document.activeElement by default; customizable
-   .mapml2geojson(): return geojson representation of geometry
-   .extent: return feature extent
-   .zoomTo: zoom to the zoom attribute value of map-feature; if the attribute is not present but a max attribute is set, zoom to the max attribute value; otherwise, zoom to the maximum zoom level that can show the feature completely
-   .pasteFeature (method of layer- element): allow users to paste new feature to a selected layer as a string or an object

- [x] create and attach a shadow root for the layer- elements with an src attribute
-   attach shadow and migrate the elements in mapml file to shadow when a layer- element is loaded
-   when a layer- element's src attribute is changed (added/removed/updated), content in shadow will change accordingly and dynamically
-   make src take precedence over the inline content; inline content can still work as a fallback when the layer's src attribute is removed

- [x] create tests for mapFeature custom element and shadow root of layer- elements
- [x] Add "zoom to here" link in pop-up; localize, closes Maps4HTML/mapml-extension#61
- [x] add min and max attributes to define the zoom value range that feature should be visible
- [x] close Maps4HTML/Web-Map-Custom-Element#803
- [x] close Maps4HTML/Web-Map-Custom-Element#807
- [x] close Maps4HTML/Web-Map-Custom-Element#507


src/map-feature.js
```diff
@@ -2,36 +2,281 @@ export class MapFeature extends HTMLElement {
...
+            this._featureLayer.closePopup();
+          } else {
+            this._featureLayer.openPopup();
+          }
+        }
+      }
+    }
+    
+    focus(event) {
+      let g = this._groupEl,
```
AliyanH, line 250: for the focus method is setting a focus on this (this.groupEl) enough? i.e. g.focus().

I was testing this method on a layer with multiple features, and checking if the right feature was being focused using:

`document.activeElement.shadowRoot.activeElement`

and It seemed to be returning a different feature? Will investigate further.

AliyanH (commented):
Just exploring all the different functionality you added so far, and a few comments about them. Looking great so far!

src/map-feature.js
```diff
@@ -2,36 +2,281 @@ export class MapFeature extends HTMLElement {
...
+        let j = json[count] = {};
+        let el = this.querySelector('map-properties');
+        // transform to gcrs
+        let source = null, dest = null, transform = false;
+        if (this._map.options.crs.code !== "EPSG:3857" || this._map.options.crs.code  !== "EPSG:4326") {
+          source = new proj4.Proj(this._map.options.crs.code);
+          dest = new proj4.Proj('EPSG:4326');
+          transform = true;
+        }
+        j.geometry = M._geometry2geojson(shapes[id]._markup, source, dest, transform);
```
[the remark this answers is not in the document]
yhy0217: Yes the _shapes[id]._markup_ represents the child element of <map-geometry> I believe.

- shapes: it is pointing to _this._featureLayer._layers_, which stores the leaflet layer representation of <map-geometry>'s children (<map-point>, <map-polygon> etc).
- shapes._markup: it is pointing to the mapml elements associated with the leaflet layers

image
(_temp1_ represents the <map-feature> element)

src/map-feature.js
```diff
@@ -2,36 +2,281 @@ export class MapFeature extends HTMLElement {
...
+          bboxExtent = [Infinity, Infinity, Number.NEGATIVE_INFINITY, Number.NEGATIVE_INFINITY];
+      for (let shape of shapes) {
+        let coord = shape.querySelectorAll('map-coordinates');
+        for (let i = 0; i < coord.length; ++i) {
+          _updateExtent(shape, coord[i]);
+        }
+      }
+      let topLeft = L.point(bboxExtent[0], bboxExtent[1]);
+      let bottomRight = L.point(bboxExtent[2], bboxExtent[3]);
+      let pcrsBound = M.boundsToPCRSBounds(L.bounds(topLeft, bottomRight), map.getZoom(), map.options.projection, cs);
```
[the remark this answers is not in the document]
yhy0217: I added a .trim() in the _getFeatureExtent method of the map-feature element to solve this problem "locally" in map-feature.js; but it doesn't really touch the nature of the problem I think

src/map-feature.js
yhy0217: Thanks! I just added g.focus() in the newest commit. However, since the current focus method also sets the CSS style of g elements, making them highlighted in the map-pane (as if the mouse were hovering over the g element), and registers an event handler (_removeFocusState) to remove the .focus() CSS style at a specific point, I'm wondering if we still need these functionalities?

src/map-feature.js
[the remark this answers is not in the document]
AliyanH: Thanks!

src/map-feature.js
```diff
@@ -2,36 +2,281 @@ export class MapFeature extends HTMLElement {
...
+            }
+            break;
+          default:
+            break;
+        }
+      }
+    }
+
+    // a .click() method that highlights the feature and show popup on map as if doing a click
+    click(event) {
```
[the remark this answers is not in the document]
AliyanH: Works great now!

src/map-feature.js
AliyanH: Take a look at how focus is done when using the arrow keys to go through different features:

https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/main/src/mapml/features/featureGroup.js#L137-L156

prushforth:
This is looking very good.  There's a bug encountered on the GeoJSON expermiments page when I copy a build into it locally:

http://localhost:8383/experiments/api/geojson/

```console
Uncaught TypeError: Cannot read properties of undefined (reading 'clone')
    at H.transform (leaflet-src.js:1827:34)
    at i._convertRing (mapml.js:6201:53)
    at i._project (mapml.js:6174:29)
    at i._reset (leaflet-src.js:8182:10)
    at i.onAdd (leaflet-src.js:8128:10)
    at i._layerAdd (leaflet-src.js:6845:10)
    at i.whenReady (leaflet-src.js:4590:15)
    at i.addLayer (leaflet-src.js:6903:10)
    at i.eachLayer (leaflet-src.js:7125:13)
    at i.onAdd (leaflet-src.js:7109:10)
```

src/layer.js
```diff
@@ -148,6 +155,27 @@ export class MapLayer extends HTMLElement {
           this.opacity = newValue;
         }
       break;
+      case 'src':
```
prushforth, line 158: In the past, we've said that src takes precedence over content; here, if a value is provided for src where there wasn't one before and a MapMLLayer had / has already been created or exists, you create a shadow root. OK, so this means the layer is being transformed to a 'remote' one, and existing inline content is going to be ignored or removed, I believe.  Is that what you're shooting for here?  If so, I like it, but it also makes me wonder if we have thought out the ramifications.

src/layer.js
```diff
@@ -148,6 +155,27 @@ export class MapLayer extends HTMLElement {
...
+          // re-load the layer element when the src attribute is changed
+          let oldOpacity = this.opacity;
+          // go through the same sequence as if the layer had been removed from
+          // the DOM and re-attached with a new URL source.
+          this.disconnectedCallback();
+          var base = this.baseURI ? this.baseURI : document.baseURI;
+          this._layer = M.mapMLLayer(newValue ? (new URL(newValue, base)).href: null, this);
+          this._layer.on('extentload', this._onLayerExtentLoad, this);
+          this._setUpEvents();
+          if (this.parentNode) {
```
prushforth, line 174: perhaps this.isConnected ?  I don't know if there might be a this.parentNode and the whole tree is not connected.

src/layer.js
prushforth: There are times when the src attribute is (re-)set by the MapMLLayer code, for example, when there is an alternate style selected by the user, or where there is a `<link rel=alternate projection="_a match for the current map's projection">`, so the client engages in "content selection" (for lack of a term).  There may be other use cases too, especially in the future for `<link rel=alternate media="a media query for text-optimized content (features)">`, that @AliyanH is looking into.  So it's already happening, and it makes sense to support it in the custom element `<layer->`.  Don't know if that rises to the level of useful, but fyi.

prushforth, line 158: Now that you've added a src attributeChangedCallback case, you'll need to move the side effects of doing `<layer->.src = foo` into the attributeChangedCallback case, or make a _private method that encapsulates those side effects, and invoke it from the attributeChangedCallback case for src.  To me that means potentially what's in the src setter in this block

```javascript
if (this._layer) {
  let oldOpacity = this.opacity;
  // go through the same sequence as if the layer had been removed from
  // the DOM and re-attached with a new URL source.
  this.disconnectedCallback();
  var base = this.baseURI ? this.baseURI : document.baseURI;
  this._layer = M.mapMLLayer(val ? (new URL(val, base)).href: null, this);
  this._layer.on('extentload', this._onLayerExtentLoad, this);
  this._setUpEvents();
  if (this.parentNode) {
    this.connectedCallback();
  }
  this.opacity = oldOpacity;
}
```

src/mapml.css
```diff
@@ -759,12 +759,14 @@ label.mapml-layer-item-toggle {
   [role="link"]:focus,
   [role="link"]:hover,
   [role="link"]:focus path,
+  [role="link"].focus path,
```
prushforth, line 762: redundant

src/layer.js
yhy0217: Yes it is definitely useful! So for now we can keep it, and the codes should be modified to ensure the newly-added src attribute can overwrite the effect of original inline content, right? I think it is better to keep the inline content instead of just removing it as it might be useful later. WDYT?

prushforth: I agree.  The inline content is / can be a fallback.

src/map-feature.js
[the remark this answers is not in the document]
AliyanH: Another thing that needs to be handled is when the feature has no `<map-properties>`, this case is throwing an error right now.

prushforth:
Seems like this will close https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/807

package.json
```diff
@@ -63,5 +63,8 @@
   "files": [
     "dist",
     "*.md"
-  ]
+  ],
+  "dependencies": {
+    "cors": "^2.8.5"
```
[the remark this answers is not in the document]
yhy0217: Not sure but I think Peter just used this for testing, so it will be removed before this PR gets merged

src/layer.js
```diff
@@ -148,7 +141,24 @@ export class MapLayer extends HTMLElement {
...
+          // but has NO EFFECT and works as a fallback
+        }
+    }
+  }
+  // re-load the layer element when the src attribute is changed
+  _reload() {
+    let oldOpacity = this.opacity;
+    // go through the same sequence as if the layer had been removed from
+    // the DOM and re-attached with a new URL source.
+    this.disconnectedCallback();
```
prushforth, line 157: I think it's kind of odd to call disconnectedCallback ourselves.  This should be done by the browser; maybe what we need is a _privateMethodThatImpliesTheProcessingThatOccursDuringDisconnectedCallback, and call that from disconnectedCallback, and here.

src/layer.js
```diff
@@ -148,7 +141,24 @@ export class MapLayer extends HTMLElement {
...
+    if (this.isConnected) {
+      this.connectedCallback();
```
prushforth, line 159: Same comment as for disconnectedCallback.

package.json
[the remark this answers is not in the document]
prushforth: I reverted both commits that are not necessary for this branch.

src/layer.js
```diff
@@ -327,46 +347,22 @@ export class MapLayer extends HTMLElement {
...
-    if(zOffset === 1 && newZoom - 1 >= 0) newZoom--;
-    map.setView(center, newZoom, {animate: false});
+    let maxZoom = this.extent.zoom.maxZoom, 
+        minZoom = this.extent.zoom.minZoom;
+    map.setView(center, M.getMaxZoom(layerBounds, map, minZoom, maxZoom), {animate: false});
   }
   mapml2geojson(options = {}){
     return M.mapml2geojson(this, options);
   }
+  _pasteFeature(text) {
```
AliyanH, line 362: when pasting features to a layer using this method or pasting through the console, using layer.zoomTo() still zooms to the old bbox, maybe that stuff should be updated when new features are pasted in.

i.e.
```js
let feature = "<map-feature><map-featurecaption>Ottawa</map-featurecaption><map-geometry><map-point class='ottawa'><map-coordinates>-75.697193 45.421530</map-coordinates></map-point></map-geometry><map-properties><table><thead><tr><th role='columnheader' scope='col'>Property name</th><th role='columnheader' scope='col'>Property value</th></tr></thead><tbody><tr><th scope='row'>Name</th><td itemprop='amenity'>Ottawa</td></tr><tr><th scope='row'>Type</th><td itemprop='name'>City</td></tr><tr><th scope='row'>Website</th><td itemprop='website'><a href='https://ottawa.ca/' target='_blank'>Ottawa</a></td></tr></tbody></table></map-properties></map-feature>"

layer._pasteFeature(feature);
```

P.S. we should rename this so it's not a 'private' method.

AliyanH: I think this should also allow pasting of not only text, but also map-feature elements.

prushforth: When editing the coordinates of a feature in light DOM, the `<map-feature>` element disappears, but the marker does move.

src/mapml/layers/MapMLLayer.js
```diff
@@ -144,6 +145,18 @@ export var MapMLLayer = L.Layer.extend({
                   var c = document.createElement('div');
                   c.classList.add("mapml-popup-content");
                   c.insertAdjacentHTML('afterbegin', properties.innerHTML);
+                  c.insertAdjacentHTML('beforeend', `<a href="" class="zoomLink">Zoom to here</a>`);
```
AliyanH, line 148: Found a bug - Pressing shift+tab when focusing back to the "Zoom to here" link in the popup -> dismisses the popup. I believe this is related to #507.

AliyanH: Can also localize this in a separate issue #821, after this is merged

AliyanH: I think the links may be getting cached or something of that nature. Does not function properly when there are two layers. 

For ex. paste the code below to index.html and zoom to the geometrycollection layer. Zooming to the point layer now zooms to the geometrycollection. If you refresh the page, and zoom the point first, it will get 'cached' or so i think.

```html
<mapml-viewer projection="OSMTILE" zoom="10" lon="-75.7" lat="45.4" controls>
      <layer- label="OpenStreetMap" src="https://maps4html.org/web-map-doc/demo/data/osm.mapml" checked></layer->
      <layer- label="Point Geometry" checked>
        <map-meta name="projection" content="OSMTILE"></map-meta>
        <map-feature>
          <map-featurecaption>Point</map-featurecaption>
          <!-- Setting the geometry coordinates to gcrs, as the default is pcrs -->
          <map-geometry cs="gcrs">
            <map-point class="point">
              <map-coordinates>-75.4916809 45.4186964</map-coordinates>
            </map-point>
          </map-geometry>
          <map-properties><h2>This is a Point</h2></map-properties>
        </map-feature>
      </layer->
      <layer- label="Geometry Collection" checked>
    <map-meta name="projection" content="OSMTILE"></map-meta>
    <map-feature>
      <map-featurecaption>Geometry Collection</map-featurecaption>
      <!-- Setting the geometry coordinates to gcrs, as the default is pcrs -->
      <map-geometry cs="gcrs">
        <map-geometrycollection>
          <map-polygon class="polygon">
            <map-coordinates>-75.5859375 45.4656690 -75.6813812 45.4533876 -75.6961441 45.4239978 -75.7249832 45.4083331 -75.7792282 45.3772317 -75.7534790 45.3294614 -75.5831909 45.3815724 -75.6024170 45.4273712 -75.5673981 45.4639834 -75.5859375 45.4656690</map-coordinates>
          </map-polygon>
          <map-linestring class="line">
            <map-coordinates>-75.6168365 45.471929 -75.6855011 45.458445 -75.7016373 45.4391764 -75.7030106 45.4259255 -75.7236099 45.4208652 -75.7565689 45.4117074 -75.7833481 45.384225 -75.8197403 45.3714435 -75.8516693 45.377714</map-coordinates>
          </map-linestring>
          <map-point class="point">
            <map-coordinates>-75.6916809 45.4186964</map-coordinates>
          </map-point>
        </map-geometrycollection>
      </map-geometry>
      <map-properties><h2>This is a Geometry Collection</h2></map-properties>
    </map-feature>
  </layer->
    </mapml-viewer>
```

prushforth:
Bug report: if you load the index.html with twohats guy, and change his coordinates to 0 0 (any value that will place him out of the viewport will do it), then pan over to him and click on his placemark, you get this in the console:

```console
mapml.js:7168 Uncaught TypeError: Cannot read properties of undefined (reading 'path')
    at i._handleFocus (mapml.js:7168:51)
    at SVGGElement.o (leaflet-src.js:2799:15)
```

yhy0217:
> ```
> mapml.js:7168 Uncaught TypeError: Cannot read properties of undefined (reading 'path')
>     at i._handleFocus (mapml.js:7168:51)
>     at SVGGElement.o (leaflet-src.js:2799:15)
> ```
I got this error message on the main branch too, and here is another one I got:
```
Uncaught TypeError: L.control.locate is not a function
    at i.onAdd (mapml.js:6192:38)
    at i.addTo (leaflet-src.js:4929:44)
    at MapViewer._createControls (mapml-viewer.js:369:55)
    at MapViewer._createMap (mapml-viewer.js:269:12)
    at MapViewer.connectedCallback (mapml-viewer.js:166:12)
    at mapml-viewer.js:1035:23
```

AliyanH:
>

AliyanH:
> and here is another one I got:

This one is likely because you need to run "npm i" before.

prushforth:
@yhy0217 this is looking really solid today.   I noticed that when rendering features via the extension (if you put a build of this branch in the extension), you can't visually delete features that are [extension-rendered](https://maps4html.org/experiments/shared/restaurants/restaurants.xml) from the layer-.shadowRoot (within the console).  This is likely an issue with the extension, but maybe you could have a look and comment, if you see anything.

image

src/mapml/features/featureGroup.js
```diff
@@ -53,12 +57,35 @@ export var FeatureGroup = L.FeatureGroup.extend({
...
+    return true;
+  },
+
   /**
    * Handler for focus events
    * @param {L.DOMEvent} e - Event that occurred
    * @private
    */
   _handleFocus: function(e) {
+    if (!(e instanceof KeyboardEvent)) return;
```
AliyanH, line 88:
```suggestion
```

src/mapml/features/featureGroup.js
```diff
@@ -26,9 +26,13 @@ export var FeatureGroup = L.FeatureGroup.extend({
         this.off("click", this._openPopup);
       }
     }
-
+    
+    L.DomEvent.on(this.options.group, "keyup keydown mousedown", this._handleFocus, this);
```
AliyanH, line 30:
```suggestion
    L.DomEvent.on(this.options.group, "keyup keydown", this._handleFocus, this);
```

AliyanH (commented):
We can remove mousedown event from the feature group, instead of omitting mouse event on the _handleFocus method. If I understand correctly _handleFocus is only handling keyboard events and nothing related to mouse events?

yhy0217:
> If I understand correctly _handleFocus is only handling keyboard events and nothing related to mouse events?

That's what I thought too.

> We can remove mousedown event from the feature group, instead of omitting mouse event on the _handleFocus method.

Yes, I believe that it could be an alternative to fix the bug, and it seems more reasonable than handling the event type inside the _handleFocus function as in the previous commit. Thanks!

src/mapml/features/featureGroup.js
yhy0217: > If I understand correctly _handleFocus is only handling keyboard events and nothing related to mouse events?

That's what I thought too.

> We can remove mousedown event from the feature group, instead of omitting mouse event on the _handleFocus method.

Yes, I believe that it could be an alternative to fix the bug, and it seems more reasonable than handling the event type inside the _handleFocus function as in the previous commit. Thanks!

AliyanH: great, thanks!

src/mapml/layers/FeatureLayer.js
```diff
@@ -144,26 +143,38 @@ export var FeatureLayer = L.FeatureGroup.extend({
...
+            if (checkRender) {
+              this.addLayer(feature);
+            }
+          }
         }
       }
     },
 
+    // substituted with the _checkRender function in FeatureGroup.js
     _clampZoom : function(zoom){
```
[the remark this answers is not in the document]
yhy0217: The code is no longer needed for the latest changes. I have kept it there and plan to remove it when the PR is ready to go, just to prevent any possible side effects. Alternatively, I could comment it out now as a more cautious approach. What do you think would be the best course of action?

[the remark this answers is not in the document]
AliyanH: I think it would be good to finalize it now, just so there are no side effects we may encounter when we are merging! Always good to remove unnecessary code!

prushforth:
This is spectacular work.  I think you're 98% done, let's discuss in the morning.  The only challenge that I can see is making the "Zoom to here" link DRY; if we could find a way to do that I would be VERY happy.  I haven't had a chance to use / test the onfocus, onblur and onclick handler methods or event handlers yet.  I would  like to set those up as an experiment for future reference, lest we forget they exist.  I'll continue with the documentation updates.

PR #804, opened 2023-03-22
Copy location as <map-feature>

### LINKED ISSUE #779: "Copy > Location" should create a <map-feature> on your clipboard
Something like:

```html
<map-feature>
  <map-featurecaption>Copied point at latitude: _ and longitude: _ </map-featurecaption>
  <map-geometry>
    <map-point>
        <map-coordinates>-75.7093517 45.4005727</map-coordinates>
    </map-point>
  </map-geometry>
</map-feature>  
```

When pasting, perhaps we should consider if we should sniff for mapml content and if found, wrap it in a \<layer-> so that it will get added to the map.  That might also apply to  pasting a \<map-extent>, which could create an empty layer with the specified extent / bounds.  

Ideas below, thank you.

prushforth:
- Change context menu to copy location as a map-feature.
- Change Util._pasteLayer to sniff for <map-feature, paste as a visible layer.

Closes #779

### DIFF
--- a/src/mapml/handlers/ContextMenu.js
+++ b/src/mapml/handlers/ContextMenu.js
@@ -130,6 +130,21 @@ export var ContextMenu = L.Handler.extend({
       .on(this._layerMenu, 'mousedown', L.DomEvent.stop)
       .on(this._layerMenu, 'dblclick', L.DomEvent.stop)
       .on(this._layerMenu, 'contextmenu', L.DomEvent.stop);
+      
+    this.t = document.createElement('template');
+    this.t.innerHTML = 
+     `<map-feature zoom="">
+        <map-featurecaption></map-featurecaption>
+        <map-properties>
+            <h2></h2>
+            <div style="text-align:center"></div>
+        </map-properties>
+        <map-geometry cs="">
+          <map-point>
+            <map-coordinates></map-coordinates>
+          </map-point>
+        </map-geometry>
+      </map-feature>`;
   },
 
   addHooks: function () {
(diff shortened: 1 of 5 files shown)


AliyanH:
How should we deal with the different "Coordinate system for location" options, for ex. I am not sure how helpful it would be to copy "tilematrix", "map" and "tile" as a `<map-feature>`. As pasting "map", and "tile" features causes an error, while pasting "tilematrix", displays the feature on the top left edge of the copied tile.

image

prushforth:
> How should we deal with the different "Coordinate system for location" options, for ex. I am not sure how helpful it would be to copy "tilematrix", "map" and "tile" as a `<map-feature>`. As pasting "map", and "tile" features causes an error, while pasting "tilematrix", displays the feature on the top left edge of the copied tile.

I think for now I will log a message to the console, open an issue to fix this when the `<map-feature>` PR gets merged.

prushforth:
@AliyanH do you think the behavior is ok as far as pasting a new layer each time goes, or should it paste into an existing layer?  Of course that will only work once the `<map-feature>` PR is merged...

AliyanH:
> @AliyanH do you think the behavior is ok as far as pasting a new layer each time goes, or should it paste into an existing layer? Of course that will only work once the `<map-feature>` PR is merged...

I think it makes sense to paste as a new layer with the localized label when pasting on to the viewer. Maybe in the future we can paste into layers through the layer control contextmenu?

Though the label you put for pasting features made me wonder, if we should change the default label for pasting layers/geojson with no label (to something similar: "Pasted layer"), currently it defaults to a localized string of: "Layer"
https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/main/src/mapml/utils/Util.js#L447
https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/main/src/mapml/utils/Util.js#L569

prushforth:
Depends on https://github.com/Maps4HTML/mapml-extension/pull/54 for localization.

prushforth:
@AliyanH :

> Though the label you put for pasting features made me wonder, if we should change the default label for pasting layers/geojson with no label (to something similar: "Pasted layer"), currently it defaults to a localized string of: "Layer"
https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/main/src/mapml/utils/Util.js#L447
https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/main/src/mapml/utils/Util.js#L569

Makes sense.  I may update this and related PR to do that, tbd.

prushforth:
@Malvoz specifically: Is https://github.com/Maps4HTML/mapml-extension/pull/54/commits/d8c0a85b288b3e7eed31bc39f2bed69e5485a867#diff-7b63e53a103af3709d1935ceb51bce87f09b0a54b199f3b0a0412de12f834fa6 acceptable?

PR #815, opened 2023-04-06
Malvoz attribution control

prushforth:
@Jacky0299 see if you can get the tests running. I think I got all the relevant changes reconciled with main, tbd. I think we'll want to delete one of the attribution controls once we get this working and tested.  

These commits were copied from [this PR](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/642), hopefully faithfully, tbd.


test/e2e/core/popupTabNavigation.test.js
```diff
@@ -215,7 +215,7 @@ test.describe("Playwright Keyboard Navigation + Query Layer Tests" , () => {
       const nh = await page.evaluateHandle(doc => doc.shadowRoot, h);
       const rh = await page.evaluateHandle(root => root.activeElement, nh);
       const f = await (await page.evaluateHandle(elem => elem.innerText, rh)).jsonValue();
-      expect(f).toEqual("Maps4HTML");
+      expect(f).toEqual("Leaflet");
```
AliyanH, line 218: Any reason why the order of the Maps4HTML and Leaflet attribution is being changed on the attribution control? I think we should keep the order it currently is in, with Maps4HTML attribution first and then Leaflet.

PR #824, opened 2023-04-14
Copy layer with relative path fails when pasted layer does not have the same base path

### LINKED ISSUE #817: Copy layer with relative path fails when pasted layer does not have the same base path
For example, when copying the only layer on [this map](https://maps4html.org/experiments/styling/alternate-styles/):

`<layer- src="natgeo.mapml" checked="" label="Basemap with Alternate Styles"></layer->`

can't be pasted onto a map that's not at the same base URL, because of the relative URL in the `src` attribute.  

Perhaps it could be made more universal by resolving the URL before / during copy.

kevinkim31:
closes #817

### DIFF
--- a/src/layer.js
+++ b/src/layer.js
@@ -234,6 +234,13 @@ export class MapLayer extends HTMLElement {
       }
     }, 0);
   }
+  getOuterHTML() {
+    let fullLayer = this;
+    if (fullLayer.hasAttribute('src')) {
+      fullLayer.setAttribute('src', this._layer.getHref());
+    }
+    return fullLayer.outerHTML;
+  }
   _onLayerChange() {
     if (this._layer._map) {
      // can't disable observers, have to set a flag telling it where
--- a/src/layer.js
+++ b/src/layer.js
@@ -249,19 +249,17 @@ export class MapLayer extends HTMLElement {
       mapLinks.forEach((mapLink) => {
        
         if (mapLink.hasAttribute('href')) {
-          mapLink.setAttribute('href', decodeURI(this.transformURL(mapLink.attributes.href.value, this.baseURI ? this.baseURI : document.baseURI)));
+          mapLink.setAttribute('href', decodeURI((new URL(mapLink.attributes.href.value, this.baseURI ? this.baseURI : document.baseURI)).href));
         }
         else if (mapLink.hasAttribute('tref')) {
-          mapLink.setAttribute('tref', decodeURI(this.transformURL(mapLink.attributes.tref.value, this.baseURI ? this.baseURI : document.baseURI)));
+          mapLink.setAttribute('tref', decodeURI((new URL(mapLink.attributes.tref.value, this.baseURI ? this.baseURI : document.baseURI)).href));
         }
       });
     }
 
     return tempElement.innerHTML;
   }
-  transformURL(url, baseUrl) {
-    return (new URL(url, baseUrl)).href;
-  }
+
   _onLayerChange() {
     if (this._layer._map) {
      // can't disable observers, have to set a flag telling it where
--- a/src/layer.js
+++ b/src/layer.js
@@ -235,11 +235,13 @@ export class MapLayer extends HTMLElement {
     }, 0);
   }
   getOuterHTML() {
-    let fullLayer = this;
-    if (fullLayer.hasAttribute('src')) {
-      fullLayer.setAttribute('src', this._layer.getHref());
+    let fullLayer = this.outerHTML;
+    if (this.hasAttribute('src')) {
+      let findSrc = /src="([^"]*)"/;
+      let newSrc = this._layer.getHref();
+      fullLayer = fullLayer.replace(findSrc, `src="${newSrc}"`);
     }
-    return fullLayer.outerHTML;
+    return fullLayer;
   }
   _onLayerChange() {
     if (this._layer._map) {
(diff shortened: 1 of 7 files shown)


src/layer.js
```diff
@@ -234,6 +234,13 @@ export class MapLayer extends HTMLElement {
       }
     }, 0);
   }
+  getOuterHTML() {
+    let fullLayer = this;
+    if (fullLayer.hasAttribute('src')) {
+      fullLayer.setAttribute('src', this._layer.getHref());
+    }
+    return fullLayer.outerHTML;
+  }
```
AliyanH, line 243: `fullLayer` is just a reference to `this` (which is the `layer-` element in this case) so all of the changes you make to `fullLayer`, the same changes are also applied to `this`. This is why the layer element's src attribute gets updated when you copy that layer. 

You should not manipulate elements to get the desired results but instead can simply just manipulate a string, as that is what this method should return.

src/layer.js
```diff
@@ -234,6 +234,34 @@ export class MapLayer extends HTMLElement {
...
+        }
+        else if (mapLink.hasAttribute('tref')) {
+          mapLink.setAttribute('tref', decodeURI(this.transformURL(mapLink.attributes.tref.value, this.baseURI ? this.baseURI : document.baseURI)));
+        }
+      });
+    }
+
+    return tempElement.innerHTML;
+  }
+  transformURL(url, baseUrl) {
```
prushforth, line 262: This member function doesn't seem necessary, tbh, it's just wrapping `new URL(url, base)).href` ; anyway the scope is public which is too much.  The scopy of it could even be **within** `getOuterHTML`, since it's defined for use by that function alone.  But, I would personally not define a function here, and more simply use `new URL(...).href`

test/e2e/core/layerContextMenu.html
```diff
@@ -24,6 +24,7 @@
 <body>
   <mapml-viewer style="width: 500px;height: 500px;" projection="CBMTILE" zoom="0" lat="47" lon="-92" controls>
     <layer- label="CBMT - INLINE" checked>
+      <map-link rel="license" title="Copyright FooBar Inc."></map-link>
```
prushforth, line 27: Seems odd tbh

src/layer.js
```diff
@@ -234,6 +234,34 @@ export class MapLayer extends HTMLElement {
...
+    if (this.querySelector('map-link')) {
+      let mapLinks = tempElement.querySelectorAll('map-link');
+
+      mapLinks.forEach((mapLink) => {
+       
+        if (mapLink.hasAttribute('href')) {
+          mapLink.setAttribute('href', decodeURI(this.transformURL(mapLink.attributes.href.value, this.baseURI ? this.baseURI : document.baseURI)));
+        }
+        else if (mapLink.hasAttribute('tref')) {
+          mapLink.setAttribute('tref', decodeURI(this.transformURL(mapLink.attributes.tref.value, this.baseURI ? this.baseURI : document.baseURI)));
```
prushforth, line 255: I wasn't certain that `new URL(...)` would work with a `tref` value as input, due to the presence of `{..}` 

How sure are we that the base we're resolving against is correct? i.e. what is the difference between `this.baseURI` and `document.baseURI`?

kevinkim31: I think the document.baseURL is just a last case scenario like mentioned in _ready()

src/layer.js
```diff
@@ -234,6 +234,32 @@ export class MapLayer extends HTMLElement {
       }
     }, 0);
   }
+  getOuterHTML() {
+    let fullLayer = this.outerHTML;
+    let tempElement = document.createElement('div');
+    tempElement.innerHTML = fullLayer;
```
AliyanH, line 240: you can use `this.cloneNode(true)` and delete the cloned element at the end of the method, instead of doing this.

test/e2e/core/layerContextMenu.test.js
```diff
@@ -40,7 +40,7 @@ test.describe("Playwright Layer Context Menu Tests", () => {
       (text) => text.value
     );
     
-    expect(copyLayer).toEqual("<layer- label=\"CBMT - INLINE\" checked=\"\">\n      <map-extent units=\"CBMTILE\" hidden=\"\">\n        <map-input name=\"zoomLevel\" type=\"zoom\" value=\"3\" min=\"0\" max=\"3\"></map-input>\n        <map-input name=\"row\" type=\"location\" axis=\"row\" units=\"tilematrix\" min=\"14\" max=\"21\"></map-input>\n        <map-input name=\"col\" type=\"location\" axis=\"column\" units=\"tilematrix\" min=\"14\" max=\"19\"></map-input>\n        <map-link rel=\"tile\" tref=\"/data/cbmt/{zoomLevel}/c{col}_r{row}.png\"></map-link>\n      </map-extent>\n    </layer->");
```
AliyanH, line 43: adding a test for a `<layer- label="..." src="..."></layer->` would be beneficial.

AliyanH (changes requested):
LGTM

prushforth:
I'm concerned with the behaviour of this when installed in experiments, especially for the scroll-to-zoom-out behaviour of [this experiment](https://maps4html.org/experiments/change-projection/zoomin-zoomout/mapml-viewer/).

prushforth:
When I copy an unchecked Restaurants layer [from here](https://maps4html.org/experiments/screenreader/restaurants/), and [paste it onto the map here](https://maps4html.org/experiments/api/geojson/), it is intially checked but not displayed.  I believe it should not be checked.  This is a bug in some other code tbd, not a concern here, as the tref in the underlying map-extent is correctly resolved.

prushforth:
LGTM

AliyanH (approved):

PR #827, opened 2023-04-19
Fix error that occurs when removing and reattaching map from document

### LINKED ISSUE #813: Detaching then reattaching <mapml-viewer> fails
If you try to remove the viewer from the document, and then re-attach it to the DOM, you get an error that results from the element lifecycle callback trying to re-create the shadow root. 

image

prushforth:
Closes #813


prushforth:
This is a bit tricky to test; couldn't figure out how to make it fail on main and pass on this branch, because it throws in the browser on main, which upsets subsequent locators, waiting seemingly forever for events that don't happen due to the earlier thrown error.

web-map-index.html
```diff
@@ -4,7 +4,7 @@
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width,initial-scale=1">
     <title>index-web-map.html</title>
-    <script type="module" src="dist/web-map.js"></script>
```
AliyanH, line 7: This does not work without "dist/"

AliyanH (approved):
LGTM!

prushforth:
> LGTM!

Thanks for your help! Merging...

PR #830, opened 2023-04-21
arrow keys to navigate contextmenu

Jacky0299:


src/mapml/handlers/ContextMenu.js
```diff
@@ -723,6 +770,85 @@ export var ContextMenu = L.Handler.extend({
         L.DomEvent.stop(e);
         this._focusOnLayerControl();
       }
+    } else if (key === 38) { //up arrow
+      if (!this._coordMenu.hasAttribute('hidden')&&document.activeElement.shadowRoot.activeElement.innerHTML === 'Map') {
```
AliyanH, line 774: This will cause issues when the contextmenu is in a different language, through the extension.

There is probably a better way to this, take a look at how `_layerMenuTabs` works.

src/mapml/handlers/ContextMenu.js
```diff
@@ -723,6 +770,85 @@ export var ContextMenu = L.Handler.extend({
         L.DomEvent.stop(e);
         this._focusOnLayerControl();
       }
+    } else if (key === 38) { //up arrow
+      if (!this._coordMenu.hasAttribute('hidden')&&document.activeElement.shadowRoot.activeElement.innerHTML === this._coordMenu.children[0].innerHTML) {
```
prushforth, line 774: Better to continue this on two or more lines than remove spaces to fit it all.  Good job finding something to compare it to, but maybe add what you're looking for as a trailing comment e.g. `// map` etc.

src/mapml/handlers/ContextMenu.js
```diff
@@ -723,6 +770,85 @@ export var ContextMenu = L.Handler.extend({
...
+        this._coordMenu.children[2].focus();
+      } else if (!this._coordMenu.hasAttribute('hidden')&&document.activeElement.shadowRoot.activeElement.innerHTML === this._coordMenu.children[1].innerHTML) {
```
prushforth, line 776: e.g.

```javascript
   } else if (!this._coordMenu.hasAttribute('hidden') && 
     document.activeElement.shadowRoot.activeElement.innerHTML === this._coordMenu.children[1].innerHTML) {
```

AliyanH (commented):
When the "toggle controls" item is disabled (by starting a map with no controls), it requires me to press the arrow key twice to go to the next item, instead of 1. Also encountered a call stack warning while going through the contextmenu.

image

prushforth:
@Malvoz we have an a11y issue we need an expert opinion on: Can we neglect use of tab keys to navigate the context menu?  Chrome context menu itself does not seem to respond to tab keys, but maybe we have got the "button" semantic in place so we need to maintain tab key support.  WDYT?

prushforth:
Small visual difference noted when you pull out the Copy > submenu, it puts a focus ring on the first item; generally the focus ring is only shown when using keyboard. 

image

Unlike the hover behaviour, the focus ring doesn't move when you move your pointer, so I believe the focus ring should not be shown in this case (when using the pointer only):
image
image

prushforth:
Awesome work Jacky 👍👏

src/mapml/handlers/ContextMenu.js
```diff
@@ -723,13 +775,135 @@ export var ContextMenu = L.Handler.extend({
         L.DomEvent.stop(e);
         this._focusOnLayerControl();
       }
+    } else if (key === 38) { //up arrow
+      if (!this._coordMenu.hasAttribute('hidden') && 
+        (document.activeElement.shadowRoot == null || //null happens when the focus is on submenu and when mouse hovers on main menu, submenu disappears
```
prushforth, line 780: You should change all the "==" operators to "===", they are quite different! 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality
vs
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Equality

src/mapml/handlers/ContextMenu.js
```diff
@@ -723,13 +775,135 @@ export var ContextMenu = L.Handler.extend({
...
+            nextIndex++;
+            if (nextIndex >= this._items.length) {
+              nextIndex = 0;
+            }
+          }
+          this._setActiveItem(nextIndex);
+        }
+      }
+    } else if (key === 39) { //right arrow
+      if (document.activeElement.shadowRoot == null) { //do nothing when null
```
prushforth, line 852: This looks funny.  I would combine the negative logic with the logic in your else and eliminate the do-nothing.  The comment should also say **why** shadowRoot is null.

src/mapml/handlers/ContextMenu.js
prushforth: I fixed it, because the test with the Escape key wasn't working due to the == operator

src/mapml/handlers/ContextMenu.js
```diff
@@ -723,13 +775,157 @@ export var ContextMenu = L.Handler.extend({
...
+          document.activeElement.shadowRoot.activeElement.innerHTML === this._coordMenu.children[2].innerHTML){
+            this._coordMenu.setAttribute('hidden','');
+            this._setActiveItem(5);
+          }
+        } else {
+          this._hide();
+        }
+      }
+    } else if (e.shiftKey && key === 9){
+      let arr = ['Back (<kbd>Alt+Left Arrow</kbd>)', 'Forward (<kbd>Alt+Right Arrow</kbd>)', 'Reload (<kbd>Ctrl+R</kbd>)',
```
prushforth, line 889: Looks like the localization is ignored here.  Let's undo these changes and log the tab issue as to be completed in the future, no need to rush it.

PR #831, opened 2023-04-25
Add "Zoom to here" link in popup

### LINKED ISSUE #778: Add a link in the popup for a feature for "Zoom to here"
The link would have to either be a graphic or a bit of internationalized text.  The link would only be presented if the an extent for the feature can be calculated.  Point geometries would have to have some heuristic applied, TBD.

Something like this might work:

image

yhy0217:
Closes #778 
- [x] create shadowRoot for templated features and queries
- [x] add functional "zoom to here" links in popups of inline-static / templated / queried features
- [x] fix bugs in MapFeature, add tests for MapFeature.addEventListener
- [x] add tests for zoomTo link


prushforth:
@yhy0217 I put a build of this into GeoServer to test it against queryable layers, and I get the following on a query that should return a `<map-feature>` with a non-null `<map-geometry>`:

image

prushforth:
> the ._extentEl is null

was mistaken, it's the shadowRoot property of ._extentEl that is null

src/mapml/handlers/QueryHandler.js
```diff
@@ -88,14 +88,18 @@ export var QueryHandler = L.Handler.extend({
...
-        if(lastOne) return displayFeaturesPopup(layer._mapmlFeatures, e.latlng);
-      }).catch((err) => {
-        console.log('Looks like there was a problem. Status: ' + err.message);
+        if(lastOne) {
+          // create connection between queried <map-feature> and its parent <map-extent>
+          for (let feature of layer._mapmlFeatures) {
+            feature._extentEl = template._extentEl;
+          }
+          displayFeaturesPopup(layer._mapmlFeatures, e.latlng);
+        }
```
AliyanH, line 102: I think you may have accidently removed catch error block.

src/mapml/layers/FeatureLayer.js
```diff
@@ -5,7 +5,12 @@ export var FeatureLayer = L.FeatureGroup.extend({
    * M.MapML turns any MapML feature data into a Leaflet layer. Based on L.GeoJSON.
    */
     initialize: function (mapml, options) {
-
+      /*
+        mapml:
```
AliyanH, line 9: Excellent work adding comments! It was a delight going over the code. Very helpful for new contributors or more importantly, for people revisiting the code and actually understand what's happening! 😄

prushforth:
Closes https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/778

prushforth:
This is really working well now - the custom projection query is even working on the shrink/swell layer - reveals a "Zoom to here" link that works - thanks for working on this, too, @AliyanH.  Also, I noticed that the extension's "Zoom to here" link is working correctly now, for example when rendering a GeoServer WFS layer, the zoom to here link is correct.  Overall, this is excellent work.  I'll keep reviewing, but I think it's ready to merge.

prushforth:
Seems like the popup is hard to get rid of after clicking on the "Zoom to here" link, check the restaurants experiment.

Issue #835, opened 2023-04-28
Tabbing through Context Menu is not synced with arrow keys navigation


Jacky0299:

Jacky0299:
When tabbing on the context menu, pressing arrow keys does not have continue to focus the next item.

Jacky0299:
Layer menu bug:
Open the layer contextmenu and try "Tab", "Up Arrow Key", and then "Tab" again.

PR #836, opened 2023-04-28
Tab compatible with Arrow key solves #835

Jacky0299:
solves #835


Jacky0299:
This solves the problem of tabbing on the context menu and pressing arrow keys are not connected. Layer Menu bug still needs to be addressed

PR #841, opened 2023-05-01
map-feature bug fixes

### LINKED ISSUE #838: Removing map-feature to leave an empty layer- throws an error
When removing the only map-feature from a layer-, map-feature disconnectedCallback throws an error.

**Steps to reproduce:**

- Go to [this experiment](https://maps4html.org/experiments/api/map-feature/)
- Click "Add feature"
- Click "Remove feature"
- Console throws an error, and the feature is not completely removed, checking unchecking the layer still renders the feature.

AliyanH:
- [X] #838 
- [X] make click/focus/blur events work and in sync with the g element (can work as attribute, API, and addEventListener).
- [X] Allow prevention of default click handling, can be done as following:
```js
// $0 represents the map-feature element
$0.addEventListener("click", (e) => {
   e.originalEvent.stopPropagation();
});
```
- [X] add support for keyup/keydown event on map-feature
- [x] Add Tests

closes #838


prushforth:
I would have liked to get rid of the click(ev), focus(ev) and blur(ev) parameter, as it seems [non-standard](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click), but perhaps that's not possible because we have got leaflet handling the underlying clicks and displaying popups and so on.

src/map-feature.js
prushforth, line 1: I think that you removed the onfocus, onclick and onblur attributes, so no need to watch them here.

prushforth: ... although we could reasonably add min and max to the observedAttributes array https://maps4html.org/web-map-doc/docs/api/map-feature-api#min

src/map-feature.js
```diff
@@ -578,34 +560,34 @@ export class MapFeature extends HTMLElement {
...
+        shapes = featureGroup._layers;
+      // close popup if the popup is currently open
+      for (let id in shapes) {
+        if (shapes[id].isPopupOpen()) {
+          shapes[id].closePopup();
         }
       }
+      if (featureGroup.isPopupOpen()) {
+        featureGroup.closePopup();
+      } else if (!clickEv.originalEvent.cancelBubble) {
```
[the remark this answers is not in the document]
AliyanH: `cancelable` is only used to check if an event can be cancelled. cancelBubble is true when when the user stops Propagation.

AliyanH:
noticed another issue just now that I will work on, removing a map-feature when it is unchecked does not work.

AliyanH:
I think this is good to go @prushforth , just a few changes I added: 
- removed non-standard parameters for click, focus, blur methods (based on suggestion by @prushforth)
- Added support for keyup/keydown on map-feature

prushforth:
Would you mind fixing [this bug](https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/855) in this PR?  If that doesn't make sense, feel free to merge this, it looks and works well.

prushforth:
I like the test for preventDefault that displays a different thing than the popup for the polygon, good idea there.  It tests the API and not just our API; that's exactly what we need to do more of.

prushforth (approved):

PR #844, opened 2023-05-04
Fix context menu tabbing

### LINKED ISSUE #835: Tabbing through Context Menu is not synced with arrow keys navigation

prushforth:
- [x] merge upstream prettier commits
- [x] get existing tests working
- [x] arrow up/down should not scroll the page when moving between context menu items
- [x] shift+tab should work like ArrowUp
- [x] add tests that show tabbing works the same as arrow key navigation of context menus
- [ ] test on firefox
- [x] add tests for preventing scroll on long pages when arrow keys used to navigate context menus
- [x] add test to show no error occurs when Shift+F10 / contextmenu is pressed on a feature

Closes #835, #846


prushforth:
Hmm still the issue of scrolling the page when arrow keys are pressed, usually it's not noticeable when there's no scroll bars, but when you have scroll bars, arrow keys on a menu get transmitted to the document.  Hopefully just a question of preventDefault, tbd.  Sorry just ignore until I fix that.

AliyanH:
Just an error I noticed that is on main also, 
* Go to this page, and open the console https://maps4html.org/experiments/screenreader/restaurants/
* tab to the first feature
* press shift+f10
The following error is displayed
image

prushforth:
> Just an error I noticed that is on main also,

OK, I think I can fix that in this PR.

src/mapml/handlers/ContextMenu.js
```diff
@@ -710,8 +810,24 @@ export var ContextMenu = L.Handler.extend({
...
+        this._clickEvent.originalEvent.clientX === 0 ||
+        this._clickEvent.originalEvent.clientY === 0
+      ) {
+        const getCenter = function (el) {
+          let w = el.getBoundingClientRect().width;
+          let h = el.getBoundingClientRect().height;
+          return { x: Number.parseInt(w / 2), y: Number.parseInt(h / 2) };
+        };
+        pt = getCenter(this._map.getContainer());
+      }
```
AliyanH, line 828: on the custom-projections/BNG/ experiment on firefox, it looks like this does not work - http://127.0.0.1:5501/custom-projections/BNG/

prushforth: Weird not just custom projections, but all the experiments afaict.  I'll check it out. I meant to add an explicit test for firefox too, thanks for noticing.

AliyanH: yup looks like for all experiments! weird!

prushforth: Really points to the need for a test for firefox here, so I'll do that.

src/mapml/handlers/ContextMenu.js
```diff
@@ -1027,82 +1121,91 @@ export var ContextMenu = L.Handler.extend({
...
-            this._coordMenu.setAttribute('hidden', '');
-            this._setActiveItem(5);
+            this._copySubMenu.setAttribute('hidden', '');
+            this._setActiveItem(this._menuItems.CTXCOPY);
           }
         } else {
           this._hide();
         }
       }
+      // I don't understand the logic here, but tests break without it...
```
AliyanH, line 1182: I think this code block is hiding the contextmenu whenever the user presses enter on any of the layer contextmenu items (except for the "Copy (C)" item)

PR #854, opened 2023-05-18
Update default CS and Projection for layer

### LINKED ISSUE #851: Layer does not render without map-meta name="projection"
Ex. for the demo [here](https://maps4html.org/web-map-doc/docs/layers/layer/) when you remove the projection map-meta, the layer does not render. The fallback projection of the map should be used.
```html
<map-meta name="projection" content="OSMTILE"></map-meta>
```

### LINKED ISSUE #852: Change the default coordinates for a layer from pcrs to gcrs
Currently the coordinates for a layer- are interpreted by default as pcrs, the default should be changed to gcrs instead - aligning with our documentation (https://maps4html.org/web-map-doc/docs/elements/meta/#default-metadata)

AliyanH:
- [X] update default cs from pcrs to gcrs, closes #852 
- [x] add a default projection (parent's projection) for a layer (when projection map-meta is not provided), closes #851
  - console logs, when map-meta not present for a layer.
- [ ] ~~https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/860~~ handled in #873
  - [ ] ~~Add web-map equivalent~~
- [x] Add Testing


src/mapml/layers/FeatureLayer.js
```diff
@@ -130,7 +130,7 @@ export var FeatureLayer = L.FeatureGroup.extend({
         M._metaContentToObject(
           mapml.querySelector('map-meta[name=cs]').getAttribute('content')
         ).content) ||
-      'PCRS';
```
prushforth, line 133: This causes problems with the GeoServer MapML module, because the _getNativeVariables: function (mapml) argument is actually just the feature, not the whole document, AFAICT.  So, even though there's a `<map-meta name="cs" content="pcrs"></map-meta>` in that document, it's not actually in the `<map-feature>` element, it's somewhere above it, earlier in the document.  Given that, we either have to change the GeoServer module to emit the cs attribute on the `<map-geometry>` instead of as a <map-meta> element, or change what gets passed to this function to include the whole MapML response, or add another parameter for cs.

AliyanH: As per our discussion, I have added the map-meta's that may come from the mapml document to the shadowRoot of map-extent along with the map-feature, they are used accordingly also by the zoom to here link.

prushforth (commented):
Looks good to me, tested with geoserver too. I noticed that in mapml2geojson.html / test there are some comments like this:
```
<!-- Setting the geometry coordinates to gcrs, as the default is pcrs -->
            <map-geometry cs="gcrs">
```
so maybe we should amend that test / others that we may find that set the default CS and projection via map-meta or cs attribute to RELY on the new default values for these?  At least we could remove the comment because it's no longer true that the default is PCRS.

prushforth:
Is this ready to merge?

src/map-feature.js
```diff
@@ -246,6 +246,7 @@ export class MapFeature extends HTMLElement {
       this._layer = parentEl._layer;
       _attachedToMap();
     }
+    this._groupEl._feature = this;
```
prushforth, line 249: Actually I think I made a mistake about where this should go.  Will fix.

AliyanH:
> Is this ready to merge?

I believe so, do you think the testing is sufficient or can you think of any other edge cases? I will give it a look also.

prushforth:
> > Is this ready to merge?
> 
> I believe so, do you think the testing is sufficient or can you think of any other edge cases? I will give it a second look also.

I encountered an issue in the other branch for the setName(newName) function.  Specifically, the `this._groupEl` property had not been defined and was throwing an error.  That issue stems from the location of the statement outside of the _updateFeature() method, which is invoked either if the parent layer is defined already or in a `createmap` event handler.  The problem is that the `createmap` event handler had not been dispatched and so there was no this._layer and no shadowRoot to append to etc etc.  The `createmap` event is the source of inumerable problems and I would like to figure out how to get rid of it.

src/map-feature.js
prushforth: Can't fix it short term.  I will work with a different layer construct in the other branch, to avoid `<map-feature>`, as I'm working on `HTMLLayerElement.label` which doesn't require a dependency on `<map-feature>`

prushforth:
I think it's safe to merge.

AliyanH:
> I think it's safe to merge.

ok will merge after final review. Thanks!

src/map-feature.js
prushforth: Actually, this looks like it's going to be an issue, probably going to have to revert this and the associated test until I can figure out how to work around the whole `createmap` issue.

AliyanH: ok that makes sense, Yes the `createmap` issue is a source of quite a few issues, It is also interfering with the map-extent development, would be good to treat is separately.

AliyanH: @prushforth Should we revert this change? this PR is ready otherwise!

prushforth: Reverted.  Flaky connection today

AliyanH:
@prushforth in ca9399d747cb305218b8dc5e86138c6ff56cef68, default projection was added, and the use of 'createmap' events was removed. 

Though there is still a bug with the _processInitialExtent/_initTemplateVars, this occurs when the map is initially loading and the custom projection is not available, so the layer's extent gets calculated (_processInitialExtent/_initTemplateVars) based on the map's current projection (OSMTILE) - which throws errors (can be seen in the BNG experiment). 

I think the layer needs to go disabled at that stage and not go through that code at that time, and the layer should only be loaded once the current projection is available with the use of 'projectionchange' events emitted by mapml-viewer, which the layer will be listening for.

PR #857, opened 2023-05-26
<map-input>, <map-link>, and <map-extent> custom element development

AliyanH:
- [X] `<map-input>` custom element development
  - [X] create classes for different map-input attribute 'type' ('height', 'width', 'zoom', 'hidden', 'location')
  - [x] for each input type class, provide a `validateInput()` method that is called by the map-input's `checkValidity()` and `reportValidity()` methods
  - [x] for each input type class, provide a `getValue()` method that is used by the map-input as a property - (`<map-input>.value`)
- [ ] `<map-link>` custom element development
- [ ] `<map-extent>` custom element development

**Issues/problems derived from #873**
- [ ] move createAndAddTemplatedLayers from on add in mapmlLayer to process content similar to how we handler other layer types


AliyanH:
Continuing in #887

PR #859, opened 2023-06-01
Remove old image

### LINKED ISSUE #858: Old images pile up when panning a templated image
Sometimes, the images visually interfere with other images that are at higher z-indexes:

image

but sometimes it's not obvious that its happening:

prushforth:
Closes #858


AliyanH (approved):
LGTM, this is why testing is important - the new test works great. Nice catch!

Issue #833, opened 2023-04-26
Can't change the label of a layer


prushforth:
When you load a layer from a remote URL via the `<layer- src>` attribute, it's expected that you won't have control over the accessible name of the layer - it's set via the `<map-title>` element in the remote document.  However, if there's no `<map-title>` element in the document, you should be able to set the `label` attribute of the `<layer->` element and see that change reflected in the layer control, but it not currently that way.  If the `<layer->` has inline content, i.e. no `src` attribute, you should be able to control the accessible name of the layer in the layer control by adding/changing the `<map-title>` child element of `<layer->`, or maybe the `<layer- label>` value as well tb discussed.

Note that [changing the accessible name of a thing is not advisable](https://sarahmhigley.com/writing/playing-with-state/#name-changes).

prushforth:
The behaviour of HTMLLayerElement.label is not yet [specified](https://github.com/Maps4HTML/MapML/issues/246), but it is noted in the docs that [this needs implementation](https://maps4html.org/web-map-doc/docs/api/layer-api/#src:~:text=HTMLLayerElement.label%20provides).

Issue #864, opened 2023-06-27
Get rid of `createmap` event, replace with default projection + `changeprojection` event + handlers


prushforth:
`createmap` was a hack that is getting in the way of API development, and it affects many elements.  It's only needed because there was no default projection that could be upgraded later, so things like layer creation and validation had to be delayed.  But the resulting code is really hard to reason about, so let's see if we can improve that situation.

PR #865, opened 2023-06-27
Enable layer name change

prushforth:
Closes https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/833 

Code's a bit ugly due to `createmap`-generated [issues](https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/864).


prushforth:
Merging because I can see that the usually-reliable fire danger WMS is slow to load, causing these tests to fail. We should remove this WMS from our test suite, bake local resources into the tests instead.

PR #870, opened 2023-07-05
Feature index overlay bug

prushforth:
Supporting [PR in mapml-extension](https://github.com/Maps4HTML/mapml-extension/pull/65)

- [x] Make map the focus after following a link e.g. on [this map](https://maps4html.org/experiments/painting/)
- [x] Make map the focus after reload() etc.
- [x] Verify that checking/unchecking layers works for feature index behaviour
- [x] Add test for post-geolocation activation, deactivation
- [x] Add test for post-link activation
- [x] Make output responsive (fit between lower controls without overlap)


prushforth:
@malvoz there are significant changes in this PR, would be great if you would have a look. Also: should this be a control?

AliyanH:
moved to #873

Issue #524, opened 2021-10-05
Re-design what mapml-viewer does with LegendURL / GetLegendGraphic links


prushforth:
WMS and other OGC mapping services provide legend information via linked graphics, typically but not necessarily images.

In the original implementation of this custom element, I wasn't sure what to do with this information, but I wanted to expose it, so I included a `<a href="LegendURL">Layer name/title/label</a>` in the `<details>'` `<summary>` that represents the layer in the layer control.

It seems possible that browsers will [in future disallow such interactive content ](https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/263#issuecomment-770064367)in the `<summary>` elements, so this issue is where we can discuss and design what to do with the (set of possible) LegendURL  or GetLegendGraphic links (or equivalents) that may be available for a map layer.  Currently these can be represented by DOM elements such as 

```html
<map-link rel="legend" href="LegendURL"></map-link>
```

but the question remains as to how to expose them to the developer, if not the user, in an accessible manner.

AliyanH:
Hi @prushforth,

I played around with three possible placements for legend graphics, would love to hear your thoughts!

**Option 1 - Legend appears in the layer settings, as a disclosure widget beneath layer Opacity disclosure widget**
<img width="551" height="407" alt="Image" src="https://github.com/user-attachments/assets/a937ce2d-3721-4f5a-9de3-b41d43433a8f" />

**Option 2 - Legend appears in the layer settings, before the Opacity disclosure widget in a div**
<img width="558" height="382" alt="Image" src="https://github.com/user-attachments/assets/a81e3ece-9cde-4463-8716-cdcf876a7966" />

**Option 3 - Legend appears outside the layer settings beneath the layer title, and is displayed as soon as the layer menu is opened**
<img width="579" height="191" alt="Image" src="https://github.com/user-attachments/assets/78e97b49-2cc2-40a3-8936-b93af645105e" />

prushforth:
I think I favour the disclosure widget approach.  At some point I think we'll want to have a parameterized Legend (by style, principally)  [underneath Sub-layer](https://github.com/Maps4HTML/MapML.js/issues/1053), so a disclosure widget seems portable to that location as well.  I've been working on [a legend example for gcds-map](https://nrcan.github.io/gcds-map/en/components/map/layer/design/#legends), and I realized that opening a plain image in another tab doesn't do much for non-visual users (to understate it).  

Thanks for tackling this!

PR #873, opened 2023-07-18
Add a default projection (OSMTILE) for the map

AliyanH:
- [X] Add default projection of OSMTILE for mapml-viewer
- [X] remove use of 'createmap' event
- [x] resolve issues related with custom projection
- [x] web-map support

**Notes**
**ToDo -** 
- ~~[ ] move createAndAddTemplatedLayers from on add in mapmlLayer to process content similar to how we handler other layer types~~ Should be done as part of map-extent custom element development.
- [X] Add a projection parameter to _convertAndFormatPCRS
- [x] get rid of 'foo' event

**current known issues on experiments (need to re-verify after latest commits) -**

- [x] experiments/painting/ Girl example, clicking on girl does not zoom
- [x] experiments/alternate-style/ maintian opacity across different styles
- [x] experiments/linking/features - clicking on BC does not work, Ontario link not working
- [x] experiments/vector tiles - not working
- [x] legend does not show up as hyperlink for layers with legend (tried a layer with inline content, legend did not appear)

**Other issues/problems found that exist on main also -**
- [x] ZoomTo from layer control on unchecked layer (get layer unchecked by checking and unchecking a few times) throws an error - **resolved in this PR**
- [ ] Need to find a better way of displaying legends, there is already a PR that started some work on this #763, A issue already exists - #524
- [ ] #884
- [ ] #885
- [ ] #882
- [ ] #880
- [ ] #883

**Things to do after this PR is pushed -** 
- Every time a map-feature is updated update the layerbounds dynamically (make GetBounds update dynamically)
- layer.extent needs to be set by the layer- getter which would call the setLayerElExtent
- create separate classes for FeatureLayer.js, for the different use cases such as statictilelayer, query, templated, etc.
- featureRenderer - link between custom element and feature renderer, maybe migrate the resetFeature logic to it later


AliyanH:
TODO 
-  move createAndAddTemplatedLayers from on add in mapmlLayer to process content similar to how we handler other layer types
- Add a projection parameter to _convertAndFormatPCRS

AliyanH:
current known issues on experiments (need to re-verify after latest commits) -
- experiments/painting/ Girl example, clicking on girl does not zoom
- experiments/alternate-style/ maintian opacity across different styles
- experiments/linking/features - clicking on BC does not work, Ontario link not working
- experiments/vector tiles - not working

Things to do after this PR is pushed - 
- Every time a map-feature is updated update the layerbounds dynamically (make GetBounds update dynamically)
- layer.extent needs to be set by the layer- getter which would call the setLayerElExtent
- featureRenderer - link between custom element and feature renderer, maybe migrate the resetFeature logic to it later

prushforth:
Tests pass on windows, fail on linux (wsl2) locally, same error messages as on CI, so need to figure that out.

prushforth:
The QueryHandler is querying the layer control for the order of map-extent elements, but it should be querying the DOM and / or ShadowDOM for that order, since the layer control may not be in use,only the content of the layer- either in light or shadow root

Issue #886, opened 2023-09-26
Create <map-extent> custom element


prushforth:
After merging #887, we need to merge/work on the following:
- [x] Geoserver PR - https://github.com/geoserver/geoserver/pull/7245
- [x] MapML Spec - https://github.com/Maps4HTML/MapML/pull/255
- [x] web-map-doc - https://github.com/Maps4HTML/web-map-doc/pull/135
- [ ] mapml-extension - https://github.com/Maps4HTML/mapml-extension/pull/66
- [ ] experiments - https://github.com/Maps4HTML/experiments/pull/122

Related

- [ ] #786
- [ ] #767
- [ ] #646
- [ ] Generate default label "Sublayer" when label attribute not specified.

prushforth:
[Related (unfixed) issue for \<layer->](https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/319).

PR #887, opened 2023-09-26
Create map-extent, map-input, map-link

prushforth:
**Tests Needed -** 
- [x] map-extent checked (when no checked attribute present)
- [x] map-extent label default and tests 
    - [x] if no label and only one map-extent hidden 
    - [x] hidden attribute tests for local and remote 
    - [x] testing positions when hiding one of the map-extents and re-adding or what not
- [x] map-extent opacity attribute, check behavior with attribute/container style/ and behavior with how opacity changes through using slider or through API
- [x] Templated feature going out of bounds, layer is disabled (was a bug on [main](https://maps4html.org/experiments/screenreader/restaurants/))
- [x] `layer-` hidden attribute test.

**extension - https://github.com/Maps4HTML/mapml-extension/pull/66**

**docs - https://github.com/Maps4HTML/web-map-doc/pull/135** 

**To Do when developing map-link** 
- handle query, its bounds, and its ability to be disabled in the layer control
- Create a new function for map-link that gets the min/max zoom limit based on it's inputs, sample code:
```js
_getMinMaxZoom() {
    let input = this.querySelector('map-input[name=zoom]') || this.querySelector("map-meta[name=zoom]") || this.parentLayer.shadowRoot? this.parentLayer.shadowRoot.querySelector("map-meta[name=zoom]") : this.parentLayer.querySelector("map-meta[name=zoom]");
    if (input) {
     //
    } else {
      // use projection fallback
    }
  }
```
- layer not disabled when out of bounds if it has a query

**Bugs that still exists** 
~~- Debug Mode - When you check and uncheck a layer in debug mode, the layer bounds don't show up when the layer is checked, but do when layer is unchecked. Behaves oppositely.~~

**Bugs discovered on main** 
- #896 
- #897

**To Do Before closing PR** 
- [X] bundle all element support js files into mapml.js ([b6991c3](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/887/commits/b6991c3e3333dcceb2fdff2ae84ef3fd9b3708fa))

Closes https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/886


AliyanH:
**Tests Needed -** 
map-extent checked (when no checked attribute present)
map-extent label default and tests 
    - if no label and only one map-extent hidden 
    - hidden attribute tests for local and remote 
    - testing positions when hiding one of the map-extents and re-adding or what not

**extension -**
locale for dfExtent ("Sub-Layer") needed

**docs -** 
map-extent label "when an extent does not have a label, it is hidden" is no longer true
      - mapml documentation for attributes, ex. checked="checked" is required for the parser

AliyanH (commented):
Ahh this makes sense. Good catch @prushforth !

src/map-extent.js
```diff
@@ -48,58 +62,724 @@ export class MapExtent extends HTMLElement {
...
+          );
+          minNativeZoom = Math.min(
+            minNativeZoom,
+            this._templateVars[j].extentZoomBounds.minNativeZoom
+          );
+        }
+      }
+    }
+    // cannot be named as layerBounds if we decide to keep the debugoverlay logic
+    this._templatedLayer.bounds = bounds;
```
prushforth, line 762: I was wondering why the _calculateBounds method should set the bounds on the _templatedLayer, and not on map-extent?

src/map-extent.js
```diff
@@ -48,58 +62,724 @@ export class MapExtent extends HTMLElement {
...
+          projectionMatch: projectionMatch,
+          projection: this.units || FALLBACK_PROJECTION,
+          tms: tms,
+          step: step
+        });
+      }
+    }
+    return templateVars;
+  }
+  createLayerControlExtentHTML() {
```
prushforth, line 429: We could probably put this method in its own file, like we did with the similar method for the layer.  Also, if we could make sure that it sets some useful properties on map-extent like the similar layer method did for layer- , we coudl use those properties in tests to simplify the visual complexity of the tests at least.

src/layer.js
```diff
@@ -310,8 +327,16 @@ export class MapLayer extends HTMLElement {
       let layer = this._layer,
         map = layer?._map;
       if (map) {
-        let count = 0,
-          total = 0,
+        let mapExtents = this.querySelectorAll('map-extent');
```
prushforth, line 330: This section is throwing a lot of errors in experiments, perhaps because we need to query both the light dom and shadow root for map-extent?

yhy0217: I think the errors are thrown because this.querySelector returns a Nodelist, which is slightly different from array and does not have a `concat` method. Replace `concat` with the spread syntax ` ... `should work.

src/map-extent.js
yhy0217: I think it is just because in our previous discussion bounds work as a replacement/alias of **_layerBounds_**, and map-extent should not have any layerbounds property because it is an element not a layer. But it should also work, and even makes more sense if we want to bind it as a property of map-extent, just rename it as _**.extent**_ or something like this, to be consistent with what we had for layer- element.

prushforth:
Small issue discovered: when you change a remote, hidden, checked map-extent from checked to not checked in the SD (remove the attribute), then remove the hidden attribute to unhide the extent in the layer control, the extent appears checked in the layer control, and changing the checkbox to unchecked now displays the extent. (discovered in geoserver testing).

PR #909, opened 2023-11-10
No query for hidden layer

### LINKED ISSUE #904: `hidden` layer containing a query throws exception
[9:28 a.m.] Haq, Aliyan
mapml.js:3860 Uncaught TypeError: Cannot read properties of undefined (reading 'length')
    at i._query (mapml.js:3860:37)
    at i._queryTopLayer (mapml.js:3758:14)
    at i.o (leaflet-src.js:2799:15)
    at i.fire (leaflet-src.js:606:11)
    at i._fireDOMEvent (leaflet-src.js:4565:17)
    at i._handleDOMEvent (leaflet-src.js:4514:10)
    at HTMLDivElement.o (leaflet-src.js:2799:15)

To reproduce: https://maps4html.org/experiments/datacube/

- set query layers to `hidden`
- issue query (click on map)

yhy0217:
Closes #904

### DIFF
--- a/test/e2e/layers/multipleExtents.test.js
+++ b/test/e2e/layers/multipleExtents.test.js
@@ -352,7 +352,7 @@ test.describe('Multiple Extents Bounds Tests', () => {
     const alabamaExtentItem = page.getByText('alabama_feature');
     await expect(alabamaExtentItem).toHaveCount(1);
     await expect(alabamaExtentItem).toHaveCSS('font-style', 'normal');
-    
+
     const alabamaMapExtent = page.locator('map-extent[label=alabama_feature]');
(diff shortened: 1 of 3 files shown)


prushforth:
Likely not related to the `hidden` layer problem, but the query that gets executed indicates that there are two features being returned, but there is only one in fact.  The actually query link template is a bit fake in this example, because it is hard-coded to return the legend graphic from each of the datacube layers as an entry in a table in a map-feature's map-properties.  That's probably what it is counting as two features tbd.

test/e2e/layers/multipleExtents.test.js
prushforth, line 1: This is going to conflict with the other PR open at the moment

prushforth (approved):

Issue #828, opened 2023-04-19
"Pagination feature" Focus controls >| doesn't work with keyboard or mouse clicks


prushforth:
Steps to reproduce:

Load [this page](https://maps4html.org/experiments/custom-projections/BNG/)
Turn on BNG Shrink/Swell WMS layer
Zoom to BNG Shrink/Swell WMS layer (right-click on layer in layer control, select Zoom To Layer)
Tab into map or click on map on blue BNG Shrink/Swell WMS layer to display query response:

![image](https://user-images.githubusercontent.com/2437285/233158499-0ef1c7b5-8a86-469c-bb88-5868c93dd6e0.png)

Click on ">|" or tab to it and hit enter/spacebar, 

Expected behaviour:

Zoom control should receive focus

Observed behaviour:

Nothing happens

prushforth:
On map-link branch, there's an error message in the console for this bug: 

![image](https://github.com/Maps4HTML/Web-Map-Custom-Element/assets/2437285/ea276c5f-4721-4d6d-bcd6-10ce7c60a815)

PR #916, opened 2023-11-15
Map link custom element implementation

### LINKED ISSUE #898: Handle Switch from local layer- content to remote content (with layer-.src) and vice versa
When switching from local content for layer- to add a src attribute for a layer- we need to remove all the local content properly so that content is not used for the remote layer. Handle the opposite also.

Sample code for layer.js attribute changed callback:
```js
case 'src':
        if (oldValue !== newValue && this._layer) {
          this._onRemove();
          if (this.children) {
            let elements = this.children;
            for (let i = 0; i < elements.length; i++) {
              elements[i].remove();
            }
          }
          if (this.isConnected) {
            this._onAdd();
          }
          // the original inline content will not be removed
          // but has NO EFFECT and works as a fallback
        }
```

### LINKED ISSUE #915: map-link
- [x] custom element
- [ ] tbd

prushforth:
Closes #915 

- [x] create `map-link.extent` property structured per `layer-.extent` property
- [ ] ~create `M.getZoomBounds(extent)` and `M.getBounds(extent)` functions~ future work
- [ ] ~ensure that the `map-link.extent` is recalculated if changes occur to e.g. in-scope `map-meta`, `map-input min/max`~ future work 
- [x] generate private `map-link._templateVars` property during `tref` attributeChangedCallback
- [x] construct `map-link._templatedLayer` during connectedCallback
- [x] when only one map-extent with image map-link is disabled in the console, the layer does not get disabled.
- [ ] ~map-extent units attribute can be named to 'projection'?~
- [x] need to check if we ever fall into this `if block` in getBounds() 

image

- [ ] map-features independence - https://github.com/AliyanH/Web-Map-Custom-Element/tree/map-link-features 
- [x] closes #898 in 0d77e492a0e41382a562c07564459234e638820d (Switching Local content to remote, remote to local content not yet supported)
- [x] TODO map-feature addzoomlimit, resetting layer bounds, etc.
- [ ] Need to document API use, and that it can only be used after connected to DOM
- [x] Fix [bug](https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/828) in paging queried features

**Known Issues**
- [x] Vector tiles example zoom to layer does not work properly.
- [x] progressive enhancement errors
- [x] map-feature API addFeature errors

**Tests Required:**
- [ ] add test for min max zoom attributes for map-features
- [x] Dynamic Layer bounds test - i.e. check if layer bound gets updated when a new map-feature is added to layer
- [ ] map-feature zoom API, fallback should be initial map location can test using zoomTo on map-feature
- [ ] map-feature min, max utilizes attributes -> map-meta zoom if provided and then fallback
- [ ] removing a feature updates the Layer's bound and also the mapmlvectors bounds (when no overlapping map-meta extent exists)
- [x] empty layer's extent is projection extent/zoom + it updates once some content is added to the layer.
**Potential map-link tests**
 - [x] extent property
 - [x] test that extent of map-link established via map-meta compares to map-link
 - [x] extent established via map-input min/max variables, according to what the docs
 - [x] say about what takes precedence (map-input takes precedence, usually)
 - [x] test that extent of map-link changes to reflect updates to map-meta and map-input min/max
 - [x] test that layer/extent is disabled if the bounds or zoom bounds of a query link go
 - [x] out of view
 - [x] test the zoomTo() function, which relies on map-extent
 - [x] test that links that shouldn't have an extent behave accordingly
 - [x] test that you can construct a map-link programmatically and have it work
 - [x] test that links in local and remote content work the same

**Documentation Required**
- [ ] the extent of a layer's render zoombounds incorporates map-extent->map-input
- [ ] map-link -> extent property, zoomTo() method

**Future Work**
- [ ] map-meta extent methods (zoomTo, etc.) - initial work at https://github.com/AliyanH/Web-Map-Custom-Element/commit/314fdc58336ed89bab5d79bfbb559172196a6857


prushforth:
We should try to fix [this bug](https://github.com/Maps4HTML/Web-Map-Custom-Element/issues/828) in this branch

prushforth:
Just going to post a couple of notes so that they don't get overlooked.

- in /experiments/painting/, the People layer is always disbled, because the MapMLLayer._mapmlvectors.zoomBounds seems to return an empty object {}.  Probably something to do with  addLayer and/or _resetFeatures (which is the zoomend handler).
- could _leafletLayer be renamed to e.g. _MapMLLayer for clarity? Everything is a Leaflet layer in the end and it's confusing (to me).
- committing a small change to DebugOverlay which makes looking for _leafletLayer conditional on its existence, since I believe that it comes across layers that are part of the DebugOverlay which are not added to the map by any MapML element.  If confirmed, would be a good place for a comment.

prushforth:
Discovered while re-ordering multiple extents within a layer (via drag and drop):

map-link.js:307 Uncaught (in promise) DOMException: Failed to execute 'attachShadow' on 'Element': Shadow root cannot be created on a host which already hosts a shadow tree.
    at MapLink._createTemplatedLink (http://.../geoserver/mapml/viewer/widget/map-link.js:3:3619)

image

src/layer.js
```diff
@@ -436,12 +436,31 @@ export class MapLayer extends HTMLElement {
...
+      });
+    };
     for (let i = 0; i < elementsGroup.length; ++i) {
       let element = elementsGroup[i];
       switch (element.nodeName) {
         case 'MAP-FEATURE':
           _addFeatureToMapMLVectors(element);
           break;
+        case 'MAP-LINK':
+          if (element.link && !element.link.isConnected)
```
AliyanH, line 456: need to check `element.rel`?

PR #938, opened 2024-03-16
Fix excess execution of fetch

### LINKED ISSUE #936: TemplatedTileLayer requests the source HTML document URL sometimes during text/mapml vector tile loading
image

The blue icons are erroneous HTML document requests. If you put a fetch breakpoint during pan/zoom, sometimes the call to fetch(this.getTileUrl(coords)...) fetches the parent document again, because this.getTileUrl(coords) returns the empty string for some legit values of coords, apparently.

prushforth:
Fix excess execution of fetch on invalid url returned from tile coordinates that are out of bounds.

Needs a test for before/after behaviour.

Closes #936


prushforth:
Flaky tests popupTabNavigation.test.js and layerContextMenuKeyboard.test.js - both deal with activeElement and focus and routinely fail on github actions.

PR #951, opened 2024-04-02
observe and deal with map-style mutations

### LINKED ISSUE #933: Create a mutation observer for the content of map-style
When the content of the \<map-style> changes, it could propagate those changes to the \<style> element to which it's linked (`.styleElement`).   There are probably other advances that could be made with constructed stylesheets and so on, tbd, but it would be nice for the style of rendered features to update in real time with changes to the stylesheet contents.

AliyanH:
resolves #933

### DIFF
--- a/src/map-style.js
+++ b/src/map-style.js
@@ -40,6 +40,16 @@ export class MapStyle extends HTMLElement {
         target.setAttribute(attribute.nodeName, attribute.nodeValue);
       });
     }
+
+    // use observer to monitor the changes in mapStyle textContent
+    this._observer = new MutationObserver(() => {
+      this.styleElement.textContent = this.textContent;
+    });
+    this._observer.observe(this, {
+      childList: true,
+      subtree: true,
+      characterData: true
+    });
   }
   disconnectedCallback() {
     if (this._stylesheetHost) {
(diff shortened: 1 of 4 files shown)


prushforth:
Quite a few problems in experiments with a build of this branch... though the tests seem mostly clean. Strange!

src/map-style.js
```diff
@@ -40,6 +40,16 @@ export class MapStyle extends HTMLElement {
         target.setAttribute(attribute.nodeName, attribute.nodeValue);
       });
     }
+
+    // use observer to monitor the changes in mapStyle textContent
+    this._observer = new MutationObserver(() => {
+      this.styleElement.textContent = this.textContent;
+    });
+    this._observer.observe(this, {
```
prushforth, line 48: I wonder if we only need to monitor the `characterData`; there shouldn't be a subtree or child nodes except textContent

prushforth: I tried removing the `childList` and `subtree`, then just `subtree`, and the test below failed both times, so I guess the answer is : no ;-P

prushforth (approved):
LGTM

src/map-style.js
AliyanH: Yeah it was pretty odd, I played around also, and it appeared these three were required. Atleast I was able to remove `attributes: true, attributeOldValue: true` which we use in map-feature. Perhaps they can also be removed from map-feature.

Issue #935, opened 2024-03-13
Error in rendering order of map-extents when checking/unchecking lower extent


prushforth:
Steps to reproduce:

Start GeoServer with release profile on localhost port 8080.  (Not essential to the bug, just convenient to show its expression.)

Go to [this map](http://localhost:8080/geoserver/wms?service=WMS&version=1.1.0&request=GetMap&layers=Img_Sample,topp%3Astates&bbox=-1.3885038382960921E7%2C2870337.130793682%2C-7455049.489182421%2C6338174.0557576185&width=768&height=414&srs=MapML%3AOSMTILE&styles=&format=text%2Fhtml%3B%20subtype%3Dmapml)

Open the layer control for the layer settings.

Un-check the imagery layer \<map-extent>
Check the imagery layer \<map-extent>

What should happen:
The imagery layer \<map-extent> should draw underneath the vector (states) layer.

What actually happens:
The imagery layer \<map-extent> draws on top of the vector (states) layer.

PR #954, opened 2024-04-04
Fix issue #935

prushforth:
Needs a test

@AliyanH have a look to see if you agree.

[Don't look too hard at layer re-ordering I think there are bugs there too :-<](https://github.com/Maps4HTML/Web-Map-Custom-Element/pull/955)


AliyanH:
It appears we were neglecting an internal Leaflet object (`LayerControl._layers`) which manages the layerControl ordering. I ported over the function (`_update`) responsible and made a modification to sort the `_layers` object to match the ordering provided by MapMLLayer. 

The `_update` function which was ported over can probably be trimmed as required, or there may also be a better to accomplish this. For discussion.

AliyanH:
@prushforth:
>In theory the sorting should be handled by a [function passed as an option,](https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/263d14aa4e6291aecda4eee732c334b7b9aeaeed/src/mapml/control/LayerControl.js#L6) but perhaps that's not working for some reason. Looks like it all depends on layer.options.zIndex so not sure what's going on.

That sorting function we pass is being used but only in [_addLayer method of Leaflet](https://github.com/Leaflet/Leaflet/blob/39ce11753b6bc193d9fc91c676d3f2acc1dbec90/src/control/Control.Layers.js#L262)

Whenever we drag and drop layers in the Layer control (MapML updates the z-index) and the LayerControl._update method gets called which is in charge of updating the layer control, hence the need to add that sorting functionality to that _update method also.

PR #961, opened 2024-04-24
"Next feature" issue 958

### LINKED ISSUE #958: Thrown errors when processing "Next Feature" / ">" in multi-extent query on CBMTILE layer
Steps to reproduce

Start a local instance of GeoServer on port 8084 (or adjust port number as required)
Set [WMS MapML setting](https://docs.geoserver.org/latest/en/user/extensions/mapml/index.html#configuration) to serialize layers as multiple extents
Load [this layer](http://localhost:8084/geoserver/wms?service=WMS&version=1.1.0&request=GetMap&layers=Img_Sample,states,spearfish&bbox=-103.87780933359383%2C44.372359995646676%2C-103.62279724319306%2C44.5022428834876&width=768&height=540&srs=MapML%3AOSMTILE&styles=&format=text%2Fhtml%3B%20subtype%3Dmapml)
Open the devTools console drawer
Zoom in on the spearfish sub-layer:
image
Click on a bunch of features in the spearfish extent
In the popup navigation footer, click "Next Feature / '>' "
What happens: StackOverFlow
What should happen: the next feature in the set of returned features' should be popped up.

prushforth:
Closes #958


src/map-feature.js
```diff
@@ -285,26 +285,30 @@ export class MapFeature extends HTMLElement {
...
+    let csMeta, cs;
     if (this._parentEl.nodeName === 'MAP-LINK') {
       // feature attaches to link's shadow
-      csMeta =
-        this._parentEl.shadowRoot.querySelector('map-meta[name=cs]') ||
-        this._parentEl.parentElement.getMeta('cs');
+      csMeta = this._parentEl.shadowRoot.querySelector(
+        'map-meta[name=cs][content]'
+      );
+      let linkCS = this._parentEl._templateVars.values.find(
```
prushforth, line 294: This will only obtain a reasonable units value when the query tref contains a bbox={xmin},{ymin},{xmax},{ymax} parameter.  In the case of a GetTile request, the location inputs will refer to tilematrix coordinates {x} and {y} (maybe) which can't be relied upon to be the same units as the returned feature.  In that case the \<map-meta name="cs"> is required data. cc @AliyanH

src/map-feature.js
```diff
@@ -285,26 +285,30 @@ export class MapFeature extends HTMLElement {
   // native cs: used by FeatureLayer._geometryToLayer(...),
   //            the fallback cs for map-geometry if its cs attribute is not specified
   _getFallbackCS() {
-    let csMeta;
+    let csMeta, cs;
     if (this._parentEl.nodeName === 'MAP-LINK') {
       // feature attaches to link's shadow
-      csMeta =
-        this._parentEl.shadowRoot.querySelector('map-meta[name=cs]') ||
-        this._parentEl.parentElement.getMeta('cs');
```
prushforth, line 293: re: `this._parentEl.parentElement.getMeta('cs');` this was not doing anything, because the _parentEl.parentElement (\<map-extent>) getMeta function only looks for meta elements of namee=`extent` or `zoom`.  We perhaps should look at extending that to include `cs` in this case.

prushforth: Have changed \<map-extent>.getMeta to return name=cs.  tbd: if it causes problems elsewhere

src/map-feature.js
prushforth: deleted

PR #966, opened 2024-07-30
Add pmtiles support / dependency

prushforth:
Integrates protomaps-leaflet into mapml.js via rollup.
A user's style module should export the pmtilesRules object per the example in pmtilesStyles.js


prushforth:
@AliyanH I think we should try to cue up simultaneous PRs against web-map-doc and experiments repos, potentially write a short blog post on the CG home page for publishing when they land.

PR #993, opened 2024-10-03
fix for issue-319

### LINKED ISSUE #319: Should untitled layers be exposed in the layer list UI?
Modifying [us_pop_density.mapml](https://github.com/Maps4HTML/Web-Map-Custom-Element/blob/master/test/e2e/data/tiles/cbmt/us_pop_density.mapml) (used in e.g. keyboardInteraction.html) results in the following:

```html
<title><!--US Population Density--></title>
```` 

empty-title

```html
<!--<title>US Population Density</title>-->
```` 
no-title

Should untitled layers be visible in the layer list?

### PR DESCRIPTION
**Pull request by yushan-mu**
closes #319

### DIFF
--- a/test/e2e/layers/layerLabel.test.js
+++ b/test/e2e/layers/layerLabel.test.js
@@ -17,14 +17,11 @@ test.describe('Layer Label Tests', () => {
 	});
 
 	test('Name of unnamed layer is Layer', async () => {
-		const map = await page.locator('body > mapml-viewer');
-		await map.evaluate((map) =>
-			map.querySelector('layer-').removeAttribute('hidden')
-		);
-
 		const label = await page.locator(
 			'body > mapml-viewer > layer-'
 		);
+
+		await page.waitForTimeout(50);
 		expect(await label.evaluate(elem => elem.label)).toEqual('Layer');
 	});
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by prushforth** [comment]
It would be good to add a test, possibly to an existing .test.js file that fails before this change and passes after the change.

File: test/e2e/layers/layerLabel.test.js [hunk 1]
```diff
@@ -0,0 +1,40 @@
...
+	test.afterAll(async function () {
+		await context.close();
+	});
+
+	test('Name of unnamed layer is Layer', async () => {
+		const map = await page.locator('body > mapml-viewer');
+		await map.evaluate((map) =>
+			map.querySelector('layer-').removeAttribute('hidden')
+		);
+
```
AliyanH, line 24: This is not needed, as the layer does not have a "hidden" attribute, I think we copied this over from a previous test.

test/e2e/layers/layerLabel.test.js
```diff
@@ -0,0 +1,34 @@
...
+      (await context.newPage());
+    await page.goto('layerLabel.html');
+  });
+
+  test.afterAll(async function () {
+    await context.close();
+  });
+
+  test('Name of unnamed layer is Layer', async () => {
+    const label = await page.locator('body > mapml-viewer > layer-');
```
prushforth, line 19: In terms of meaningful variables, label is actually the layer.  I think you could do page.locator(...).evaluate() to return the actual label string.  Of course you would have to do that after the waitForTimeout, because locators work fast but evaluates actually access the attribute in the page.  Since you had already done the await, you could do the expect without the await inside.

Issue #992, opened 2024-10-01
Link to GeoJSON resource should not give a parser error


prushforth:
It might be nice to have the ability, not  only to paste actual GeoJSON but also to paste a link to a GeoJSON resource and have it automagically transform to a layer full of \<map-feature>s.  Currently I think it tries to evaluate the geojson text as if it was a link to a MapML document.  GeoJSON is going to be actually more common use case, so it would be nice to support that, where possible i.e. maybe not yet in ff or safari.

e.g. https://api.weather.gc.ca/collections/climate-normals/items?f=json&limit=10&lang=en

AliyanH:
+1 currently, all links are treated as MapML documents, I think we should check the URL and content type of the link and treat it accordingly, GeoJSON links (even GeoJSON files) are definitely more common then GeoJSON text.

PR #996, opened 2024-10-10
Add support for pasting GeoJSON links

yushan-mu:
fixes issue #992


src/mapml/utils/Util.js
```diff
@@ -647,25 +647,43 @@ export const Util = {
...
-      mapEl.insertAdjacentHTML('beforeend', l);
-      mapEl.lastElementChild.whenReady().catch(() => {
-        if (mapEl) {
-          // should invoke lifecyle callbacks automatically by removing it from DOM
-          mapEl.removeChild(mapEl.lastChild);
+      // get the content type of the link
+      const response = await fetch(text);
+      const contentType = response.headers.get('Content-Type');
+      if (
+        contentType == 'application/json' ||
```
prushforth, line 658: Always (almost always) use '===', sigh JavaScript

src/mapml/utils/Util.js
```diff
@@ -647,25 +647,43 @@ export const Util = {
   //    used for pasting layers through ctrl+v, drag/drop, and pasting through the contextmenu
   // _pasteLayer: HTMLElement Str -> None
   // Effects: append a layer- element to mapEl, if it is valid
```
AliyanH, line 649: update function description to reflect changes

src/mapml/utils/Util.js
```diff
@@ -643,29 +643,47 @@ export const Util = {
...
+      if (
+        contentType === 'application/json' ||
+        contentType === 'application/geo+json'
+      ) {
+        // try to process as GeoJSON
+        const textContent = await response.text();
+        try {
+          mapEl.geojson2mapml(JSON.parse(textContent));
+        } catch {
+          console.log('Invalid link!');
```
prushforth, line 666: I think thIS error message should be more along the lines of "Error parsing GeoJSON from:" + URL .  It can be hard to figure out what's gone wrong, but if we say what we were trying to do, sometimes it makes it a bit easier to understand.

src/mapml/utils/Util.js
```diff
@@ -643,29 +643,47 @@ export const Util = {
...
+          M.options.locale.dfLayer +
+          '" checked=""></layer->';
+        mapEl.insertAdjacentHTML('beforeend', l);
+        mapEl.lastElementChild.whenReady().catch(() => {
+          if (mapEl) {
+            // should invoke lifecyle callbacks automatically by removing it from DOM
+            mapEl.removeChild(mapEl.lastChild);
+          }
+          // garbage collect it
+          l = null;
```
prushforth, line 684: I don't believe this is necessary, since the let for l scopes it to this block. It should get collected after the block closes.

AliyanH:
Hi @yushan-mu, I was playing with some data from the [Region of Waterloo portal](https://rowopendata-rmw.opendata.arcgis.com/). 

Layers that have the correct content type work great, i.e. https://hub.arcgis.com/api/v3/datasets/259750cc5e8c44f78c56d27ac28b04ed_16/downloads/data?format=geojson&spatialRefId=4326&where=1%3D1

but I noticed sometimes a file would have the incorrect content type but they would infact be a geojson file, i.e. the following 2 files were geojson files, but their content type was "application/octet-stream".
 - https://services1.arcgis.com/qAo1OsXi67t7XgmS/arcgis/rest/services/Hospitals/FeatureServer/replicafilescache/Hospitals_-5969300951031363690.geojson
 - https://stg-arcgisazurecdataprod1.az.arcgis.com/exportfiles-12773-206341/Hospitals_-5969300951031363690.geojson?sv=2018-03-28&sr=b&sig=ENesGE40QyKE4HclImHiDHOLDrv9%2FK0xAucokAuIgjA%3D&se=2024-11-22T18%3A15%3A31Z&sp=r

I pushed a commit to also look for file extensions, could you kindly review and add the necessary tests, i.e. content type that are not geojson/json but file extension is geojson/json.

AliyanH (changes requested):
Excellent test coverage @yushan-mu!  

Just a few observations:
 - Please also add test coverage for the map.test.js file as you did for mapml-viewer.test.js
 - I like how you navigated server.js with the 2 geojson files, it would be good to provide meaningful names to the files so its easier for us to work with them in the future, it might be better if you rename items.json -> geojsonPoints.json and map.json -> geojsonPolygon.json
 - I noticed items.json has a lot of properties, might be worth it to remove them to make the json file smaller and easier to use in the future.

Great work!

test/server.js
```diff
@@ -113,6 +113,50 @@ app.get('/data/noMapMeta', (req, res, next) => {
     }
   );
 });
+app.get('/data/query/map', (req, res, next) => {
```
prushforth, line 116: I see that some of these are configured to simulate a badly-configured server that serves the wrong mime-type, but we could take advantage also of the abiility to map a file that has a particular file type e.g. `foo.geojson` to the correct content-type header response.  See lines 7-13 of server.js

prushforth (approved):
Good work!  Thanks! Merge when ready!

PR #1009, opened 2024-11-05
Re-focus after a layer in checked/unchecked in the layer menu, added tests

### LINKED ISSUE #1007: Checking layer using keyboard gets rid of focus
Steps to reproduce:
1. go to https://maps4html.org/experiments/screenreader/restaurants/
2. Navigate to the layer control by pressing tab and open it
3. uncheck the layer

Expected outcome: the focus should stay on the checkbox, but instead the focus disappears from the layer control until you press tab again.

yushan-mu:
Fixes #1007


test/e2e/core/kbdFocusLayerMenu.test.js
```diff
@@ -0,0 +1,75 @@
+import { test, expect, chromium } from '@playwright/test';
+
+test.describe('Focus stays on checkboxes when using keyboard navigation', () => {
```
AliyanH, line 3: test looks good, and works well! since its a small bug we should probably integrate it into an existing test rather than a new test file. WDYT about putting it at [layerControl.test.js](https://github.com/Maps4HTML/MapML.js/blob/main/test/e2e/layers/layerControl.test.js)?

yushan-mu: That should work! We should be able to run these two particular tests in slowmo without slowing down the rest of that test file :)

AliyanH (approved):
LGTM! Please Squash and merge. Great work!

PR #1017, opened 2024-11-28
Make map-properties popup size dynamic

### LINKED ISSUE #893: Make `map-properties` popup size dynamic
We should use css to size our map-properties popup, currently if we open a popup on a mobile (or a small map) it takes up the whole screen:
image

possibly related to #354

yushan-mu:
Fixes #893

### DIFF
--- a/src/mapml/layers/MapMLLayer.js
+++ b/src/mapml/layers/MapMLLayer.js
@@ -310,12 +310,20 @@ export var MapMLLayer = LayerGroup.extend({
         _leafletLayer: layer,
         mapEl: layer._layerEl.parentElement,
         onEachFeature: function (properties, geometry) {
+          const map = layer._layerEl.parentElement._map;
+          const popupOptions = {
+            autoClose: false,
+            autoPan: true,
+            maxHeight: map.getSize().y * 0.5 - 50,
+            maxWidth: map.getSize().x * 0.7,
+            minWidth: 165
+          };
           // need to parse as HTML to preserve semantics and styles
           if (properties) {
             var c = document.createElement('div');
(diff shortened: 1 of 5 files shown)


src/mapml/layers/MapMLLayer.js
```diff
@@ -312,10 +312,18 @@ export var MapMLLayer = LayerGroup.extend({
         onEachFeature: function (properties, geometry) {
           // need to parse as HTML to preserve semantics and styles
           if (properties) {
+            const map = layer._map;
+            const popupOptions = {
+              autoClose: false,
+              autoPan: true,
```
prushforth, line 318: I believe the autoPan value was the default `false` and that was intentional so that the map wouldn't keep panning and pulling in new features, which might be a problem for a screenreader user (?).  That was my theory anyhow.  I guess we can keep this value and see if it engenders any comment from anyone else though.

src/mapml/layers/TemplatedFeaturesLayer.js
```diff
@@ -70,11 +70,18 @@ export var TemplatedFeaturesLayer = Layer.extend({
         projection: map.options.projection,
         mapEl: this._linkEl.getMapEl(),
         onEachFeature: function (properties, geometry) {
+          const popupOptions = {
```
prushforth, line 73: I believe the autoPan value was the default `false` and that was intentional so that the map wouldn't keep panning and pulling in new features, which might be a problem for a screenreader user (?).  That was my theory anyhow.  I guess we can keep this value and see if it engenders any comment from anyone else though.

prushforth (approved):
Good work!  Merge when ready. Thanks!

PR #1043, opened 2026-03-23
Handle null geometry in GeoJSON query responses

prushforth:
Replace null/undefined geometries in GeoJSON query responses with a synthesized point at the click location before passing to geojson2mapml. This prevents geojson2mapml from crashing on null geometry and ensures feature properties are properly rendered as an HTML table (the default geojson2mapml behavior) rather than falling through to the raw HTML fallback which dumps the entire JSON response as text.

Adds test for null geometry scenario and updates existing test to account for the new fourth query extent (1/3 → 1/4).


prushforth:
REALLY closes https://github.com/Maps4HTML/MapML.js/issues/1039

PR #1055, opened 2026-07-03
Add Legend in layer control settings

### LINKED ISSUE #524: Re-design what mapml-viewer does with LegendURL / GetLegendGraphic links

AliyanH:
closes #524


src/mapml/elementSupport/layers/createLayerControlForLayer.js
```diff
@@ -139,12 +139,38 @@ export var createLayerControlHTML = async function () {
...
+
     legendLink.href = this._layer._legendUrl;
     legendLink.target = '_blank';
+    legendLink.rel = 'noopener noreferrer';
     legendLink.draggable = false;
-    layerItemName.appendChild(legendLink);
+    legendLink.className = 'mapml-layer-item-legend-link';
+
+    legendImage.src = this._layer._legendUrl;
+    legendImage.alt = `${this._layer._title} legend`;
```
AliyanH: need to localize

AliyanH (commented):
#

